PR #8, opened 2021-10-27
FW/CPUID: add support for Linux 5.16 state saving requests

thiagomacieira:
With this implementation, we will only issue the new system call if
CPUID indicates AMX support, in which case we ignore XCR0. This is
designed for the likely scenario of AMX-capable kernels being used for
AMX-capable hardware. If, instead, we're running on a system that fails
to enable all CPU features, they won't begrudge us a few more CPU cycles
to find out.

We should also support the case in which the 5.16+ Linux kernel does
support AMX but refuses to offer it to us. For that case, we expect the
ARCH_GET_XCOMP_SUPP call to tell us "no AMX".

### DIFF
--- a/framework/cpuid_internal.h
+++ b/framework/cpuid_internal.h
@@ -74,6 +74,49 @@ static uint64_t adjusted_xcr0(uint64_t xcr0, uint64_t xcr0_wanted)
     }
     return xcr0;
 }
+#elif defined(__linux__)
+#  include <sys/syscall.h>
+#  include <asm/prctl.h>
+#  ifndef ARCH_GET_XCOMP_SUPP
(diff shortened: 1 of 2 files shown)


thac0 (dismissed):
👍 LGTM.

thiagomacieira:
Wait for 5.16.0 to be tagged?  https://github.com/torvalds/linux/tags/v5.16

thac0:
> Wait for 5.16.0 to be tagged? https://github.com/torvalds/linux/tags/v5.16

Roger.

thiagomacieira:
Confirmed both versions (with the opt-in and without) run without problem on an actual SPR machine and using https://git.kernel.org/pub/scm/linux/kernel/git/tip/tip.git/log/?h=x86/fpu

PR #12, opened 2021-10-29
FW: cpuid_internal: include sandstone_p

thac0:
Depending on build flags, LOG_LEVEL_QUIET may be used. That define is in
sandstone_p.h, so let's include it.


ahkok (approved):

thac0:
Thanks @ahkok !

PR #14, opened 2021-11-17
meson: add 'meson test' functionality

thac0:
Nothing fancy. Just enables the "meson test" wrapper in case folks want to use it.

Side effect: unittests target is always built. If that's undesirable, we can do without this.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -232,7 +232,7 @@ executable(
 	],
 )
 
-executable(
+unittests = executable(
 	'unittests',
 	files(
 		'framework/sysdeps/linux/malloc.cpp',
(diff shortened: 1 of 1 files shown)


busykai:
I'd rather have it in CI (build unittests, run them). What do you think? At any rate, we need to add a little section to `README.md`, after building, on how to run the tests...

thac0:
No strong opinion on it-- figured I'd enable the `meson test` feature in case anyone wanted to use it. The most interesting benefit is the [malloc_perturb][1] support.

But, if we'll just be running `unittests` directly in CI and have no interest in having our actions execute `meson test`, this PR adds zero value.

[1]: https://mesonbuild.com/Unit-tests.html#malloc_perturb_

thac0:
No apparent interest, so abandoning.

PR #17, opened 2021-11-23
FW/Unix: add a SIGSEGV test for #GP

thiagomacieira:
All the tests so far were #PF. The easiest way to cause a #GP is with a
non-canonical address.

Log:
```yaml
    - { level: error, text: 'E> Received signal 11 (Segmentation fault) code=128 (SI_KERNEL), RIP = 0x44e2c7, CR2 = (nil), trap=13 (GP), error_code = 0x0' }
    - level: warning
      text: |1
       W> #4  0x000000000044e2c7 in cause_sigsegv_noncanonical () at ../../../src/arkose/arkose-one/opendcdiag/framework/selftest.cpp:429
       429          asm volatile ("movl (%1), %0" : "=r" (result) : "r" (ptr));
       => 0x44e2c7 <cause_sigsegv_noncanonical()+42>:   mov    (%rax),%eax
    - level: info
      text: |1
       Registers:
        rax   = 0x8000000000000111
```

### DIFF
--- a/framework/selftest.cpp
+++ b/framework/selftest.cpp
@@ -401,12 +401,30 @@ static void cause_sigbus()
     close(fd);
 }
 
-static void cause_sigsegv()
+static void cause_sigsegv_null()
+{
+    // not exactly null, but first page
+    uintptr_t ptr = rand() & 0xfff;
+    int result;
+    asm volatile ("movl (%1), %0" : "=r" (result) : "r" (ptr));
+}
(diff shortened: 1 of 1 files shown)


markdryan (approved):
LGTM.  You could also try a jmp or a call to a non-canonical address.  LAM isn't supposed to apply to these instructions so it might be an interesting test.

PR #18, opened 2021-11-23
FW:Meson: simplify handling of things, especially sysdeps

thiagomacieira:
This PR removes a number of unnecessary static libraries and replaces them with a simple list of files, to be added to the actual static libraries. The rest of the static libraries is `build_by_default: false` so they're only built if they're used by one of the targets that is build (namely, the opendcdiag binary or the unittests one).

Additionally, "sysdeps/win32" is renamed to "sysdeps/windows" because that's what Meson says the target machine's system name is. A few files are moved around to simplify.


framework/meson.build
```diff
@@ -1,16 +1,16 @@
...
+	'sysdeps/' + target_machine.system(),
+	'..',
+	'.',
+	'test_selectors',
+])
+if target_machine.system() != 'windows'
 	subdir('sysdeps/unix')
-	subdir('sysdeps/linux')
 endif
+subdir('sysdeps/' + target_machine.system())
```
thac0, line 13: Slick.

meson.build
```diff
@@ -255,7 +252,6 @@ executable(
 summary({
 		'Logging format': get_option('logging_format'),
 		'Min architecture': get_option('march_base'),
-		'Using main function': framework_main.name(),
```
thac0, line 260: Was that intentional?

thac0 (approved):

PR #21, opened 2021-11-30
FreeBSD port

thiagomacieira:
This PR adds a series of commits that, as far as I can tell, make OpenDCDiag compile and run properly on FreeBSD. I've only tested with GCC 10 (installed from the ports tree), which is the compiler that we support for Linux and Windows too.

The majority of commits are build fixes, correcting improper assumptions of Linux behaviour in the Unix code, such as using `splice(2)` (a Linux-specific system call) or Linux-specific constants elsewhere. There are also fixes to existing FreeBSD functionality that wasn't correct or was incomplete, because it had never been compiled or run. Finally, it turns out that FreeBSD does have memfds, so that is enabled for it.

### DIFF
--- a/framework/sandstone_iovec.h
+++ b/framework/sandstone_iovec.h
@@ -38,7 +38,7 @@ template <typename... Args>
     return writev(fd, vec, std::size(vec));
 }
 
-#if _POSIX_VERSION < 200809L && !defined(__GLIBC__)
+#ifdef _WIN32
 [[maybe_unused]] int dprintf(int fd, const char *fmt, ...)
 {
     std::string msg = va_start_and_stdprintf(fmt);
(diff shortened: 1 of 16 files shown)


thiagomacieira:
FYI:
```
$ ./opendcdiag -o- | awk '(NR > 1 && NR < 6) || /exit:/'
command-line: 'opendcdiag -o-'
version: opendcdiag-5186f4011f42
os: FreeBSD 13.0-RELEASE-p4
exit: pass
```

framework/sysdeps/freebsd/malloc.cpp
```diff
@@ -0,0 +1,85 @@
+/*
```
thac0, line 1: Ooo, I like this.

framework/sandstone_iovec.h
```diff
@@ -38,7 +38,7 @@ template <typename... Args>
     return writev(fd, vec, std::size(vec));
 }
 
-#if _POSIX_VERSION < 200809L && !defined(__GLIBC__)
+#ifdef _WIN32
```
thac0, line 41: Ooo, good catch.

thac0 (approved):
👍 Nice.

PR #22, opened 2021-12-02
macOS port

thiagomacieira:
Complementing now the FreeBSD port, this is the macOS port. This PR allows OpenDCDiag to run on a macOS system, but due to the lack of an affinity-setting API, the **macOS port is not reliable**.

But it does run:
```
$ ./opendcdiag -o- | awk '(NR > 1 && NR < 6) || /exit:/'
command-line: 'opendcdiag -o-'
version: opendcdiag-e6b8da7dc018
os: Darwin 19.6.0
exit: pass
```

### DIFF
--- a/framework/logging.cpp
+++ b/framework/logging.cpp
@@ -27,6 +27,7 @@
 #include <string.h>
 #include <sys/param.h>
 #include <sys/stat.h>
+#include <sys/time.h>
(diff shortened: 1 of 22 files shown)


thac0 (approved):
👍 Nice-- those were some esoteric fixes.

PR #23, opened 2021-12-06
meson: add dependency_link option

thac0:
Cleans up how folks configure library dependency links. Pass
'-Ddependency_link=static' if that is your preference.

Only influences dependency() static_link behavior. A riff on meson's
default_library core option.


thiagomacieira (approved):

thac0:
Obrigado!

PR #24, opened 2021-12-07
Fix Windows builds & add toolchain file

thiagomacieira:
This needs to be added to the CI. The build was failing because we are using `gmtime_r`, which MinGW implements as an inline function in time.h but only if the `_POSIX_C_SOURCE` macro is defined.


thac0 (approved):
👍

PR #25, opened 2021-12-08
FW: win32_stdlib.h: inline asm macros

thac0:
Only one consumer of macros.S, so just inline them in win32_stdlib.h
directly.


thiagomacieira (approved):

thac0:
Thanks for the review, Thiago!

PR #29, opened 2021-12-13
CI: Explicitily specify when the workflow should build.

busykai:
Currently there's no easy way to allow a job running if dependency is
either success or skipped. See https://github.com/actions/runner/issues/491


thac0 (approved):
👍

PR #30, opened 2021-12-13
FW: child_debug: add missing FreeBSD include

thac0:
The thr_self() function is defined in sys/thr.h, so we'll include it.
Without it, OpenDCDiag may not build on FreeBSD.

Bumped into this compiler error when I attempted to compile on FreeBSD
13.0 with GCC-11.


thac0:
"Works for me!" 😸

thiagomacieira (approved):

PR #31, opened 2021-12-13
CI: disable git-sanity check temporarily.

busykai:
It will not work for PRs from forked repos (no ref will be found).


thac0 (approved):
Thanks 👍

thac0 (approved):
Awesome 👍

PR #32, opened 2021-12-14
CI: make build a reusable workflow.

busykai:


thac0 (dismissed):
👍

.github/workflows/pr.yaml
```diff
@@ -39,56 +34,6 @@ jobs:
...
-            mingw64-pkg-config \
-            mingw64-winpthreads-static \
-            mingw64-zstd \
-            perl
-    - uses: actions/checkout@v2
-    - name: meson setup
-      run: meson setup builddir-windows-cross --cross-file meson-cross-win32.ini
-    - name: ninja build
-      run: ninja -C builddir-windows-cross
+    uses: opendcdiag/opendcdiag/.github/workflows/build.yaml@kai/ci-separate-workflows
```
thiagomacieira, line 39: I guess the ref name will be changed?

thiagomacieira (changes requested):

.github/workflows/pr.yaml
busykai: Yes, this was for testing. Needed it in PR cause it would fail otherwise.

thiagomacieira (approved):

PR #33, opened 2022-01-04
[FW] Allow fracturing to be disabled on the command line

markdryan:
By specifying --max-test-loop-count=0.  Previously, this option was allowed by the framework but it had no effect.  This PR changes this behaviour so that it essentially has the effect of disabling test fracturing and making test execution time limited (rather than loop count limited).


thac0 (approved):
👍 Good one!

PR #34, opened 2022-01-05
FW: compute effective CPU frequency per-thread; present average of all thread eff frequencies at test end

thac0:


fenrus75 (approved):

framework/sandstone.cpp
```diff
@@ -14,6 +14,7 @@
 #include <new>
 #include <map>
 #include <numeric>
+#include <valarray>
```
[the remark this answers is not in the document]
thac0: Good catch. That's a leftover from my initial tinkering. I'll remove it.

framework/logging.cpp
```diff
@@ -2217,6 +2222,17 @@ void YamlLogger::print(int, ChildExitStatus status)
     logging_printf(LOG_LEVEL_VERBOSE(1), "  test-runtime: %s\n",
                    format_duration(test_duration, FormatDurationOptions::WithoutUnit).c_str());
 
+    std::valarray<long double> freqs(num_cpus());
+    for (size_t i = 0; i < num_cpus(); i++)
```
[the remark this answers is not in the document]
thac0: That's my muscle memory at work. I'll remove it.

framework/logging.cpp
```diff
@@ -2217,6 +2222,17 @@ void YamlLogger::print(int, ChildExitStatus status)
     logging_printf(LOG_LEVEL_VERBOSE(1), "  test-runtime: %s\n",
                    format_duration(test_duration, FormatDurationOptions::WithoutUnit).c_str());
 
+    std::valarray<long double> freqs(num_cpus());
```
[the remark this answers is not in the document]
thac0: In this case, primarily for the overloaded `sum()` method. To be fair, nothing I'm doing here *requires* it. If there's an aversion to using a `valarray`, I'm fine using a `vector`.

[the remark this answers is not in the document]
markdryan: I guess I was really wondering why use a collection at all instead of simply doing 

```
long double freqs = 0.0;
for (int i = 0; i < num_cpus(); i++)
  freqs += data->eff_freq_mhz;

const long double freq_avg = 1.0L * freqs / num_cpus();
```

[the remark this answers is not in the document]
thac0: Tl;dr - You're right, I'll nix the container.

In an early iteration I did some additional statistics, and the valarray had some attractive conveniences. After I stripped it down to only the mean, the valarray stuck around, and I'd convinced myself the `sum()` method was oh-so-nice. In this instance, I missed the forest for the trees 🤣 .

fenrus75:
also note that you can't just average over all cpus
if the user provides a cpu mask.. not all are in use
so likely need to skip the ones with a zero frequency

thac0:
> also note that you can't just average over all cpus if the user provides a cpu mask.. not all are in use so likely need to skip the ones with a zero frequency

👍 Noted.

framework/sandstone.cpp
```diff
@@ -852,6 +859,18 @@ static void *thread_runner(void *arg)
             }
             test_end(new_state);
 
+            /* cpu freq: capture the "after" values */
+            get_aperf_and_mperf(&aperf_aft, &mperf_aft);
+            tsc_aft = __builtin_ia32_rdtscp(&tsc_aux_aft);
+            (void) get_monotonic_time_now(&time_aft);
```
busykai, line 865: Use `IGNORE_RETVAL` instead of `(void)`.

framework/sandstone.cpp
```diff
@@ -870,6 +895,26 @@ static void *thread_runner(void *arg)
...
+
+            const long double secs = (time_aft.tv_sec - time_bef.tv_sec) + (time_aft.tv_nsec - time_bef.tv_nsec) / 1000000000.0L;
+            const long double tsc_freq = (tsc_aft - tsc_bef) / secs;
+            const long double perf_ratio = 1.0L * (aperf_aft - aperf_bef) / (mperf_aft - mperf_bef);
+            return tsc_freq * perf_ratio / 1000000.0L;
+        }
+
+        void get_aperf_and_mperf(uint64_t *aperf, uint64_t *mperf)
+        {
+            (void) read_msr(thread_number, 0xe8, aperf);
```
busykai, line 914: this will not work when `taskset` or `--cpuset` option is used. OS cpu id should be passed to `(read|write)_msr`. `thread_number` is sandstone's thread id which is translated to OS cpu id as `cpu_info[thread_number].cpu_number`.

framework/sandstone.cpp
```diff
@@ -870,6 +895,26 @@ static void *thread_runner(void *arg)
                 // no rethrow
             }
         }
+
+        long double get_effective_freq_mhz(void) const
+        {
+            // For when we get bogus values (e.g. running unprivileged)
+            if (mperf_bef >= mperf_aft || aperf_bef >= aperf_aft || tsc_bef >= tsc_aft) {
```
busykai, line 902: As opposed to doing this, please implement error handling. `(read|write)_msr()` returns `false` if the value cannot be read or written.

busykai (changes requested):

framework/sandstone.cpp
```diff
@@ -852,6 +859,18 @@ static void *thread_runner(void *arg)
             }
             test_end(new_state);
 
+            /* cpu freq: capture the "after" values */
+            get_aperf_and_mperf(&aperf_aft, &mperf_aft);
+            tsc_aft = __builtin_ia32_rdtscp(&tsc_aux_aft);
```
thiagomacieira, line 864: `_rdtscp` instead of the compiler-specific builtin, please.

framework/sandstone.cpp
```diff
@@ -825,6 +825,13 @@ static void *thread_runner(void *arg)
         int ret = EXIT_FAILURE;
         int thread_number;
 
+        /* Data necessary for computing this thread's cpu frequency */
+        struct timespec time_bef, time_aft;
```
thiagomacieira, line 829: Please don't abbreviate "before" and "after".

framework/sandstone.cpp
thiagomacieira: But the values could unexpectedly roll-back, especially if we're virtualised, so range checking is still necessary.

framework/logging.cpp
```diff
@@ -2003,6 +2003,10 @@ void YamlLogger::print_thread_header(int fd, int cpu, int verbosity)
         dprintf(fd, "%s  - thread: %d\n", indent_spaces().data(), cpu);
         dprintf(fd, "%s    id: %s\n", indent_spaces().data(), thread_id_header(cpu, verbosity).c_str());
 
+        const long double eff_freq_mhz = sApp->shmem->per_thread[cpu].eff_freq_mhz;
```
thiagomacieira, line 2006: Move into the block below for verbosity level 2.

thiagomacieira (changes requested):
Can you add to your commit message why this is important? It's difficult to judge the value of this new feature without that. If it's just a curiosity, I'd rather not have it: the code is complex and adds to the per-thread output, which increases log sizes.

framework/sandstone.cpp
thac0: I'll leave it as-is then, unless folks have/want a better way of doing this.

framework/logging.cpp
```diff
@@ -2217,6 +2221,17 @@ void YamlLogger::print(int, ChildExitStatus status)
     logging_printf(LOG_LEVEL_VERBOSE(1), "  test-runtime: %s\n",
                    format_duration(test_duration, FormatDurationOptions::WithoutUnit).c_str());
 
+    long double freqs = 0.0L;
```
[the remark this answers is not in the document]
thac0: Ooo. Good catch.

framework/sandstone.cpp
thac0: Muchas gracias for the good catch.

thiagomacieira (approved):
Looks  good, just nitpicky changes left, so I am approving. You can fix the nitpicks in a new PR if you wish.

You may want to use `std::chrono::steady_clock` and its `time_point` (see sandstone_p.h's `MonotonicTimePoint`) instead of `get_monotonic_time_now` if that helps you.

thac0:
> Looks good, just nitpicky changes left, so I am approving. You can fix the nitpicks in a new PR if you wish.

Will do, thanks!

PR #36, opened 2022-01-12
CI: enable git sanity.

busykai:
This version uses the right ancestor to the merge that the checkout
github action fetches as opposed to assuming the branch is in the same
repository.


busykai:
@thiagomacieira, @thac0 seems to be working fine this way. No longer depends on the branch being in the same repo. Now it does rely on the `actions/checkout` to fetch the merge point for the PR.

thac0 (approved):
👍

PR #37, opened 2022-01-14
Adding support for native building on Windows using MSYS.

pwlazlyn:
* Use of "cygwin" build target in meson
If user uses meson package provided with python from MSYS repo (not
MINGW64) for configuration Python's sys.platform is set to "cygwin".
Meson declares to support Cygwin as a target and makes project
configuration frictionless under MSYS environment. This unfortunetly
means we have to check for "cygwin" target during meson configuration
when we check for "windows" targets, but makes configuration much
easier.

### DIFF
--- a/framework/meson.build
+++ b/framework/meson.build
@@ -1,18 +1,25 @@
 # -*- indent-tabs-mode: t -*-
 # SPDX-License-Identifier: Apache-2.0
 
+sysdeps_dir = 'sysdeps/' + target_machine.system()
+
+if target_machine.system() == 'cygwin'
+  sysdeps_dir = 'sysdeps/windows'
+endif
+
 framework_incdir = [
 	top_incdir,
 	include_directories([
 		'.',
 		'test_selectors',
-		'sysdeps/' + target_machine.system(),
+		sysdeps_dir,
 	])
 ]
-if target_machine.system() != 'windows'
+
+if target_machine.system() != 'windows' and target_machine.system() != 'cygwin'
 	subdir('sysdeps/unix')
 endif
-subdir('sysdeps/' + target_machine.system())
+subdir(sysdeps_dir)
 
 generated_cpu_features_h = configure_file(
 	output : 'cpu_features.h',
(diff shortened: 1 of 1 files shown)


thiagomacieira:
What does `host_machine.system()` say?

This looks like you're trying to create a Cygwin executable, which we don't really support. Using Cygwin to "cross-compile" to MinGW should be supported, though.

thiagomacieira:
Sounds like a meson bug. Could you report to https://github.com/mesonbuild/meson/issues, saying that the MSYS/Cygwin-based tool reports that the target/host system to be "cygwin" when it's producing mingw ("windows") binaries?

thiagomacieira (approved):
Patch is acceptable as a workaround.

thac0:
Purely FYI: https://mesonbuild.com/Reference-tables.html#operating-system-names

pwlazlyn:
> Sounds like a meson bug. Could you report to https://github.com/mesonbuild/meson/issues, saying that the MSYS/Cygwin-based tool reports that the target/host system to be "cygwin" when it's producing mingw ("windows") binaries?

Not sure if this is a bug. I think they just use whatever sys.platform holds. Cygwin is a special case and meson is configured differently when cygwin is detected - that is my understanding of documentation.

pwlazlyn:
I am going to do some minor refactor and additional testing. Will fix the issues this week and update PR.

framework/meson.build
```diff
@@ -1,16 +1,22 @@
...
+
+if target_machine.system() == 'cygwin'
+  sysdeps_dir = 'sysdeps/windows'
+endif
+
 framework_incdir = include_directories([
-	'sysdeps/' + target_machine.system(),
 	'..',
 	'.',
+	sysdeps_dir,
```
thiagomacieira, line 13: This section will conflict now. Please rebase.

thiagomacieira (changes requested):
Looks good, approved, but your code now conflicts after #45. Sorry about that.

framework/meson.build
pwlazlyn: Rebased

thiagomacieira (approved):

PR #46, opened 2022-01-21
FW/linux: fix build with GCC 10

thiagomacieira:
_rdtscp is in ia32intrin.h, which isn't included by immintrin.h with
GCC 10.


thac0 (dismissed):
👍

thac0 (approved):
Good eye. 👍

PR #47, opened 2022-01-25
Meson: add -Werror=discarded-qualifiers to the C flags

thiagomacieira:
Because C allows this:
```c
  void f(int *x);
  void g(const int *x)
  { f(x); }
```


thac0 (approved):
👍 Good call.

PR #48, opened 2022-01-27
Eigen: move HSW tests into the baseline

thiagomacieira:
These tests do not contain any Haswell-specific code (read: AVX2 or
FMA), so they don't need to be in the tests_hsw library. That library
should be used only for tests that always require HSW.

This does not change the default OpenDCDiag binary contents, because we
default to `-march=haswell` anyway. However, it does make these tests
available for the other (`-march=westmere`) build.


thac0 (approved):
👍 LGTM.

PR #50, opened 2022-02-02
[built-in list] Option to select the tests from built-in list

jposwiata:
Option selector for built-in list added


framework/sandstone.cpp
```diff
@@ -158,6 +158,7 @@ enum {
     test_delay_option,
     test_index_range_option,
     test_list_file_option,
+    test_builtin_list_option,
```
thiagomacieira, line 161: NItpick: sort order

framework/sandstone.cpp
```diff
@@ -3039,9 +3046,17 @@ int internal_main(int argc, char **argv)
             break;
 
         case test_list_file_option:
+            if ((optarg == NULL) || (*optarg == '\0')) {
+                fprintf(stderr, "No input file given\n");
```
thiagomacieira, line 3050: Please prepend `argv[0]` ("%s:") to the output

thiagomacieira (changes requested):
Please remove the "#729" from the commit message and PR header, since there's no such issue in OpenDCDiag. Other than that and the minor comments below, this looks good to me.

@brad-kelly I'd like your review on this too, please.

fenrus75:
this is a user interface change.... which may well break existing users

can you elaborate a little bit more on the why of this change? without understanding the reason it's hard to justify breaking users

jposwiata:
Old behavior is not well described and very implicit. More explicit user interface is always better. Speaking about the commit, [release note] should be added, I'll add.
I don't really know if there's anything in built-in list included (I'm not aware about releasing process yet). In the development the list is empty (and event the names does not suggest the purpose). Does any customer use this feature?

framework/sandstone.cpp
jposwiata: yes, you do ;)
Neither enum/array, nor cases are sorted. These are kept together according the features, I kept this pattern.

framework/sandstone.cpp
```diff
@@ -3032,16 +3039,24 @@ int internal_main(int argc, char **argv)
...
             } else {
-                fprintf(stderr, "Cannot determine weighted testrunner type (%s is invalid - use repeat/non-repeat)", optarg);
+                fprintf(stderr, "%s: Cannot determine weighted testrunner type (%s is invalid - use repeat/non-repeat)", argv[0], optarg);
                 return EX_USAGE;
             }
             sApp->weighted_testrunner_runtimes = NormalTestrunTimes;
             break;
 
         case test_list_file_option:
+            if ((optarg == NULL) || (*optarg == '\0')) {
```
thiagomacieira, line 3049: It can't be null because it's a `required_argument`, but it's no harm to check here.

thiagomacieira (approved):
Considering this feature is not tested and has clearly been busted for a while, I'm ok with revamping it.

framework/sandstone.cpp
```diff
@@ -3203,18 +3218,23 @@ int internal_main(int argc, char **argv)
     // If we want to use the weighted testrunner we need to initialize it
 
     if (test_list_file_path) {
-        test_selector = create_list_file_test_selector(sApp->test_list, test_list_file_path,
-                                                       sApp->starting_test_number, sApp->ending_test_number,
-                                                       sApp->test_list_randomize);
-    } else {
-        struct weighted_run_info *weighted_test_list = all_weighted_runinfo;
-        if (sApp->test_list_randomize) {
```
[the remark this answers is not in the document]
jposwiata: As I described in original issue (sandstone repo), the whole idea was to keep "best known list" of tests (does weighted_testlist stick to the purpose at all?) to execute by some customers. The list is empty either.

If a requirement for builtin list exists, my solution (+properly filled list, I'd expect new name for this list in the code as well) would be best. If not.. some code might be removed.

framework/sandstone.cpp
jposwiata: Removed to not introduce more "not very harmfull" code.

framework/sandstone.cpp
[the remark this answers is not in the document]
thiagomacieira: I know. That's my point: this feature is currently unusable right now, so we might as well do what you're proposing.

thiagomacieira (changes requested):
I can't make a review on the commit message, so it goes here. Please use `[ChangeLog][Framework]` (the "[ChangeLog]" part is the magic part).

The code is fine.

framework/sandstone.cpp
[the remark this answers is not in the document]
jposwiata: Agreed on adding some tests to dcdiag (step #1) with an option to have separate file in sandstone (step #2).

PR #51, opened 2022-02-03
FW/topology: improvements to PPIN and microcode obtention code

thiagomacieira:
Added PPIN support from /sys on Linux.

Made parsing of /proc/cpuinfo happen only once.


thac0 (approved):
👍 LGTM.

PR #52, opened 2022-02-07
FW: make the Sandstone before-main() code runs really early

thiagomacieira:
Both Linux and Windows execute the init constructors in a specific order
    which we can hook into. And Linux's init-time constructors get argv too,
    so we can do the fallback trick if we need to.


markdryan (approved):
Tested this on Linux today and it works fine.

markdryan:
> Tested this on Linux today and it works fine.

Works fine on windows as well.

PR #53, opened 2022-02-08
Meson: Force -O3 in release builds

thiagomacieira:


meson.build
```diff
@@ -64,6 +64,7 @@ elif get_option('buildtype') == 'debugoptimized'
 else
 	debug_c_flags = [
 		'-DNDEBUG',
+		'-O3',
```
thac0, line 67: Practically speaking, this does what you want. I won't hinder merge-- as my recommended changes would, in my judgement, be an undue burden on you. Especially since you're in-step with surrounding code.

I won't weigh this down with a rambling wish list, so I'll open issue(s) describing them. 😄

thac0 (approved):
👍 Does what you want.

PR #58, opened 2022-02-11
Enabling movdiri and movdiri64b instructions.

dawid-sabat:
Enabling two instructions in simd.conf: movdiri and movdir64b.
Signed-off-by: Dawid Sabat <dawid.sabat@intel.com>

### DIFF
--- a/framework/simd.conf
+++ b/framework/simd.conf
@@ -62,8 +62,8 @@ avx512vpopcntdq 	Leaf07_00ECX	    14	avx512f	# AVX512 Population Count
 #la57			Leaf07_00ECX	    16		# 5-level page tables
 #rdpid			Leaf07_00ECX	    22		# RDPID instruction
 #cldemote		Leaf07_00ECX	    25		# Cache Line Demotion
-#movdiri		Leaf07_00ECX	    27		# Move Direct-store Integer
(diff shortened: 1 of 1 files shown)


framework/simd.conf
```diff
@@ -62,8 +62,8 @@ avx512vpopcntdq 	Leaf07_00ECX	    14	avx512f	# AVX512 Population Count
 #la57			Leaf07_00ECX	    16		# 5-level page tables
 #rdpid			Leaf07_00ECX	    22		# RDPID instruction
 #cldemote		Leaf07_00ECX	    25		# Cache Line Demotion
-#movdiri		Leaf07_00ECX	    27		# Move Direct-store Integer
-#movdir64b		Leaf07_00ECX	    28		# Move Direct-store 64 bytes
+movdiri		Leaf07_00ECX	    27		# Move Direct-store Integer
```
thiagomacieira, line 65: Nitpick: can you add an extra tab here to keep the columns aligned?

thiagomacieira (changes requested):
Good otherwise.

framework/simd.conf
dawid-sabat: Done

PR #59, opened 2022-02-19
Update the x86simd_generate.pl script to generate ARCH target strings

thiagomacieira:
`__attribute__((target("arch=xxxx")))` does not work because the compilers
(GCC at least) don't test the CPU features that they are targetting, so
we keep getting "inline failed" compiler errors.

GCC bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=90129

### DIFF
--- a/framework/scripts/x86simd_generate.pl
+++ b/framework/scripts/x86simd_generate.pl
@@ -28,6 +28,7 @@ open(FH, '<', $input_conf_file) or die $!;
 
 my $i = 0;
 my @features;
+my %feature_ids;
 my @architecture_names;
 my %architectures;
 my @xsaveStates;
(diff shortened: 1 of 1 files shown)


framework/scripts/x86simd_generate.pl
```diff
@@ -160,6 +156,33 @@
...
+
+    my @features = @{$arch->{features}};
+    #@features = map { defined($feature_ids{$_}) ? $_ : () } @features;
+    if (scalar @features) {
+        $featurestr .= ' ",' if length $featurestr;
+        $featurestr .= '"' unless length $featurestr;
+        $featurestr .= join(',', @features);
+        $featurestr .= '"';
+    }
+    printf "#define QT_FUNCTION_TARGET_STRING_ARCH_%-12s %s\n", uc($arch->{id}), $featurestr;
```
fenrus75, line 183: QT_FUNCTION ?

thiagomacieira: This script is shared with Qt and "QT_FUNCTION_TARGET_STRING" appears elsewhere in the script. I didn't bother to change that when we imported it and since they were never used, they didn't need fixing.

Look at the `framework/cpu_features.h` file in your build and you'll see.

thiagomacieira: Do you want me to change this? I'd rather not.

thac0 (approved):
👍 Sigh, compiler bugs.

thiagomacieira:
> +1 Sigh, compiler bugs.

Strictly speaking, design limitations, not bugs.

PR #61, opened 2022-02-23
tests: Remove zstd_aaa and zlib_aaa older versions.

AthenasJimenez:
Removing 2 tests that are older versions of another tests. Both have been
already removed from build process, so no reason to keep them around.

tests/zlib_aaa/test.c
tests/zstd_aaa/test.c


thiagomacieira (approved):
Left-over files. Good catch.

PR #67, opened 2022-03-21
FW: Log a failure to set the CPU affinity as a platform problem

thiagomacieira:
It's not fatal, but it's bad. I thought about simply refusing to run,
but some tests will deadlock if not all threads get run, in which case
we'd compound the problem.

Removed the perror() calls.

This will produce logs like:
```yaml
  - thread: main
    messages:
    - { level: warning, text: 'W> Platform issue: Failed to set affinity to CPU 10 (logical CPU 10): Invalid argument' }
  - thread: 10
    id: { logical: 10, package: 0, core: 11, thread: 0, family: 6, model: 0x55, stepping: 4, microcode: 0x2006c0a, ppin: null }
    loop-count: 0
    messages:
    - { level: warning, text: 'W> Setting CPU affinity failed, content is likely not running on CPU 10 (logical CPU 10)' }
```


thac0 (approved):
W00t. 👍

PR #68, opened 2022-03-22
Add context to exit handler

thac0:
In order to write richer KVM content, we'll need (imho) more context provided to the exit handler. This series has the generic_run function pass `struct kvm_ctx`, `struct test`, and `cpu` info/state to the handler.

I'm fine with squashing these-- broke them up to help with review.


thiagomacieira (approved):
Would be nice to see this used in a selftest, but good enough.

PR #69, opened 2022-03-23
fw: kvm: Add setup and check handlers

markdryan:
This commit adds two new optional kvm handlers that can be used to
initialise a VM before the payload is run and to access the state
of the VM after the payload has finished executing.

A new selftest is also added to test the new handlers.


thac0 (commented):
LGTM, but I'll defer to @thiagomacieira .

PR #70, opened 2022-03-23
framework: Fix timeout calculation when --timeout flag is set.

AthenasJimenez:

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -619,7 +619,7 @@ static Duration test_timeout(Duration regular_duration)
 {
     // use the override value if there is one
     if (sApp->max_test_time != Duration::zero())
-        return sApp->max_test_time;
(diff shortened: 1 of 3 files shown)


markdryan:
Is this to fix the selftests?  I found another way of doing this  (just remove the artificial 2s limit)

https://github.com/opendcdiag/opendcdiag/pull/71

Note, this artificial 2s limit creates timeout issues for other tests as well.  It makes some of the kvm tests time out and these tests don't set max_threads.

thiagomacieira (dismissed):

AthenasJimenez:
@markdryan wouldn't --strict-runtime flag work for that purpose? Anyway, setting a specific time is kind of tricky imho, all tests have different runtimes.

markdryan:
@jtkukunas What do you think about this?  It would change the behaviour for --timeout for tests that specify max_threads.

thiagomacieira (approved):

PR #71, opened 2022-03-24
fw: selftests: remove the 2s timeout

markdryan:
A 2s timeout is currently imposed on selftests that do not specifically
request a given timeout.  This causes issues with some of the tests
who's runtimes are proportional to the number of threads in the system
under test.  These tests, e.g., selftest_timedpass_maxthreads1, cannot
declare a timeout at compile time that will work for all machines, so
it's best just to get rid of the artificial 2s limit.


markdryan:
Now I could go through the existing selftests whose runtimes are independent of the number of threads, and specifically set a 2 second limit, but I don't really see the point in doing this.  Maybe I'm missing something though.  Is the 2s timeout for selftests important?

AthenasJimenez (approved):

PR #72, opened 2022-03-24
fw: kvm: don't use report_fail_msg

markdryan:
It's better to signal an error in a KVM test by letting the functions
unwind naturally rather than cancelling the thread.  This allows the
framework to cleanup the VMs it's created and crucially to report
the error more quickly.  It takes the framework 3.5 seconds to report
the failure of a simple test on my test machine when the failure is signalled via
report_fail_msg.  After this patch, the failure is reported in 400ms.
The problem can be reproduced using the selftests, e.g.,

opendcdiag --selftests -e kvm_prot_64bit_fail


markdryan:
I guess the downside of this change is that it will make it harder to debug payload failures.

thac0 (approved):
👍

PR #73, opened 2022-03-30
fw: selftest: drop maxthreads1 from positive group

thac0:
It's a good self-test, but it's best for it to stand alone. Its
execution time can increase as core count increases. By dropping this
test from the @positive group, someone (like myself) wishing to run
@positive under strict constraints, especially timing, will no longer
have to account for a "long pole" self-test.


markdryan (approved):
LGTM

PR #75, opened 2022-04-13
simd.conf: Add Gracemont and add the marketing name for Alder Lake

thiagomacieira:
https://ark.intel.com/content/www/us/en/ark/products/codename/147470/products-formerly-alder-lake.html


thac0 (approved):
👍

PR #76, opened 2022-04-21
Fixing off by one in gdb backtrace

lukasz-tuz:
Following outcome of static code analysis, fixing an off by one error, where string null termination character was written outside of string's buffer.


thiagomacieira (changes requested):
Looks good but you need to add the signed-off-by to your commit.

lukasz-tuz:
> Looks good but you need to add the signed-off-by to your commit.

Done, git sanity checks have passed.

thiagomacieira (approved):

PR #79, opened 2022-05-04
Fix control thread pinning

thiagomacieira:
We used to do that before March 2021, but thought better of it.
Unfortunately, I accidentally brought it back when I wrote the
abstracted sysdeps/*/cpu_affinity.cpp when I moved the thread naming
functionality into pin_to_logical_processor().

As a consequence, on Unix systems, run_threads() starts each of the
worker threads with the same affinity, so they're all competing for CPU
time on the same processor. They'll eventually move themselves off to
another CPU, but not without a lot of context-switching. On systems with
100+ logical processors (such as dual-socket CLX and ICX), that's a lot
of contention.

### DIFF
--- a/framework/topology.cpp
+++ b/framework/topology.cpp
@@ -692,9 +692,6 @@ static const fill_topo_func topo_impls[] = { fill_topo_cpuid, fill_topo_sysfs };
 
 void load_cpu_info(const LogicalProcessorSet &enabled_cpus)
 {
-    int curr_cpu,
-        i;
-
     if (!sApp->thread_count)
         sApp->thread_count = enabled_cpus.count();
     assert(sApp->thread_count);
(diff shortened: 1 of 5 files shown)


framework/topology.cpp
```diff
@@ -709,22 +706,30 @@ void load_cpu_info(const LogicalProcessorSet &enabled_cpus)
...
     }
 
-    // ### start a thread with this code
-    for (i = 0, curr_cpu = 0; i < sApp->thread_count; ++i, ++curr_cpu) {
-        auto lp = LogicalProcessor(curr_cpu);
-        while (!enabled_cpus.is_set(lp)) {
-            lp = LogicalProcessor(++curr_cpu);
-            assert(curr_cpu != MAX_THREADS);
+    auto detect = [](void *ptr) -> void * {
+        auto enabled_cpus = *static_cast<const LogicalProcessorSet *>(ptr);
```
thac0, line 710: Nit and/or personal preference showing. `ptr` as an arg name seems a bit... general. `lpset` or something like it? I won't gate merge on this.

thac0 (approved):
Good catch. Left a nit, but nothing that'll gate a merge. 👍

markdryan (approved):
LGTM

framework/topology.cpp
thiagomacieira: Well, it is a generic pointer, until I cast it back to something useful. That's what this line is.

PR #80, opened 2022-05-04
FW: don't execute gdb to find out if GDB is present

thiagomacieira:
Running it ensures that it can be launched, but it's overkill and
underkill at the same time. It's overkill because its launch procedure
is quite expensive, loading Python libraries and all. But at the same
time, it doesn't tell us that it can actually debug our process, for
some esoteric reason.

So instead just check that the executable exists in PATH and has the
necessary +x bit. The access(2) system call is very cheap.


thiagomacieira:
See also https://codereview.qt-project.org/c/qt/qtbase/+/409391

thac0 (approved):
Nice. 👍

PR #81, opened 2022-05-06
Buildsystem: fix passing -march flags to the HSW and SKX libraries

thiagomacieira:
Passing march_flags followed by -march=skylake-avx512 is usually
fine... except that march_flags had -mtune-skylake and that has an
effect. We don't want non-AVX512 tuning for AVX512 code, so let GCC the
best tuning for it.

For AVX2 code, we do want -mtune=skylake, as there are a lot of those
out there, both in Xeon form (Xeon D, E and W) and in client form. At
some point in the future, we may want to switch to -mtune=alderlake,
but we're not there yet.

### DIFF
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -1,3 +1,4 @@
+# -*- indent-tabs-mode: t -*-
 # Copyright 2022 Intel Corporation.
 # SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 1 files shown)


tests/meson.build
```diff
@@ -120,11 +120,13 @@ if host_machine.cpu_family() == 'x86_64'
 		c_args : [
 			tests_common_c_args,
 			'-march=haswell',
+                        '-mtune=skylake',
```
busykai, line 123: does this have a tab?

thiagomacieira: Tabifying the file.

markdryan (approved):
LGTM

PR #82, opened 2022-05-06
FW: Pin the thread that runs the tests' init() function

thiagomacieira:
This partially reverts commit 0bbb15f7922da7d78b881507ec1d7cac916182e5,
by ensuring that the user code in the test->init() function is run in a
predictable logical processor. Since our abstraction doesn't allow us to
save and restore the ambient affinity state, just create a thread that
we can then pin (the Windows implementation is defective due to
limitations of the processor group API).

On my 4-core x 2-thread system, strace reports these sched_setaffinity()
calls on Linux, when I run:
 taskset -c 1,4,6,7 ./opendcdiag --selftests -e selftest_logs_options --disable=mce_check

```
sched_setaffinity(0, 8, [1, 4, 6, 7])   = 0
[pid 454934] sched_setaffinity(0, 128, [1]) = 0
[pid 454934] sched_setaffinity(0, 128, [4]) = 0
[pid 454934] sched_setaffinity(0, 128, [6]) = 0
[pid 454934] sched_setaffinity(0, 128, [7]) = 0
[pid 454936] sched_setaffinity(0, 128, [1]) = 0
[pid 454937] sched_setaffinity(0, 128, [1]) = 0
[pid 454938] sched_setaffinity(0, 128, [4]) = 0
[pid 454939] sched_setaffinity(0, 128, [6]) = 0
[pid 454940] sched_setaffinity(0, 128, [7]) = 0
```
The first call is taskset(1). Then the next set of calls are
load_cpu_info() and apply to the same thread (same TID is reported),
walking over each of the available logical processors in order. Then we
see one new thread started and pinned to the lowest available LP, then
we see 4 threads started and each pinned to a different processor.

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -1533,8 +1533,22 @@ static TestResult run_thread_slices(/*nonconst*/ struct test *test)
         std::fill_n(test->per_thread, sApp->thread_count, test_data_per_thread{});
 
         sApp->test_tests_init(test);
-        if (test->test_init)
-            ret = test->test_init(test);
+        if (test->test_init) {
+            // ensure the init function is run pinned to a specific logical
+            // processor but its pinning doesn't affect this control thread
+            auto init_thread_runner = [](void *testptr) {
+                auto test = static_cast</*nonconst*/ struct test *>(testptr);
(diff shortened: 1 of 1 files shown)


busykai (approved):
LGTM.

PR #85, opened 2022-05-13
GCC 12 and FP16 support

pwlazlyn:


framework/fp_vectors/Floats.h
```diff
@@ -8,34 +8,230 @@
...
+    static constexpr bool has_denorm_loss = false;
+    static constexpr bool is_iec559 = true;
+    static constexpr bool is_bounded = true;
+    static constexpr bool is_modulo = false;
+    static constexpr bool traps = false;
+    static constexpr bool tinyness_before = false;
+    static constexpr std::float_round_style round_style =
+            std::round_toward_zero;   // unlike std::numeric_limits<float>::round_style
+
+    static constexpr Float16 min()              { return Float16(Holder{0x0400}); }
```
fenrus75, line 148: calling something min() that isn't min(a,b) is a little... surprise causing

pwlazlyn: I don't think we can make ISO change their minds ;p
https://en.cppreference.com/w/cpp/types/numeric_limits/min

It is made that way so we have the same API as std::numeric_limits.

PR #86, opened 2022-05-17
FW: Add CPU Frequency Manager

AthenasJimenez:
The end goal of this is adding an option that allows OpenDcdiag tool to change randomly CPU frequency during runtime.

NOTE: This is a WIP PR, so some things still don't work and some others are simply hard-coded.

### DIFF
--- a/framework/sandstone.h
+++ b/framework/sandstone.h
@@ -387,6 +387,11 @@ struct test {
     /// free them in the test_cleanup function.
     void *data;
     struct test_data_per_thread *per_thread;
+
+    /// wheter frequency flags apply to this test or not.
+    /// 0       default (for now). Ignore frequency flags
+    /// 1       Honor frequency flags
+    int variable_frequencies;
 };
 
 /* internal function; see C macro and C++ templates at the end of this file */
--- a/framework/sysdeps/linux/frequency_manager.hpp
+++ b/framework/sysdeps/linux/frequency_manager.hpp
@@ -1,0 +1,60 @@
+/*
+ * Copyright 2022 Intel Corporation.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef GENERIC_FREQUENCY_MANAGER_HPP
+#define GENERIC_FREQUENCY_MANAGER_HPP
+
+#ifndef __x86_64__
+#  include "../generic/frequency_manager.hpp"
+#else
+
+#include <limits>
+#include <fstream>
+
+#define DEFAULT_SYS_PATH "/sys/devices/system/cpu/"
+
+class FrequencyManager
+{
+public:
+    FrequencyManager()
+    {
+        /* Discover supported and current frequencies */
+    }
+
+    void alternate_frequency()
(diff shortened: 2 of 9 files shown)


AthenasJimenez:
@thiago and @lukasz-tuz any comments on this?

framework/sandstone.h
```diff
@@ -387,6 +387,11 @@ struct test {
     /// free them in the test_cleanup function.
     void *data;
     struct test_data_per_thread *per_thread;
+
+    /// wheter frequency flags apply to this test or not.
```
lukasz-tuz, line 391: Perhaps we could make this an enum?

framework/sysdeps/linux/frequency_manager.hpp
```diff
@@ -0,0 +1,139 @@
...
+    FrequencyManager()
+    {
+        /* Discover supported and current frequencies */
+        std::filesystem::path cpuinfo_max_freq_p = "/sys/devices/system/cpu/cpu0/cpufreq/cpuinfo_max_freq";
+        max_freq_supported = get_frequency_from_file(cpuinfo_max_freq_p);
+
+        std::filesystem::path cpuinfo_min_freq_p = "/sys/devices/system/cpu/cpu0/cpufreq/cpuinfo_min_freq";
+        min_freq_supported = get_frequency_from_file(cpuinfo_min_freq_p);
+
+        std::filesystem::path scaling_max_freq_p = "/sys/devices/system/cpu/cpu0/cpufreq/scaling_max_freq";
```
lukasz-tuz, line 36: For completeness of frequency control, FrequencyManager should also include support for uncore frequency scaling on supported processors (`/sys/devices/system/cpu/intel_uncore_frequency/`).

lukasz-tuz (approved):
Looks good, just a couple of comments which might as well be resolved in subsequent commits.

framework/sysdeps/linux/frequency_manager.hpp
AthenasJimenez: For uncore frequency, I'd like to leave it for a future commit. But sure, we definitely should do it.

framework/sysdeps/linux/frequency_manager.hpp
```diff
@@ -0,0 +1,139 @@
...
+
+        for (int i=0; i< cpu_number; i++)
+        {
+            // Concatenate path
+            scaling_max_freq_p = DEFAULT_SYS_PATH;
+            scaling_max_freq_p += std::to_string(i);
+            scaling_max_freq_p += SCALING_MAX_FREQ_P;
+            
+            write_file(scaling_max_freq_p, max_freq_s);
+        }
```
lukasz-tuz, line 76: I think it'd make more sense to make this method set frequency only for a single CPU and leave setting frequencies on other cores to the caller's discretion.

framework/sandstone.h
lukasz-tuz: Ok, we obviously can't make this an enum :)

PR #87, opened 2022-06-06
meson: some nit-picky cleanups

thac0:
Finally getting around to twitchy changes that have been bothering me for some time. 😆

### DIFF
--- a/framework/sysdeps/linux/meson.build
+++ b/framework/sysdeps/linux/meson.build
@@ -2,21 +2,21 @@
 # SPDX-License-Identifier: Apache-2.0
 
 if (host_machine.cpu_family() == 'x86_64')
-        sysdeps_arch_files = files(
+	sysdeps_arch_files = files(
 		'interrupt_monitor.cpp',
-                'kvm.c',
-                'msr.c',
-        )
+		'kvm.c',
+		'msr.c',
+	)
 else
-        sysdeps_arch_files = files(
-                '../generic/kvm.c',
-                '../generic/msr.c',
-        )
+	sysdeps_arch_files = files(
+		'../generic/kvm.c',
+		'../generic/msr.c',
+	)
 endif
 
 _sysdeps_a = static_library(
 	'sysdeps_linux',
-        sysdeps_unix_files,
+	sysdeps_unix_files,
 	sysdeps_arch_files,
 	files(
 		'cpu_affinity.cpp',
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,17 @@
+# See: https://editorconfig.org/
+
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+
+[*.{c,cpp,h,hpp}]
+indent_style = space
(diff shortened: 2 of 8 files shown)


framework/simd-amd.conf
```diff
@@ -1,4 +1,3 @@
-# -*- mode: conf; indent-tabs-mode: t -*-
```
thiagomacieira, line 1: Don't remove the mode line. I'm the one who uses them, because my Emacs is configured for spaces by default (which is what we use everywhere, except these Tab-Separated Files and the meson.build files)

thiagomacieira (changes requested):

framework/simd-amd.conf
eli-schwartz: Personally, I have found https://editorconfig.org/ to be helpful here.

thac0: Here's my commit message:

> Modelines were placed in some, but not all, source files. Let's remove them for
> now. If there's demand for them, we'll add an EditorConfig file or something at
> the top level.

So, a) mode lines as-needed, or b) scrap modelines in favor of editoconfig?

thiagomacieira: I think we should add an .editorconfig for OpenDCDiag that clarifies that the Meson files use tabs and the C/C++/Perl/Python sources don't.

The simd*conf files should have in-file configuration. Does editorconfig support that?

thac0: I believe so. You can assign settings based on wildcards.

thac0: Re-pushed with a new editorconfig-- I can squash it into the "remove modelines" commit once/if folks are fine with it.

thiagomacieira: > I believe so. You can assign settings based on wildcards.

So the answer is "no". I don't want a wildcard. I want something **in** the .conf files because they are used elsewhere.

thac0: Modelines in the *.conf files is fine. I'll just have the `.editorconfig` not match them. Feels like a happy middle ground.

thac0: Latest push should do the trick. Resolving convo.

thac0:
Could you also have a look, @dawid-sabat ?

framework/sysdeps/linux/meson.build
```diff
@@ -2,21 +2,21 @@
...
+	sysdeps_arch_files = files(
+		'../generic/kvm.c',
+		'../generic/msr.c',
+	)
 endif
 
 _sysdeps_a = static_library(
 	'sysdeps_linux',
-        sysdeps_unix_files,
+	sysdeps_unix_files,
```
lukasz-tuz, line 19: Line 25 below, with 'physicaladdress.cpp', has spaces instead of tabs.

lukasz-tuz (changes requested):
Looks good, but there's just one line with spaces instead of tabs that slipped through the cracks.

thiagomacieira (approved):

PR #88, opened 2022-06-09
[float16] NaN handling with Float16/BFloat16

jposwiata:
Functions to check properties of float16 numbers: if it is general NaN,
Inf, "real" NaN, sNaN, qNaN, denormalized number, etc.
Both C and C++ interfaces covered. Generic macros (__builtin style) added.
Small copy-paste mistakes fixed.


framework/Floats.c
```diff
@@ -18,7 +18,16 @@ Float16 new_float16(unsigned sign, unsigned exponent, unsigned mantissa)
     return f;
 }
 
-Float32 new_float32(unsigned sign, unsigned exponent, unsigned mantissa)
+BFloat16 new_bfloat16(uint32_t sign, uint32_t exponent, uint32_t mantissa)
+{
+    BFloat16 f;
+    f.sign = sign & 1u;
+    f.exponent = exponent & FLOAT16_EXPONENT_MASK;
+    f.mantissa = mantissa & FLOAT16_MANTISSA_MASK;
```
thiagomacieira, line 26: Wrong constants. Just don't mask, because the bitfield will do it for you.

framework/Floats.c
```diff
@@ -55,6 +64,15 @@ Float16 new_random_float16(){
     return f;
 }
 
+BFloat16 new_random_bfloat16()
+{
+    BFloat16 f;
+    f.sign = random32() & 1u;
+    f.exponent = random32() & FLOAT32_EXPONENT_MASK;
+    f.mantissa = set_random_bits((random32() % 6) + 1, 7); // set between 1 and 6 bits of the mantissa
```
thiagomacieira, line 72: One `random32()` (and one `rand()`) has enough entropy bits for a bfloat16

framework/fp_vectors/Floats.h
```diff
@@ -30,31 +31,46 @@ typedef __fp16 fp16_t;
...
+#define BFLOAT16_NAN_EXPONENT 0xffu
 #define FLOAT32_NAN_EXPONENT  0xffu
 #define FLOAT64_NAN_EXPONENT  0x7ffu
 #define FLOAT80_NAN_EXPONENT  0x7fffu
 
+static_assert(FLOAT16_NAN_EXPONENT == FLOAT16_EXPONENT_MASK, "All Float16::NaNs have all exponent bits set");
+static_assert(FLOAT16_NAN_EXPONENT == FLOAT16_INFINITY_EXPONENT, "Float16::Inf is a special case of Float16::NaN");
+
+static_assert(BFLOAT16_NAN_EXPONENT == BFLOAT16_EXPONENT_MASK, "All BFloat16::NaNs have all exponent bits set");
+static_assert(BFLOAT16_NAN_EXPONENT == BFLOAT16_INFINITY_EXPONENT, "BFloat16::Inf is a special case of BFloat16::NaN");
```
thiagomacieira, line 55: You could also compare that they are the same as the FLOAT32 ones because bfloat16 is just a truncated float32

framework/fp_vectors/Floats.h
```diff
@@ -273,16 +352,60 @@ typedef union {
...
+#define IS_QNAN(v) \
+    _Generic((v),\
+        Float16: Float16_is_qnan,\
+        BFloat16: BFloat16_is_qnan\
+    )(v)
+#define GET_NAN_PAYLOAD(v) \
+    _Generic((v),\
+        Float16: Float16_get_nan_payload,\
+        BFloat16: BFloat16_get_nan_payload\
+    )(v)
```
thiagomacieira, line 407: Suggestion: these can be extended for the other three floating point types. Later.

thiagomacieira (changes requested):

framework/Floats.c
```diff
@@ -4,82 +4,114 @@
...
+/** @} */
+
+// function definitions
+Float16 new_random_float16()
 {
     Float16 f;
-    f.sign = sign & 1u;
-    f.exponent = exponent & FLOAT16_EXPONENT_MASK;
-    f.mantissa = mantissa & FLOAT16_MANTISSA_MASK;
+    f.as_hex = random();
```
pwlazlyn, line 84: Getting all the bits at random has a problem of getting mantissa == 0, which earlier implementation never returned.
I am not sure if we can do that. Some implementations might depend on that.
Consider reading `random32()` return value to local variable and use the bits from a single call to fill the parts of the float value. 

Something like this:
```C
    uint32_t random_bits = random32();
    
    Float32 f;
    
    f.sign = random_bits & 1u;
    random_bits >>= 1;
    
    f.exponent = random_bits & FLOAT32_EXPONENT_MASK;
    random_bits >>= FLOAT16_EXPONENT_BITS;
    
    // This internally calls `random()`, but maybe we can change it and pass random_bits as a random bits generator/pool
    f.mantissa = set_random_bits((random_bits % 9) + 1, 10);
```

framework/fp_vectors/Floats.h
```diff
@@ -98,28 +135,31 @@ extern "C" {
...
 
         struct __attribute__((packed)) {
-            uint16_t mantissa : 10;
-            uint16_t exponent : 5;
-            uint16_t sign : 1;
-        };
+            uint16_t payload  : FP16_MANTISSA_BITS - 1;
+            uint16_t quiet    : 1;
+            uint16_t exponent : FP16_EXPONENT_BITS; // BFLOAT16_NAN_EXPONENT
+            uint16_t sign     : FP16_SIGN_BITS;
```
pwlazlyn, line 148: `quiet` field uses `1` and `sign` uses special constant `FP16_SIGN_BITS`.
This is inconsistent. Let's do everything one way or the other.

framework/fp_vectors/Floats.h
```diff
@@ -98,28 +135,31 @@ extern "C" {
...
+        } nan;
```
pwlazlyn, line 149: Consider naming it `as_nan` since it is not a separate field in the structure but a way of viewing the data inside it.

framework/fp_vectors/Floats.h
```diff
@@ -223,66 +295,202 @@ struct BFloat16
...
+ * @brief C/C++ builders (inlined)
+ *
+ *
+ * @{
+ */
+#ifdef __cplusplus
+#define STATIC_INLINE static inline constexpr
+#else
+#define STATIC_INLINE static inline
+#endif
```
pwlazlyn, line 413: I feel a bit uneasy about this macro leaking outside this header.
It looks like framework-wide used macro, but defined and used only inside Floats.h.

Maybe we should do:
```cpp
#ifdef __cplusplus
#define SANDSTONE_CONSTEXPR constexpr
#else
#define SANDSTONE_CONSTEXPR
#endif
```
and move it to `sandstone_data.h`.

Also, keep in mind that `inline` specifier is different in C and C++.
Maybe we need something like `#define SANDSTONE_FORCE_INLINE __attribute__((always_inline))`.

btw macro like that will soon be added to one of the framework headers.

framework/fp_vectors/Floats.h
```diff
@@ -223,66 +295,202 @@ struct BFloat16
...
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmissing-braces"
+
+STATIC_INLINE Float16 new_float16(uint32_t sign, uint32_t exponent, uint32_t mantissa)
+{
+    return (Float16) { mantissa, exponent, sign };
```
pwlazlyn, line 420: This way of returning the value depends on order of declaration inside Float16 struct, which looks like a source of future bugs. Consider using old way of filling the mantissa, sign and exponent fields.

framework/Floats.c
```diff
@@ -4,82 +4,114 @@
...
-    Float80 f;
-    f.sign = sign & 1u;
-    f.jbit = jbit & 1u;
-    f.exponent = exponent & FLOAT80_EXPONENT_MASK;
-    f.mantissa = mantissa & FLOAT80_MANTISSA_MASK;
+    Float16 f;
+    f.sign = random32();
+    f.exponent = random32();
+    f.mantissa = set_random_bits(random32() % (FP16_MANTISSA_BITS + 1), FP16_MANTISSA_BITS);
+
```
thiagomacieira, line 74: nitpick: you're using 96 bits of random to fill a 16-bit value. Due to internals of how `random32()` works, you're actually doing **six** random number calculations to get here. Please use one single `random()` call, which gives you 31 bits of information

framework/Floats.c
```diff
@@ -4,82 +4,114 @@
...
-    Float32  f;
-    f.sign = random32() % 2;
-    f.exponent = random32() & FLOAT32_EXPONENT_MASK;
-    f.mantissa = set_random_bits((random32() % 22) + 1, 23); // set between 1 and 22 bits of the mantissa
+Float32 new_random_float32()
+{
+    Float32 f;
+    f.sign = random32();
+    f.exponent = random32();
+    f.mantissa = set_random_bits(random32() % (FLOAT32_MANTISSA_BITS + 1), FLOAT32_MANTISSA_BITS);
```
thiagomacieira, line 93: These could be one single `random` too, but let's not change them from what they are today, so we continue to generate the same random numbers as we've always used to.

framework/Floats.cpp
```diff
@@ -0,0 +1,79 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <unistd.h>
+#include <assert.h>
+#include <sandstone_data.h>
+#include <sandstone.h>
+
+/**
+ * @brief C++ assertion validators
```
thiagomacieira, line 12: Can you just merge Floats.c into this file?

framework/fp_vectors/Floats.h
```diff
@@ -30,34 +34,72 @@ typedef __fp16 fp16_t;
...
 #define FLOAT32_EXPONENT_BIAS  0x7fu
 #define FLOAT64_EXPONENT_BIAS  0x3ffu
 #define FLOAT80_EXPONENT_BIAS  0x3fffu
 
 #define FLOAT16_MANTISSA_MASK  0x3ffu
+#define BFLOAT16_MANTISSA_MASK 0x7fu
 #define FLOAT32_MANTISSA_MASK  0x7fffffu
 #define FLOAT64_MANTISSA_MASK  0xfffffffffffffu
-#define FLOAT80_MANTISSA_MASK  0xffffffffffffffffu
+#define FLOAT80_MANTISSA_MASK  0x7fffffffffffffffu
```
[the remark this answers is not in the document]
thiagomacieira: I'm not sure. Remember the J bit: float80 **has** 64 bits of mantissa, except that the J bit must always be set for any non-denormal.

framework/fp_vectors/Floats.h
```diff
@@ -30,34 +34,72 @@ typedef __fp16 fp16_t;
...
+
+#define FP16_SIGN_BITS        1
+#define FP16_EXPONENT_BITS    5
+#define FP16_MANTISSA_BITS    10
+#define FP16_QUIET_BITS       1
+
+#define BFLT16_SIGN_BITS      1
+#define BFLT16_EXPONENT_BITS  8
+#define BFLT16_MANTISSA_BITS  7
+#define BFLT16_QUIET_BITS     1
```
thiagomacieira, line 86: Can we choose one of "FLT" or "FLOAT" only? Clean ups of existing API can be a new PR.

thiagomacieira (changes requested):
Very minor changes left. I'd like to see some of patryk's comments addressed too.

framework/Floats.c
brad-kelly: Thiago, do you mean something like this:
```
auto rand = random32()
f.sign = (ran & 1)
f.exponent = rand >> 1
f.mantissa = set_random_bits(rand >> 17 % (FP16_MANTISSA_BITS + 1), FP16_MANTISSA_BITS);
```

thiagomacieira: Using `random`, not `random32`, but yes.

framework/fp_vectors/Floats.h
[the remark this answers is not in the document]
thiagomacieira: There's exactly one use of this constant and it's in Floats.c/cpp. Could we just move these constants out of the header so no test accidentally mis-uses them? (Can be new PR)

framework/fp_vectors/Floats.h
```diff
@@ -30,34 +34,72 @@ typedef __fp16 fp16_t;
...
+#define FLOAT64_QUIET_BITS    1
+
+#define FLOAT80_SIGN_BITS     1
+#define FLOAT80_EXPONENT_BITS 15
+#define FLOAT80_JBIT_BITS     1
+#define FLOAT80_MANTISSA_BITS 63
+#define FLOAT80_QUIET_BITS    1
 
 #define FP16_DECIMAL_DIG        5
 #define FP16_DENORM_MIN         5.96046447753906250000000000000000000e-8
```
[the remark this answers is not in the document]
thiagomacieira: Agreed. Can be new PR.

framework/Floats.cpp
```diff
@@ -0,0 +1,194 @@
...
+    return f;
+}
+
+Float80 new_random_float80()
+{
+    Float80 f;
+    f.sign = random32();
+    f.exponent = random32();
+    f.jbit = 1;
+    f.mantissa = set_random_bits(random32() % (FLOAT80_MANTISSA_BITS + 1), FLOAT80_MANTISSA_BITS);
```
thiagomacieira, line 188: This is the only use of `FLOAT80_MANTISSA_BITS`. But because we do set `jbit` above, the mask without the MSB is *correct*.

What's not correct is the use of only 32 bits of randomness for a 64-bit field. Ditto for the `double` case above. But pre-existing.

thiagomacieira (approved):
No more comments of mine that need to be addressed in *this* PR. Clean ups can come in new ones.

thiagomacieira (dismissed):

thiagomacieira:
@jposwiata can you add the signed-off-by and re-push?

thiagomacieira (approved):

framework/fp_vectors/Floats.h
```diff
@@ -150,32 +197,62 @@ struct Float16
...
 };
-static_assert(sizeof(struct Float16) == 2, "Float16 structure is not of the correct size");
+typedef struct Float16 Float16;
+
+// C interface
+static inline bool     Float16_is_negative(Float16 f)         { return f.sign != 0; }
+static inline bool     Float16_is_zero(Float16 f)             { return (f.exponent == FLOAT16_DENORM_EXPONENT) && (f.mantissa == 0); }
+static inline bool     Float16_is_denormal(Float16 f)         { return (f.exponent == FLOAT16_DENORM_EXPONENT) && (f.mantissa != 0); }
+
+static inline bool     Float16_is_general_nan(Float16 f)      { return f.exponent == FLOAT16_NAN_EXPONENT; }
```
brad-kelly, line 225: I'm not liking this name.  It covers infinity, which is a number.
Maybe something like:  is_inf_or_nan(f) or exponent_all_ones(f) or something.

Though this can be done as separate PR.  No need to block the PR for this.
If you feel like updating it - i can re-approve.

PR #89, opened 2022-06-14
ifs: suppress modprobe's error if it can't find the module

thiagomacieira:
It is either telling us that the module doesn't exist or that we don't
have permissions to load it. Neither is a fatal issue so we shouldn't
let this be shown in logging verbosity 1.

According to kernel developers, the module should be auto-loaded on
machines that have IFS support, so the modprobe shouldn't even be
needed. However, on the machines we tested, that autoloading was somehow
broken, so we're choosing to leave it in.

### DIFF
--- a/tests/ifs/ifs.c
+++ b/tests/ifs/ifs.c
@@ -51,7 +51,7 @@ static int scan_init(struct test *test)
                 /* modprobe kernel driver, ignore errors entirely here */
                 pid_t pid = fork();
                 if (pid == 0) {
-                        execl("/sbin/modprobe", "/sbin/modprobe", "intel_ifs", NULL);
+                        execl("/sbin/modprobe", "/sbin/modprobe", "-q", "intel_ifs", NULL);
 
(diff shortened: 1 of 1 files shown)


busykai:
> > So this is the test code, not the framework. Can we instead wait for the child to finish, capture its output and print it only if verbosity is > 1?
> 
> Yes, but I don't want to do that. It requires a pipe and it could deadlock the child if the child writes more than the pipe buffer size and we're waiting for it to end.

OK, I agree. Let's maybe just wait for the status? We can do `modprobe -q` and skip if it's non-zero.

thiagomacieira:
> OK, I agree. Let's maybe just wait for the status? We can do modprobe -q and skip if it's non-zero (and print meaningful diagnostic message, too).

We already do wait for the status. Changing to `modprobe -q`.

busykai (approved):
Thanks @thiagomacieira! I swear when I was looking at the code I didn't see the handling of the child pid. It must be the jetlag! :)

PR #92, opened 2022-06-15
[WIP] Adding background scan to OpenDCDiag.

dawid-sabat:
This is WIP for now. Opening PR to get initial feedback - we also need some modifications aroun l10n logs so do not merge it before we have them.

Background scan is run by calling the application with argument --service.
The tests are run in smaller number (max 24 tests a day) and the application can run in the background to monitor CPU utilization to run tests when the system meets idle criteria.
When all tests from the test list are executed then the testing process starts from the beginning of the list.
Also new a new way to run delays between tests is introduced. When needed a delay with some deviation can be introduced so the application will not wait the same amount of time when
running such delay.

Signed-off-by: dawid-sabat <dawid.sabat@intel.com>


thac0:
Once you mark this "ready for review", @dawid-sabat , I/we will have a look at it. 😄

dawid-sabat:
I'm waiting for changes in loggii18n which should come first. Then I mark this as ready for review.

thac0:
> I'm waiting for changes in loggii18n which should come first. Then I mark this as ready for review.

Ooof! My bad. I'm almost there, stay tuned.

framework/sandstone.cpp
```diff
@@ -2685,6 +2723,112 @@ static void warn_deprecated_opt(const char *opt)
...
+
+static void background_scan_daemon(void) 
+{
+    uint64_t t_now = 0;
+    // Don't run tests unless load is low or it's time to run a test anyway
+    while(1) {
+        // wait ~5 mins no matter what
+        sApp->delay_between_tests = sApp->background_scan.minimum_delay_between_tests;
+        wait_delay_between_tests_with_deviation(
+                sApp->background_scan.skip_wait,
```
thiagomacieira, line 2798: What's the point of this `skip_wait`? Was it meant for testing?

framework/sandstone.cpp
```diff
@@ -2685,6 +2723,112 @@ static void warn_deprecated_opt(const char *opt)
...
+        sApp->background_scan.load_idle_threshold_init +
+        (((t_now - sApp->background_scan.timestamp.back()) / sApp->background_scan.number_seconds_in_hour) * 
+        sApp->background_scan.load_idle_threshold_inc_val);
+
+    // prevent the idle threshold form rising above 0.8
+    if(sApp->background_scan.load_idle_threshold > sApp->background_scan.load_idle_threshold_max)
+        sApp->background_scan.load_idle_threshold = sApp->background_scan.load_idle_threshold_max;
+}
+
+static void background_scan_daemon(void)
```
thiagomacieira, line 2790: Misleading function name. Can you rename to something like `background_scan_wait()`?

framework/sandstone.cpp
```diff
@@ -3250,52 +3394,68 @@ int main(int argc, char **argv)
...
+                triage_tests.push_back(test);
+            }
+
+            if (total_tests_run >= sApp->max_test_count)
+                break;
+
+            if(sApp->service_background_scan == true)
+                background_scan_daemon();
+        }
+    } while(sApp->service_background_scan == true);
```
thiagomacieira, line 3458: The variable is never reset back to `false`, so we'll run forever even in the case of failures detected. I don't think we want that. Or am I missing something?

thiagomacieira (changes requested):
Thanks, idea is good. Notes on implementation.

thac0:
> I'm waiting for changes in loggii18n which should come first. Then I mark this as ready for review.

See my WIP #96 . As of writing it's not ready for review, but it should give you an idea of what's coming.

framework/sandstone.cpp
dawid-sabat: The idea here was to write wait_delay_between_tests_with_deviation in similar way as 

static void wait_delay_between_tests(bool skip_wait) which uses skip_wait or rather bool restarting = true;

framework/sandstone.cpp
dawid-sabat: I checked what's going to happen if a test returns TestFailed and we exit the loop. 
However, you raised an important issue here which reminds me to check what's going to happen when we go over the whole test list.

framework/sandstone.cpp
```diff
@@ -2423,6 +2423,44 @@ static void wait_delay_between_tests(bool skip_wait)
...
+        else {
+        useconds -= (useconds / deviation_percentage);
+        }
+
+        useconds += (2 * useconds_deviation);
+    }       
+
+    //logging_printf(LOG_LEVEL_VERBOSE(1), "Sleep time %d\n", useconds);
+    // make the system call even if useconds == 0
+    usleep(useconds);
```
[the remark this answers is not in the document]
dawid-sabat: I wanted to keep wait_delay_between_tests_with_deviation written in similar way as wait_delay_between_tests.

dawid-sabat:
@thac0 & @thiagomacieira  do your honors :)

thac0 (commented):
It looks okay, ... but the rub is policy-- namely, timing. So I'll defer approval to Thiago, since he's got more insights into this than I do.

framework/sandstone.cpp
```diff
@@ -2423,6 +2423,27 @@ static void wait_delay_between_tests(bool skip_wait)
...
+    if (!skip_wait) {
+        random_factor = frandom_scale(2.0) - 1.0; // get a random number from -1.0 to 1.0
+        random_factor = random_factor * deviation_percentage / 100;
+        Duration deviation =
+            std::chrono::duration_cast<std::chrono::microseconds>(sApp->delay_between_tests * random_factor);
+        sleep_time =
+            std::chrono::duration_cast<std::chrono::microseconds>(sApp->delay_between_tests + deviation).count();
+    }       
+
+    logging_printf(LOG_LEVEL_VERBOSE(1), "Sleep time %" PRId32"\n", sleep_time);
```
thiagomacieira, line 2442: This line needs to start with a hash mark so it's a comment on the YAML output. I'd also increase it to level 2.

framework/sandstone.cpp
```diff
@@ -2423,6 +2423,27 @@ static void wait_delay_between_tests(bool skip_wait)
...
+    double random_factor = 0.0;
+    useconds_t sleep_time = 0;
+
+    if (!skip_wait) {
+        random_factor = frandom_scale(2.0) - 1.0; // get a random number from -1.0 to 1.0
+        random_factor = random_factor * deviation_percentage / 100;
+        Duration deviation =
+            std::chrono::duration_cast<std::chrono::microseconds>(sApp->delay_between_tests * random_factor);
+        sleep_time =
+            std::chrono::duration_cast<std::chrono::microseconds>(sApp->delay_between_tests + deviation).count();
```
thiagomacieira, line 2439: Suggestion: we have a `using namespace std::chrono` earlier in this file, so  you can drop the prefixes and only say `duration_cast` and `microseconds`. The return type of the cast is also `microseconds`, not `Duration`, so you can skip the second `duration_cast`.

But this code is correct as-is.

framework/sandstone.cpp
```diff
@@ -2685,6 +2706,121 @@ static void warn_deprecated_opt(const char *opt)
             program_invocation_name, opt);
 }
 
+
+static bool system_is_idle(float idle_threshold)
+{
```
thiagomacieira, line 2711: Future PR: put this in sysdeps/unix & sysdeps/windows

framework/sandstone_p.h
```diff
@@ -221,6 +222,31 @@ template <> struct test_the_test_data<true>
...
+    const Duration half_day_delay_between_tests = std::chrono::seconds(43200);
+
+    Duration time_now = Duration::zero();
+
+    const bool init_timestamp = true;
+    const bool skip_wait = false;
+
+    const uint32_t number_of_timestamps = 24;
+    const uint32_t number_seconds_in_day = 86400;
+    const uint32_t number_seconds_in_hour = 3600;
```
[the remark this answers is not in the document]
thiagomacieira: FYI:
```c++
    static constexpr number_of_seconds_in_minute = std::chrono::minutes::period::num;
    static constexpr number_of_seconds_in_hour = std::chrono::hours::period::num;
```
etc

framework/sandstone_p.h
```diff
@@ -215,6 +215,34 @@ template <> struct test_the_test_data<true>
...
 
+struct SandstoneBackgroundScan
+{
+    static constexpr uint32_t number_seconds_in_minute = 60;
+    static constexpr uint32_t number_seconds_in_hour = (60 * number_seconds_in_minute);
+    static constexpr uint32_t number_seconds_in_day = (24 * number_seconds_in_hour);
+
+    static constexpr Duration minimum_delay_between_tests = std::chrono::seconds(5 * number_seconds_in_minute);
+    static constexpr Duration day_delay_between_tests = std::chrono::seconds(number_seconds_in_day);
+    static constexpr Duration half_day_delay_between_tests = std::chrono::seconds(number_seconds_in_day / 2);
```
thiagomacieira, line 226: Or you could write:
```c++
    static constexpr Duration minimum_delay_between_tests = std::chrono::minutes(5);
    static constexpr Duration day_delay_between_tests = std::chrono::hours(24);
    static constexpr Duration half_day_delay_between_tests = day_delay_between_tests / 2;
```

framework/sandstone.cpp
```diff
@@ -2685,6 +2706,121 @@ static void warn_deprecated_opt(const char *opt)
...
+    MonotonicTimePoint now = MonotonicTimePoint::clock::now();
+
+    // init timestamps to 1day old - this quickstarts testing on first run
+    if ((sApp->background_scan.init_timestamp == true) && (sApp->service_background_scan == true)) {
+        for (auto i = 0; i < sApp->background_scan.number_of_timestamps; i++) {
+            MonotonicTimePoint timestamp_init_val = 
+                    (now - sApp->background_scan.day_delay_between_tests);
+            sApp->background_scan.timestamp[i] = timestamp_init_val;
+        }
```
thiagomacieira, line 2750: C++ hints (totally unnecessary change, but for learning). You can write this for loop using a ranged for:
```c++
    MonotonicTimePoint timestamp_init_val =
        (now - sApp->background_scan.day_delay_between_tests);
    for (MonotonicTimePoint &p : sApp->background_scan.timestamp)
        p = timestamp_init_val;
````

Or you can just use `std::array`'s own [`fill`](https://en.cppreference.com/w/cpp/container/array/fill) function:
```c++
    sApp->backgroun_scan.timestamp.fill(now - sApp->background_scan.day_delay_between_tests));
```

Sometimes I prefer writing the `for` loop myself too, when it's un-obvious what the <algorithm> function is doing or it looks uglier. Therefore, since it's about taste, I'll leave this up to you.

framework/sandstone.cpp
```diff
@@ -2685,6 +2706,121 @@ static void warn_deprecated_opt(const char *opt)
...
+    }
+}
+
+static void background_scan_update_load_threshold(void)
+{
+    MonotonicTimePoint now = MonotonicTimePoint::clock::now();
+
+    int64_t time_from_last_test = 
+        ((now - sApp->background_scan.timestamp[sApp->background_scan.timestamp_newest]) /
+        sApp->background_scan.number_seconds_in_hour).count();
```
thiagomacieira, line 2760: If you just wanted the number of hours since the last test:
```c++
    hours time_since_last_test = duration_cast<hours>(now - last_timestamp);
```
And below you use `hours.count()`.

3599.999999999 seconds since the last timestamp will still be zero hours.

framework/sandstone.cpp
```diff
@@ -2685,6 +2706,121 @@ static void warn_deprecated_opt(const char *opt)
...
+static void background_scan_wait(void) 
+{
+    MonotonicTimePoint now = MonotonicTimePoint::clock::now();
+    // Don't run tests unless load is low or it's time to run a test anyway
+    while(1) {
+        // wait ~5 mins no matter what
+        sApp->delay_between_tests = sApp->background_scan.minimum_delay_between_tests;
+        sApp->background_scan.wait_deviation_percent = 10.0;
+        wait_delay_between_tests_with_deviation(
+                sApp->background_scan.wait_deviation_percent,
```
thiagomacieira, line 2781: Do you need to save `wait_deviation_percent` in a global variable? It doesn't seem to be used anywhere outside of this function, and you always initialise it here anyway.

framework/sandstone.cpp
```diff
@@ -2685,6 +2706,121 @@ static void warn_deprecated_opt(const char *opt)
...
+        background_scan_update_load_threshold();
+
+        // if the system is idle, run a test
+        if (system_is_idle(sApp->background_scan.load_idle_threshold))
+            break;
+
+        // if we haven't run *any* tests in the last x hours, run a test
+        // because of day/night cycles, 12 hours should help typical data center
+        // duty cycles.
+        if (now > (sApp->background_scan.timestamp[sApp->background_scan.timestamp_newest] + sApp->background_scan.half_day_delay_between_tests))
```
thiagomacieira, line 2806: Suggestion: rename the "day" variables to indicate their purpose, not their values. Name them something like "minimum time to force test running" or "max time for 24 tests".

Naming the variables after the values when you need to type more to get to the variable than the value they're holding is not ideal. You can just write `24h` and `12h` (literally) in this file.

framework/sandstone.cpp
thiagomacieira: But what happens when the inner loop exits due to a failed test? Do we mean to continue running? I think that is in conflict with `fatal_errors` and even `fatal_skips`.

I don't even understand what you're trying to accomplish here. If you meant to run forever, then I think the user should pass `-T forever` in the command-line (and we can enforce that in the restricted command-line if we want to).

Since the condition in the inner `while` is going to become complex, I recommend:
```c++
    bool should_run_more_tests;
    do {
        // other code goes here

        ++total_tests_run;
        should_run_more_tests = (total_tests_run < sApp->max_test_count);
        if (r == TestFailed) {
            ++total_failures;
            should_run_more_tests = should_run_more_tests && !fatal_errors;
        } else if (r == TestPassed) {
            ++total_successes;
        } else if (r == TestSkipped) {
            ++total_skips;
            should_run_more_tests = should_run_more_tests && !fatal_skips;
        }
    } while (should_run_more_tests);
```

thiagomacieira (changes requested):

framework/sandstone.cpp
dawid-sabat: Good idea. We don't need to refactor them in the future if we decide to change requirements and for example force to run more tests earlier or later. We'll just need to modify values then.

framework/sandstone.cpp
dawid-sabat: Let me explain what I wanted to achieve:

I thought we want the background scan to go over the whole test list and if no errors are found just start a next pass when the time is right. Running a background scan which ends at the end of test list doesn't make sense to me.
If we want that we should return something to tell the service has ended.

I thought we need an external loop to have this. However, now I think I was incorrect here.

I dug more into the code and I see we already have this:

```
        case service_option:
            fatal_errors = true;
            sApp->endtime = MonotonicTimePoint::max(); <-- here
            sApp->service_background_scan = true;
            break;
```

This should not interfere with test failures or test skips. From what I see the max value should be big enough so id doesn't overflow in reasonable time period.

framework/sandstone.cpp
```diff
@@ -2685,6 +2709,118 @@ static void warn_deprecated_opt(const char *opt)
...
+
+static void background_scan_wait(void) 
+{
+    MonotonicTimePoint now = MonotonicTimePoint::clock::now();
+    // Don't run tests unless load is low or it's time to run a test anyway
+    while(1) {
+        // wait ~5 mins no matter what
+        sApp->delay_between_tests = sApp->background_scan.minimum_delay_between_tests;
+
+        double wait_deviation_percent = 10.0;
```
thiagomacieira, line 2778: Note: 10% of an hour is 6 minutes. We may want to lower this deviation, once we get experience.

framework/sandstone.cpp
```diff
@@ -2685,6 +2709,118 @@ static void warn_deprecated_opt(const char *opt)
...
+        background_scan_update_load_threshold();
+
+        // if the system is idle, run a test
+        if (system_is_idle(sApp->background_scan.load_idle_threshold))
+            break;
+
+        // if we haven't run *any* tests in the last x hours, run a test
+        // because of day/night cycles, 12 hours should help typical data center
+        // duty cycles.
+        if (now > (sApp->background_scan.timestamp[sApp->background_scan.timestamp_newest] + sApp->background_scan.time_to_force_next_test_running))
```
thiagomacieira, line 2806: Hint: for the `static constexpr` variables, you can write `SandstoneBackgroundScan::time_to_force_next_test_running`

thiagomacieira (approved):
LGTM!

thiagomacieira:
@thac0: please merge once you're done with the submodule integration for the internal tool; that way, we can do it in stages.

PR #93, opened 2022-06-17
meson: generate sandstone_config.h

thac0:
Series uses the existing `sandstone_config.h` as a template/`.in` file (familiar pattern if you've used autotools in recent memory). This brings some "sneaky" preprocessor checks out of the header and into the framework's meson build file.

In the future I/we may find it useful to generate other headers, but we'll start with `sandstone_config.h`.

The heaviest change was to SANDSTONE_DEBUG handling. Meson's _release_ build type sets `b_ndebug` (automagically adds `-DNDEBUG`). This shifts the -D burden to meson, rather than us. SANDSTONE_DEBUG is set _unless_ the "release" build type is configured. If folks have strong feelings against this, it's trivial for me to change it back to prior behavior.

When reviewing the meson files, you'll see the `set10()` method called. Easiest to read it as "set macro to 1 or 0".


thac0:
Could you also have a look, @dawid-sabat ?

dawid-sabat:
LGTM

lukasz-tuz (approved):
Looks good.

PR #94, opened 2022-06-22
forkfd: implement vfork(2) support

thiagomacieira:
fork() works by implementing Copy-On-Write for all pages that either the
parent or the child process write to. So if the parent process continues
running while the child is between fork(2) and execve(2), then it will
keep causing page faults and requiring the OS to duplicate those pages,
with the accompanying page table update and a TLB flush. This problem
is aggravated if the parent process is multithreaded, as the simple act
of running in the parent will cause those threads' stacks to cause page
faults.

The BSD solution for that was vfork(), which has two differences in
behavior: it blocks the parent from running and it shares memory with
it. But it's always been tricky, so POSIX.1-2001 deprecated it and 2008
removed its definition completely. Because it is still sharing the same
pages, we can't return from the forkfd() function, so to implement this
functionality vforkfd() adds a callback of the same signature as glibc's
clone(2) wrapper (something that hadn't occurred to me when we attempted
to use CLONE_VFORK last time).

On FreeBSD, we ignore the request for vfork() and simply call
pdfork(). It's a compliant implementation for fork() and vfork() to be
identical, so we make it so for forkfd() and vforkfd().

On other OSes, use of vfork() is not possible, because forkfd() requires
the parent to run before the child execve()s, in order to save the child
PID in the list of children we're going to handle SIGCHLD for in a
non-racy way.

Fixes https://bugreports.qt.io/browse/QTBUG-104493. Matching Qt review:
https://codereview.qt-project.org/c/qt/qtbase/+/417829

### DIFF
--- a/framework/forkfd/forkfd.c
+++ b/framework/forkfd/forkfd.c
@@ -99,6 +99,7 @@
 
 static int system_has_forkfd(void);
 static int system_forkfd(int flags, pid_t *ppid, int *system);
+static int system_vforkfd(int flags, pid_t *ppid, int (*)(void *), void *, int *system);
 static int system_forkfd_wait(int ffd, struct forkfd_info *info, int ffdwoptions, struct rusage *rusage);
 
 static int disable_fork_fallback(void)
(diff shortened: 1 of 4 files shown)


thac0 (approved):
👍 Good one. Trophy unlocked. 😄

PR #96, opened 2022-06-23
prune away i18n/l10n functionality, repurpose for restricted logging scenario

thac0:
WIP. Not ready for prime time just yet, so hold off reviewing for now.

Putting this up so folks get the gist of what I'm trying to do. Yes, it fails CI, and I know why.

### DIFF
--- a/framework/test_knobs.cpp
+++ b/framework/test_knobs.cpp
@@ -85,22 +85,6 @@ void clear_test_knobs(){
 
 // external interface methods
 
-const char *get_testspecific_knob_value_string(const struct test *test, const char *key,
-                                               const char *value_if_not_present)
-{
-    TestKeyWrapper k(test, key);
-    std::string_view s = TestKnobSingleton::get_knob(k);
-    if (s.data()) {
-        logging_mark_knob_used(k, s, KnobOrigin::Options);
-        return s.data();
-    }
-    std::string_view v;
-    if (value_if_not_present)
-        v = value_if_not_present;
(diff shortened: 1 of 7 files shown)


framework/test_knobs.cpp
```diff
@@ -85,9 +85,11 @@ void clear_test_knobs(){
 
 // external interface methods
 
-const char *get_testspecific_knob_value_string(const struct test *test, const char *key,
-                                               const char *value_if_not_present)
+const char *get_testspecific_knob_value_string(const struct test *test, const char *key, const char *value_if_not_present)
 {
+#if SANDSTONE_I18N_LOGGING
```
thiagomacieira, line 90: This should be `SANDSTONE_RESTRICTED_CMDLINE` or (better) an `if constexpr (SandstoneConfig::RestrictedCommandLine)`

framework/test_knobs.h
```diff
@@ -18,12 +18,6 @@ extern "C" {
 #endif
 
 #include "sandstone_config.h"
-#if SANDSTONE_RESTRICTED_CMDLINE
```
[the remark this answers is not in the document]
thac0: > These should stay, because we don't want even the input strings to be written to the executable.

Gotcha.

framework/test_knobs.cpp
thac0: > This should be `SANDSTONE_RESTRICTED_CMDLINE` or (better) an `if constexpr (SandstoneConfig::RestrictedCommandLine)`

Ah, good call.

thiagomacieira (dismissed):

PR #98, opened 2022-06-27
FW/logging: include the version of the libc we're using

thiagomacieira:
In the past, we've seen some tests' effectiveness depend on the libc
version, in particular just what vector instructions memcpy(), memset(),
and memcmp() use. So it might be useful to know this when trying to
reproduce an issue later on.

Dynamic build on OpenSUSE Tumbleweed:
```
$ ./opendcdiag -o - --selftests -e selftest_pass | yq .os
"Linux 5.18.6-1-default, glibc 2.35"
```

Static build on Fedora 33, run on the same Tumbleweed:

```
$ ./opendcdiag -o - --selftests -e selftest_pass | yq .os
"Linux 5.18.6-1-default, glibc 2.32 (statically linked)"
```

### DIFF
--- a/framework/logging.cpp
+++ b/framework/logging.cpp
@@ -86,6 +86,10 @@ RtlGetVersion(
 #  include <syslog.h>
 #endif
 
+#if __has_include(<gnu/libc-version.h>)
+#  include <gnu/libc-version.h>
+#endif
+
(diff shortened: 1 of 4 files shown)


thac0 (approved):
👍

PR #99, opened 2022-06-28
meson: correct DEBUG, NDEBUG handling

thac0:
First time around, I wasn't setting SANDSTONE_DEBUG correctly. Silly me.

Then, it was realized that SANDSTONE_DEBUG wasn't necessary. So, brought back DEBUG and preprocessor handling for the DEBUG, NDEBUG macros in sandstone_config.h.

### DIFF
--- a/framework/meson.build
+++ b/framework/meson.build
@@ -66,8 +66,6 @@ framework_config.set10('SANDSTONE_GA_DEV', get_option('flavor') == 'ga-dev')
 framework_config.set10('SANDSTONE_NO_LOGGING', get_option('flavor') == 'ga')
 framework_config.set10('SANDSTONE_RESTRICTED_CMDLINE', get_option('cmdline') == 'restricted')
 
-framework_config.set10('SANDSTONE_DEBUG', get_option('b_ndebug') == 'false')
-
 framework_config.set10('SANDSTONE_CHILD_BACKTRACE', get_option('child_backtrace'))
 
 framework_config_h = configure_file(
--- a/framework/sandstone_config.h.in
+++ b/framework/sandstone_config.h.in
@@ -6,7 +6,17 @@
 #ifndef SANDSTONE_CONFIG_H
 #define SANDSTONE_CONFIG_H
 
-#mesondefine SANDSTONE_DEBUG
+#if !defined(DEBUG) && !defined(NDEBUG)
+#  define DEBUG 1
+#endif
+
+#if defined(DEBUG) && defined(NDEBUG)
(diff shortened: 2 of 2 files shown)


thiagomacieira (approved):

framework/meson.build
```diff
@@ -66,7 +66,7 @@ framework_config.set10('SANDSTONE_GA_DEV', get_option('flavor') == 'ga-dev')
 framework_config.set10('SANDSTONE_NO_LOGGING', get_option('flavor') == 'ga')
 framework_config.set10('SANDSTONE_RESTRICTED_CMDLINE', get_option('cmdline') == 'restricted')
 
-framework_config.set10('SANDSTONE_DEBUG', get_option('b_ndebug') == 'false')
+framework_config.set10('SANDSTONE_DEBUG', get_option('buildtype') != 'release')
```
thiagomacieira, line 69: Actually, I'd like you **not** to set `SANDSTONE_DEBUG` at all. It isn't used any where, but `NDEBUG` and `DEBUG` are. Meson appears to automatically add `-DNDEBUG` to release builds, but #93 removed setting `DEBUG` if not. So we need to bring back this section of code:
```c
// Make sure only one of DEBUG / NDEBUG is defined
#if !defined(DEBUG) && !defined(NDEBUG)
#  define DEBUG 1
#endif
#if defined(DEBUG) && defined(NDEBUG)
#  undef NDEBUG
#elif defined(NDEBUG)
#  define DEBUG 0
#endif
#if DEBUG
#  define SANDSTONE_DEBUG       1
#else
#  define SANDSTONE_DEBUG       0
#endif
```

thiagomacieira (changes requested):

framework/meson.build
thac0: Will do.

framework/logging.cpp
```diff
@@ -539,7 +539,7 @@ void logging_init_global_child()
     assert(sApp->current_fork_mode() == SandstoneApplication::child_exec_each_test);
 
     file_log_fd = real_stdout_fd = STDOUT_FILENO;
-#ifndef NDEBUG
+#ifdef DEBUG
```
thiagomacieira, line 542: `#ifndef NDEBUG` was correct. `NDEBUG` is the C standard way of saying "no debugging" and is used by  `<assert.h>`. The double negative must stay.

See https://cigix.me/c17#7.2

thac0: Gotcha. Oof.

thiagomacieira (approved):

PR #100, opened 2022-06-29
meson: fix set10 deprecation warning

thac0:
set10() needs to be called with a boolean, rather than an integer. Looks like
the integer argument was deprecated.


thac0:
Upstream meson deprecated this as of the 0.62 release. Still works, but officially deprecated.

```9c7ddf59f ("interpreterobjects: deprecated passing a number to configuration_data.set10")```

PR #101, opened 2022-06-29
FW: add a way to specify a version suffix

thiagomacieira:
We have multiple builds of our internal tool and we actually have a good CI process where everything gets build from the same source... but then we can't tell apart runs from different builds by just looking at the logs. So add a way for providing suffixes.
```
$ meson configure -Dversion_suffix=thiago
$ ninja
[48/48] Linking target opendcdiag
$ ./opendcdiag --version
opendcdiag-3a84c774ec3e-thiago
```

### DIFF
--- a/framework/logging.cpp
+++ b/framework/logging.cpp
@@ -90,6 +90,8 @@ RtlGetVersion(
 #  include <gnu/libc-version.h>
 #endif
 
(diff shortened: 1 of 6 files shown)


thac0 (approved):
👍 Slick. Worked for me locally.

PR #102, opened 2022-06-29
Meson config updates

thiagomacieira:
Use sandstone_config.h more extensively.

### DIFF
--- a/framework/meson.build
+++ b/framework/meson.build
@@ -63,9 +63,10 @@ endif
 framework_config.set10('SANDSTONE_GA', get_option('flavor') == 'ga')
 framework_config.set10('SANDSTONE_GA_DEV', get_option('flavor') == 'ga-dev')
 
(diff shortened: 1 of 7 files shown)


framework/meson.build
```diff
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 framework_config = configuration_data()
-framework_config.set10('SANDSTONE_STATIC', 0) # sysdeps may override
+framework_config.set10('SANDSTONE_STATIC', false) # sysdeps may override
```
thac0, line 5: You beat me! 😁

thac0 (dismissed):
Smiles all around. 👍

framework/meson.build
thac0: Oh-- missed the one in sysdeps/unix. Won't hold up merge for it-- we can snag it later.

thac0 (approved):
👍

PR #103, opened 2022-06-30
FW: Change the memory randomisation from malloc() to mmap()

thiagomacieira:
The malloc() call wasn't helping much. It was indeed moving the heap
allocation pointer for the main thread, but that had never affected the
worker threads. Moreover, since commit 4248649b550477eb299efcdbfd35be85b
(PR #82), the init() function isn't run in the main thread either, so
not even memory allocated in the tests' init functions would be
randomised.

Instead, if we use mmap(), we shift the entire address space, causing
the stacks and malloc() arenas to shift.

Implementation detail: because we use pthread_create() in
load_cpu_info(), libpthread has created at least one mmap for a thread's
stack. As of version 2.35, it caches that, so the stack will be used for
the thread for init() and for thread_num == 0. Moreover, the threaded
code in load_cpu_info() allocates memory, so a large (128 MB) arena is
created for it too.

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -1961,9 +1961,12 @@ TestResult run_one_test(int *tc, const struct test *test, SandstoneApplication::
     TestResult state = TestSkipped;
     int fail_count = 0;
     int iterations;
-    std::unique_ptr<char[]> random_allocation;
     MonotonicTimePoint current_test_endtime, first_iteration_target;
     bool auto_fracture = false;
+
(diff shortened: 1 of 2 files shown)


thiagomacieira:
I'd like @fenrus75's opinion too.

thiagomacieira:
Looks like I have some uncommitted changes related to this. Moving to draft.

thiagomacieira:
Will be subsubmed by the work to scale to 8-socket systems.

PR #104, opened 2022-06-30
FW: move the CPU-supported checking into sandstone_main()

thiagomacieira:
Benefits:
* Can use the loaded CPU info from load_cpu_info()
* Runs after the command-line is processed, so --version and --help work
* Adds some code after the execution, to maybe modify it


thac0:
I want to sink my teeth into this and explore its ramifications, so I'll get to this on Tuesday morning.

PR #105, opened 2022-07-02
Meson: fix the version suffix when the commit is tagged

thiagomacieira:
Partially reverts commit 6e7b7a44a5c7bd53565b35a0a53d9df4da0ab96f (PR #101),
which added the concept of version suffix, but it was inserted in the
incorrect place for tagged releases.

For example:
```
$ git tag opendcdiag-2459761.5
$ git archive @ : | tar -x .hash
$ cat .hash
ea83aacaa2302d36f6fc5c3572d71f85963d166d HEAD -> main, tag: opendcdiag-2459761.5
```
That used to produce:
```c
#define GIT_ID "v2459761.5 (ea83aacaa2302d36f6fc5c3572d71f85963d166d)"
```

Which would result in OpenDCDiag printing:

```yaml
version: v2459761.5 (ea83aacaa2302d36f6fc5c3572d71f85963d166d)-foobar
```

This commit moves it to the right place:

```c
#define GIT_ID "v2459761.5-foobar (ea83aacaa2302d36f6fc5c3572d71f85963d166d)"
```

### DIFF
--- a/framework/logging.cpp
+++ b/framework/logging.cpp
@@ -90,11 +90,7 @@ RtlGetVersion(
 #  include <gnu/libc-version.h>
 #endif
 
-#ifdef SANDSTONE_VERSION_SUFFIX
-#  define PROGRAM_VERSION         EXECUTABLE_NAME "-" GIT_ID "-" SANDSTONE_VERSION_SUFFIX
-#else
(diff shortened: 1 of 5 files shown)


thac0 (approved):
👍

PR #106, opened 2022-07-05
[RFC] fw: remove --max-test-count

thac0:
The semantics of this option can get tricky, especially with fracturing
enabled. I don't believe this option gets much play, so removing it.


thiagomacieira (approved):

busykai:
@thac0 this is good! are you going to finish this up?

thac0:
Oh! This! 🤣 Forgot about it. I'll take a peek when I have a cycle.

PR #113, opened 2022-07-07
Reorganise the main loop, like we had been asking for

thiagomacieira:
This also updates the background scan to wait if the last test failed (though usually background scans are run with `--fatal-errors`).


thac0 (approved):
👍

PR #114, opened 2022-07-07
Meson: fix make-gitid.pl with some older Git versions

thiagomacieira:
They didn't understand the -C option, resulting in the script failing to
get a valid info from Git and thus producing
```c
 #define GIT_ID "<unknown version>"
```
Incidentally, this fixes running the script without $MESON_SOURCE_ROOT
set, which would have caused a run of `git -C describe --always`, which
also fails and also produces that unknown version.


lukasz-tuz (approved):
LGTM

PR #115, opened 2022-07-07
FW: fix which timestamp we were checking for the batch time

thiagomacieira:
We wanted the oldest, not newest, as evidenced by the fact that `timestamp_oldest` wasn't used anywhere.


thac0 (approved):
👍

PR #116, opened 2022-07-07
Add builtin test list

thiagomacieira:
This replaces the previous weighted test list, which was missing, with a simple sequential order.


thac0 (dismissed):
👍 Solid.

framework/scripts/generate-test-list.pl
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/perl
+use strict;
```
thiagomacieira, line 2: Missing copyright notice.

thac0 (approved):
👍

framework/scripts/generate-test-list.pl
thac0: Gah, I missed that.

PR #119, opened 2022-07-08
Background scan refactor 1

thiagomacieira:
This is a set of simple no-op changes, just to improve the code a little
 * removed (void) parameters
 * passing "now" as a variable to background_scan_update_load_threshold()
 * merged background_scan_update() into background_scan_wait()
 * removed unnecessary init_timestamp constant variable
 * removed the SandstoneBackgroundScan::timestamp_* constants
 * moved the background scan's time constants to a namespace
 * renamed and capitalised said constants names


thac0 (approved):
👍

PR #120, opened 2022-07-08
Background scan refactor (2/2)

thiagomacieira:
This PR continues the fixing and refactoring started in the previous refactor (#119)
 * fix std::move of the array, which we attempted to fix but didn't do the job properly
 * fix initialisation of the background scan timestamps
 * add some more logging info to the background scan
 * replace system_is_idle() with a function that returns the load
 * move wait_delay_between_tests_with_deviation() into a lambda
 * refactor the background scan's random time-waiting code
 * hoist the long wait out of the loop
 * don't wait too long between test starts

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -2800,7 +2800,7 @@ static void background_scan_wait()
 
     // move all timestaps except the oldest one
     auto array_data = sApp->background_scan.timestamp.data();
-    std::move(array_data + 1, sApp->background_scan.timestamp.end(), array_data);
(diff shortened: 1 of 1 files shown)


thiagomacieira:
This isn't right yet. I'm forcing /proc/loadavg to a specific value via `mount --bind` trick and it's doing:
```
# Sleeping for 300 +/- 30= 272.487 s
# Sleeping for 86400 +/- 86= 86456.060 s
# Sleeping for 300 +/- 30= 316.978 s
# Sleeping for 86400 +/- 86= 86478.525 s
```

lukasz-tuz (approved):
Looks good.

PR #121, opened 2022-07-09
FW: remove the max 127 test failure limitation

thiagomacieira:
We used to exit the tool with the number of tests that had failed a long time ago. We don't do this any more, because each exit code has a special meaning.


lukasz-tuz (approved):
:ok:

PR #123, opened 2022-07-09
FW: move the setting of the verbosity up, before we print warnings

thiagomacieira:
Otherwise, LOG_LEVEL_QUIET isn't enough.


lukasz-tuz (approved):
:ok:

PR #124, opened 2022-07-12
FW: update the default restricted-cmdline options

thiagomacieira:
I've removed the ones that only reset the values back to their defaults.


lukasz-tuz (approved):
:ok:

PR #129, opened 2022-07-14
[WIP][rng_bits] Building floats with less intense usage of RNG

jposwiata:
Functions to fetch required number of bits from RNG and to create
random chains of 0s and 1s to build random floats.

This PR depends on https://github.com/opendcdiag/opendcdiag/pull/139 (which renames some #defs)


brad-kelly:
I'm still not convinced that this will result in a better distribution of values.

I think this needs top be run through some deep analysis of the values created to determine if this is an improvement or a drop in FP dataspace coverage.  I also think this needs to go through the SDE cluster to check the impact on existing testing.

framework/random.cpp
```diff
@@ -597,40 +597,121 @@ void *memset_random(void *buf, size_t n)
     return buf;
 }
 
-uint64_t set_random_bits(unsigned num_bits_to_set, uint32_t bitwidth) {
-    if (num_bits_to_set >= 64 && bitwidth >= 64)
-        return 0xFFFFFFFFFFFFFFFF;  // can't be handled by shifting and subtracting :-(
-    else if (num_bits_to_set >= bitwidth || num_bits_to_set >= 64)
-        return (1ul << bitwidth) - 1ul;
+bool random_compatitibilitiy_mode()
```
thiagomacieira, line 600: Spelling of "compatibility".

framework/sandstone.cpp
```diff
@@ -173,6 +173,7 @@ enum {
     use_predictable_file_names_option,
     version_option,
     weighted_testrun_option,
+    random_compatibility_mode,
```
thiagomacieira, line 176: Sorting order. Repeats below for the `struct options` and for the `switch` case label.

framework/sandstone.cpp
```diff
@@ -1115,13 +1116,16 @@ Common command-line options are:
  -s <STATE>, --rng-state=<STATE>
      Specify the random generator state to reload. The seed is in the form:
        Engine:engine-specific-data
+ --random-compatibility-mode
+     Set compatibility mode of Random Number Generators (with very intense usage
```
thiagomacieira, line 1120: Need a clearer phrasing here. In particular, is the default behaviour same as it has always been? Or does this option turn on a new random mode?

Given the name, I suppose that the new default is to use the new technique, so the description should be "Enables generation of random floating-point numbers with random mantissa bits" or something similar.

framework/sandstone.h
```diff
@@ -493,12 +529,62 @@ static inline void *aligned_alloc_safe(size_t alignment, size_t size)
...
+ *
+ * It is allowed to follow old pattern of random number (exact number of bits consumed from
+ * RNG each time), to execute the tests with exactly the same data with previous (pre-100)
+ * suites. For compatibility mode old set_random_bits() is used, otherwise mantissa of floats
+ * is composed from 0s/1s chains (of particular lengths)
+ *
+ * @return true if compatibility mode was requested
+ * @return false to follow new patterns of random data
+ */
+bool random_compatitibilitiy_mode();
```
thiagomacieira, line 543: Does this need to be in sandstone.h? That is, does or will any test need this to make decisions? If not, then place place it in sandstone_p.h (private framework API).

If it is meant to stay here, since this is in the C section of the header, add `void` to the parameter list. Also, please move this next to the other random functions (please move `set_random_bits` too, I don't know why `aligned_alloc_safe` is between it and the rest of the random functions) and summarise the docs like the others (no @brief, no @return) - a single sentence starting with "Returns whether" suffices.

framework/sandstone.h
```diff
@@ -493,12 +529,62 @@ static inline void *aligned_alloc_safe(size_t alignment, size_t size)
...
+ *
+ * @param bits  number of bits to fetch (and to "consume" from RNG)
+ * @param state RNG state
+ * @return uint64_t
+ */
+uint64_t random_bits(uint8_t bits, random_bits_state_t* state);
+
+/// macro to find most significant bit in u16 value
+/// Expected to be used in compile time expressions
+#define _POW2_BIT(n, b) ((n) & (1 << ((b) - 1))) ? (b) :
```
thiagomacieira, line 568: Nitpick: underscore + capital is reserved for the compiler. I suggest moving the underscore to the end.

Same question as above: do tests need this or can this be in sandstone_p.h or just random.cpp?

framework/sandstone_p.h
```diff
@@ -334,6 +334,7 @@ struct SandstoneApplication : public InterruptMonitor, public test_the_test_data
     Duration delay_between_tests = std::chrono::milliseconds(5);
 
     std::unique_ptr<RandomEngineWrapper, RandomEngineDeleter> random_engine;
+    bool random_compatibility_mode = false;
```
thiagomacieira, line 337: Please find a group of `bool` to move this next to. There's a 1-byte padding hole right now near `ud_on_failure` and `service_background_scan`.

Additionally, this variable seems to be needed in the exec state of the child process, so please add it to the `APP_STATE_VARIABLES` macro.

framework/sandstone.h
```diff
@@ -145,6 +145,42 @@ extern "C" {
...
+ * after expansion V might be used with single (e.g. where float value is passed)
+ * or with two arguments (e.g. where number might be expressed as rational number)
+ * All possible expressions must be defined, otherwise the code won't compile.
+ * @{
+ */
+#define NARGN(\
+            __1,__2,__3,__4,__5,__6,__7,__8,__9,_10,_11,_12,_13,_14,_15,_16,_17,_18,_19,_20,_21,_22,_23,_24,_25,_26,_27,_28,_29,_30,_31,\
+            _32,_33,_34,_35,_36,_37,_38,_39,_40,_41,_42,_43,_44,_45,_46,_47,_48,_49,_50,_51,_52,_53,_54,_55,_56,_57,_58,_59,_60,_61,_62,_63,\
+            __N, ...) \
+        __N
```
thiagomacieira, line 166: Nitpick: double underscores are reserved to the compiler.

thiagomacieira (changes requested):
If I'm reading the patch correctly, in addition to the renaming of some constants (thanks, BTW), you're doing three things: you're adding `random_changins`, you're changing `set_random_bits` to produce sequences of 1s and zeroes instead of giving each bit a 50-50 chance of being 1, and you're adding a method of using the RNG engine much less.

Please split into multiple commits for ease of reviewing. I also don't think that using the RNG engine less is useful. It requires overhead in the caller, so I don't expect any test to ever use them. If we want to internally make ideal use of the two RNGs we have, then sure, let's do it, but I'd like a separate PR so we can review it, because the AES generator always generates 128 bits on every operation, so using `random32()` discards 96 useful bits, while the LCG one generates 31 bits per iteration, so asking for 32 makes it generate 62 and discard 30.

PR #130, opened 2022-07-14
FW: implement stateful timestamps for the background scan

thiagomacieira:
This allows the tool to be restarted (for example, after an update) and
retain its average CPU consumption. There's no intention in this commit
to restart the scan wherever it was interrupted, though.

This will store the timestamp file in the runtime directory, which MUST
be specified via an environment variable. systemd does that for us if
you set the RuntimeDirectory= entry in the .service file[1]. I've chosen
the RuntimeDirectory because it's erased after boot (it's on tmpfs on
Linux).

This commit does NOT implement a race condition prevention, despite use
of atomics. If two instances of OpenDCDiag are launched with --service
and with the same runtime directory, they will step on each other's
toes. Preventing that is an exercise for the upper layer, like systemd.

This commit also does NOT implement stateful storage for Windows.

[1] https://www.freedesktop.org/software/systemd/man/systemd.exec.html#RuntimeDirectory=

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -47,12 +47,12 @@
 #  include <sys/eventfd.h>
 #  include <sys/prctl.h>
 #  include <sys/types.h>
-#  include <sys/stat.h>
 #endif
 #ifdef __unix__
 #  include <sys/resource.h>
 #endif
 #include <sys/mman.h>
+#include <sys/stat.h>
 #include <unistd.h>
 
 #include <math.h>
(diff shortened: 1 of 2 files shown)


dawid-sabat (approved):
LGTM

dawid-sabat (dismissed):

thiagomacieira:
Fixed small issue compiling on FreeBSD

lukasz-tuz (approved):

PR #133, opened 2022-07-15
FW: Add openssl support for Opendcdiag

AthenasJimenez:


tests/meson.build
```diff
@@ -38,6 +38,14 @@ tests_set_base.add(
 	)
 )
 
+crypto_dep = dependency('libcrypto', version: ['>=1.0', '<3.0'], required: false)
```
thiagomacieira, line 41: This shouldn't appear twice. Just add to the top-level meson.build.

thiagomacieira (commented):
Good start. We'll need the linked mode configuration too, probably via a `framework_options` in Meson.

The new `sandstone_ssl_init()` function must be called from sandstone.cpp somewhere to initialise the state. That means we'll need a new config in sandstone_config.h to activate it.

thiagomacieira (commented):
+1

framework/logging.cpp
```diff
@@ -90,6 +90,10 @@ RtlGetVersion(
 #  include <gnu/libc-version.h>
 #endif
 
+#if __has_include(<openssl/opensslv.h>)
```
thiagomacieira, line 93: On some systems (like mine, OpenSUSE), this will be true whether we're linking to OpenSSL or not, because the openssl/ dir is directly in /usr/include:
```
$ ls -l /usr/include/openssl/opensslv.h
-rw-r--r-- 1 root root 4101 Jul  8 05:07 /usr/include/openssl/opensslv.h
```

Instead of `__has_include`, can you use `SANDSTONE_SSL_BUILD` from sandstone_config.h, like you did in sandstone.cpp?

framework/logging.cpp
```diff
@@ -2295,6 +2302,9 @@ void YamlLogger::print_header(std::string_view cmdline, Duration test_duration,
     logging_printf(LOG_LEVEL_QUIET, "command-line: '%s'\n", cmdline.data());
     logging_printf(LOG_LEVEL_QUIET, "version: " PROGRAM_VERSION "\n");
     logging_printf(LOG_LEVEL_VERBOSE(1), "os: %s\n", os_info().c_str());
+#if __has_include(<openssl/opensslv.h>)
+    logging_printf(LOG_LEVEL_VERBOSE(1), "ssl: %s\n", OPENssl_VERSION_TEXT);
```
thiagomacieira, line 2306: I'd suggest adding to `os_info()` so it gets printed like the libc version on the line above.

framework/sandstone_ssl.cpp
```diff
@@ -0,0 +1,84 @@
...
+#include <exception>
+#ifdef _WIN32
+#  include <windows.h>
+#else
+#  include <dlfcn.h>
+#endif
+
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+
+#ifndef SANDSTONE_OPENSSL_LINKED
```
thiagomacieira, line 17: This one also needs to go to sandstone_config.h, but we can work on it later.

thiagomacieira (changes requested):
Nearly there. If you fix the `__has_include`, we can merge this and move on to the next problem.

thiagomacieira (commented):
Very interesting!

What's left now?

tests/openssl/ssl_sha.cpp
```diff
@@ -0,0 +1,156 @@
...
+                "sha256sum values does not match.");
+        memcmp_or_fail(&our_elem->sha384sum[0], &golden_elem->sha384sum[0], SHA256_DIGEST_LENGTH,
+                "sha384sum values does not match.");
+        memcmp_or_fail(&our_elem->sha512sum[0], &golden_elem->sha512sum[0], SHA256_DIGEST_LENGTH,
+                "sha512sum values does not match.");
+    }
+    return EXIT_SUCCESS;
+}
+
+DECLARE_TEST(ssl_sha, "Test calculating differnt sha checksums")
```
thiagomacieira, line 152: Please name the file and the test with "openssl" so it's clear we're talking about it.

thiagomacieira (commented):
One more thing: can you add openssl-dev to the CI build (at least the Linux one) so we know this compiles?

AthenasJimenez:
@thiagomacieira I think this is it, unless you have further comments, of course :)

thiagomacieira (approved):
Yup!

I'm going to override the CI because you're not touching Windows here.

.github/workflows/build.yaml
```diff
@@ -24,11 +24,12 @@ jobs:
             libzstd-dev \
             meson \
             ninja-build \
+            openssl-dev \
```
thiagomacieira, line 27: The package isn't `openssl-dev`

Looks like the workflow that got run isn't the one in the PR, so it's now failing.

PR #134, opened 2022-08-01
[Eigen 3.4] Backward compatibility with Eigen tests

jposwiata:
Eigen 3.3.8 uses FMA feature to run calculations, while 3.4.0 is able
to use AVX512F (with main focus on vfmadd with ZMM registers). To keep
backward compatibility eigen_svd_cdouble must be kept with FMA and
new test (eigen_svd_cdouble_avx512) added when AVX512F is properly handled
(what implies at least Eigen 3.4.0 and ICX architecture).

### DIFF
--- a/tests/eigen_svd/svd_cdouble_avx512.cpp
+++ b/tests/eigen_svd/svd_cdouble_avx512.cpp
@@ -5,16 +5,17 @@
  * Copyright 2022 Intel Corporation.
  * SPDX-License-Identifier: Apache-2.0
  *
- * @test @b eigen_svd_cdouble
+ * @test @b eigen_svd_cdouble_avx512
  * @parblock
  * This piece of code aims to stress test the FMA execution units of
  * the CPU, among others, by repetitively solving the singular value
  * decomposition problem on given input matrices, which involve a lot
  * of matrix multiplication operations underneath.
  *
- * The test is intended to verify FMA feature, must be built without
- * AVX512 features enabled (t.i with Eigen pre-3.4 or without AVX512F
- * feature available).
+ * The test is intended to use AVX512 instructions, which are available
+ * with Eigen 3.4 (and up) with AVX512 enabled. The test should not be
+ * created for other configurations (Eigen 3.3.x or GCC pre-12 or below
+ * SKX target)
  *
  * The logic comes from the 3rd party library Eigen. The first thread
  * that gets to run computes a "golden value" of the results, those
@@ -30,8 +31,7 @@
  * @endparblock
  */
 
-// Test is expected *not" to test AVX2/FMA only (to follow Eigen 3.3.8)
-#define SANDSTONE_EIGEN_VECTORIZATION SANDSTONE_EIGEN_AVX2
+#define SANDSTONE_EIGEN_VECTORIZATION SANDSTONE_EIGEN_AVX512
 #include <sandstone_eigen_configurator.h>
 
 #include "sandstone_eigen_common.h"
@@ -44,11 +44,12 @@ typedef Eigen::BDCSVD < Mat > SVD;
 #define M_DIM 300               // weird dim on purpose
 
 using eigen_svd_cdouble_test = EigenSVDTest<SVD, M_DIM>;
-DECLARE_TEST(eigen_svd_cdouble, "Eigen SVD (Singular Value Decomposition) solving payload, which issues a bunch of matrix multiplies underneath, now operating on std::complex<double>")
+DECLARE_TEST(eigen_svd_cdouble_avx512, "Eigen SVD (Singular Value Decomposition) solving payload, which issues a bunch of matrix multiplies underneath, now operating on std::complex<double>")
   .groups = DECLARE_TEST_GROUPS(&group_math),
   .test_init = eigen_svd_cdouble_test::init,
   .test_run = eigen_svd_cdouble_test::run,
   .test_cleanup = eigen_svd_cdouble_test::cleanup,
+  .minimum_cpu = cpu_feature_avx512f,
   .fracture_loop_count = 5,
-  .quality_level = TEST_QUALITY_PROD,
+  .quality_level = TEST_QUALITY_BETA,
 END_DECLARE_TEST
--- a/framework/sandstone_eigen_configurator.h
+++ b/framework/sandstone_eigen_configurator.h
@@ -1,0 +1,146 @@
+/*
+ * Copyright 2022 Intel Corporation.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/*
+ * NOTE: This is a private include file of the sandstone framework to configure Eigen library
+ * and should be included only by the test using Eigen. Must be very first included file
+ * (with prior definition of SANDSTONE_EIGEN_VECTORIZATION)
+ */
+
+#ifndef SANDSTONE_EIGEN_CONFIGURATOR_H
+#define SANDSTONE_EIGEN_CONFIGURATOR_H
+
+#define SANDSTONE_EIGEN_DEFAULT      0
+#define SANDSTONE_EIGEN_AVX          1
+#define SANDSTONE_EIGEN_AVX2         2
+#define SANDSTONE_EIGEN_AVX512       3
+#define SANDSTONE_EIGEN_AVX512_DQ_ER 4
+#define SANDSTONE_EIGEN_FP16         5
+
+
+#ifdef SANDSTONE_EIGEN_VECTORIZATION
+    // disable eigen's configurator
+    #define EIGEN_DONT_VECTORIZE
+
+    // common vector operations
+    #define EIGEN_VECTORIZE
+    #define EIGEN_VECTORIZE_SSE
+    #define EIGEN_VECTORIZE_SSE2
+
+    #ifdef __SSE3__
+    #define EIGEN_VECTORIZE_SSE3
+    #endif
+    #ifdef __SSSE3__
+    #define EIGEN_VECTORIZE_SSSE3
+    #endif
+    #ifdef __SSE4_1__
+    #define EIGEN_VECTORIZE_SSE4_1
+    #endif
+    #ifdef __SSE4_2__
+    #define EIGEN_VECTORIZE_SSE4_2
+    #endif
+
+    #if (SANDSTONE_EIGEN_VECTORIZATION == SANDSTONE_EIGEN_DEFAULT)
+        #warning Default compiler handling is used
+
+    #elif (SANDSTONE_EIGEN_VECTORIZATION == SANDSTONE_EIGEN_AVX)
+        #if !defined(__AVX__)
+            #error AVX feature is not available
+        #elif !defined(__FMA__)
+            #error FMA feature is not available
+        #elif defined(__AVX2__)
+            #error AVX2 feature is available
+        #elif defined(__AVX512F__)
+            #error AVX512F feature is available
+        #else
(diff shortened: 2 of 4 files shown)


pwlazlyn (commented):
LGTM, but I am not sure if we should keep `svd_cdouble.cpp` and `svd_cdouble_noavx512.cpp`.
Do they result in different code to be generated?
I understand we probably don't want to remove a prod quality test.

jposwiata:
> LGTM, but I am not sure if we should keep `svd_cdouble.cpp` and `svd_cdouble_noavx512.cpp`. Do they result in different code to be generated? I understand we probably don't want to remove a prod quality test.

Yes, we need to keep both (here identical) tests. Later on one of them should be marked as deprecated

thiagomacieira:
I need a little more clarification.

Do you mean that Eigen 3.3.8 uses 256-bit FMA instructions, while Eigen 3.4 uses 512-bit ones?

What tests do we get when built with 3.4.0? if I'm reading this corretly, we get all three? And what's the difference between "eigen_svd_cdouble" and "eigen_svd_cdouble_noavx512" ?

Also, food for thought (new PRs):
* does it make sense to duplicate all tests with "avx512" suffixes, not just cdouble?
* does it make sense to duplicate them completely without FMA (i.e., pre-HSW, maybe just plain SSE4.2)?
* does it improve compile times to merge the multiple source files? Though that may change the amount of inlining and thus change the nature of the test.

pwlazlyn:
Also, don't merge this before we confirm it doesn't cause regressions.

jposwiata:
> Do you mean that Eigen 3.3.8 uses 256-bit FMA instructions, while Eigen 3.4 uses 512-bit ones?
3.3.8 uses FMA (with ymm's) with both "old" tests. GCC 12 keeps the pattern (no AVX512 instructions were found in the execution).
3.4 is able to use 512-bit instructions (when AVX512F is available, here for SKL builds as well).

> What tests do we get when built with 3.4.0? if I'm reading this corretly, we get all three? And what's the difference between "eigen_svd_cdouble" and "eigen_svd_cdouble_noavx512" ?
The main point is to keep "old" tests same. Locally built these are identical (need to verify with SDE official builds, either).
As discussed with Patryk, compiler version might be quite important.

> * does it make sense to duplicate all tests with "avx512" suffixes, not just cdouble?
> * does it make sense to duplicate them completely without FMA (i.e., pre-HSW, maybe just plain SSE4.2)?
For me, yes for both questions.
Need to check if/when/which Eigen handles SSE. The only "big" question here is compile time (see next point).

> * does it improve compile times to merge the multiple source files? Though that may change the amount of inlining and thus change the nature of the test.
No, such merge is not possible. Eigen is a template library (driven by compiler flags, in 3.4 "official" ones are used, in particular ` __AVX51F__`). What more, merging these won't help, as each instance is a separate object/set of objects and is "optimized" during compilation/linking (and as these objects are big.. it takes a lot of time).

jposwiata:
> Also, don't merge this before we confirm it doesn't cause regressions.

The change doesn't introduce any functional changes in existing tests (when compiled with Eigen 3.3.8).
'eigen_svd_cdouble' is explicitly built with FMA (in both 3.3.8 and 3.4.0). There no "verification" which feature is being used, so I've used conditional compilation on the version/flag to warn if "something" is wrong.
New test is "beta", so will be promoted to "prod" after the tests (when Eigen 3.4 *is* available, this is only a prework).

thiagomacieira:
> > What tests do we get when built with 3.4.0? if I'm reading this corretly, we get all three? And what's the difference between "eigen_svd_cdouble" and "eigen_svd_cdouble_noavx512" ?
> 
> The main point is to keep "old" tests same. Locally built these are identical (need to verify with SDE official builds, either). As discussed with Patryk, compiler version might be quite important.

Sorry, this didn't answer the last question. What'd the difference between the new `eigen_svd_cdouble` and the old `eigen_svd_cdouble_noavx512` ? I'd like to deprecate the old name, but since it's such a valuable test, it's likely showing up in a lot of databases, so we shouldn't do it.

> > * does it improve compile times to merge the multiple source files? Though that may change the amount of inlining and thus change the nature of the test.
> 
> No, such merge is not possible. Eigen is a template library (driven by compiler flags, in 3.4 "official" ones are used, in particular ` __AVX51F__`). What more, merging these won't help, as each instance is a separate object/set of objects and is "optimized" during compilation/linking (and as these objects are big.. it takes a lot of time).

Sorry, I didn't express myself very well.

Precisely because it's just a bunch of templates, we **could** merge them. E.g.:
```c++
using eigen_svd_test = EigenSVDTest<Eigen::BDCSVD<Matrix <float, Dynamic, Dynamic>>, 256>;
DECLARE_TEST(eigen_svd, "Eigen SVD (Singular Value Decomposition) solving payload, which issues a bunch of matrix multiplies underneath")
  .groups = DECLARE_TEST_GROUPS(&group_math),
  .test_init = eigen_svd_test::init,
  .test_run = eigen_svd_test::run,
  .test_cleanup = eigen_svd_test::cleanup,
  .fracture_loop_count = 2,
  .quality_level = TEST_QUALITY_PROD,
END_DECLARE_TEST

using eigen_svd_cdouble_noavx512_test = EigenSVDTest<Eigen::BDCSVD<Matrix<std::complex<double>, Dynamic, Dynamic>>, 300>;  // weird dim on purpose
DECLARE_TEST(eigen_svd_cdouble_noavx512, "Eigen SVD (Singular Value Decomposition) solving payload, which issues a bunch of matrix multiplies underneath, now operating on std::complex<double>")
  .groups = DECLARE_TEST_GROUPS(&group_math),
  .test_init = eigen_svd_cdouble_noavx512_test::init,
  .test_run = eigen_svd_cdouble_noavx512_test::run,
  .test_cleanup = eigen_svd_cdouble_noavx512_test::cleanup,
  .fracture_loop_count = 5,
  .quality_level = TEST_QUALITY_PROD,
END_DECLARE_TEST
```
in a single .cpp. There's a chance that the overall build time is smaller because the compiler can avoid the duplicated source parsing. But I suspect that the majority of time is spent in optimisation passes and codegen, which wouldn't benefit from this. The gains would be outweighed by the fact that a single .cpp that takes 10 minutes can't be parallelised, so the overall time is more than 5 .cpp files that each take 2 minutes.

pwlazlyn (changes requested):
I think, that maybe we don't need any changes to the `meson.build` file. We just have to make sure we don't see regressions with eigen 3.4.

`eigen_svd/svd_cdouble.cpp` is the only file that will be affected by the eigen update. Eigen internal vectorization was disabled causing AVX512 instructions to be emitted, but only in the code "around" the actual matrix multiplication. For matrix multiplication AVX/AVX2 was emitted. I think it was not intentional, but it was finding problems, so I think we should just make sure we still hit them with full AVX512 version.

If we see any regressions with full AVX512 version, I suggest to add a assembly program generated with eigen 3.3.8 (the one that has a mix of AVX2 and AVX512) as a `svd_cdouble` test. This way we can update eigen and omit the ifdefs.

thiagomacieira:
There's currently exactly one test that was compiled with AVX512:
```
tests_set_skx.add(
        when : eigen3_dep,
        if_true : files(
                'eigen_svd/svd_cdouble.cpp',
        )
)
```

Given what you've said, I don't think there's any way we can keep the same type of codegen using Eigen 3.4. If you use 3.4, you'll either get 512-bit FMA, or you won't get 512-bit registers at all.

I think we need data to show whether we need to keep a precompiled Eigen 3.3.x build of this test. And if we don't, we may be able to clean up the naming scheme too:
* no suffix: AVX2+FMA (256-bit)
* "avx512" suffix: AVX512+FMA (512-bit)
* "nofma" suffix: SSE4.2 (128-bit)

jposwiata:
> it's just a bunch of templates, we **could** merge them

I've done an experiment (with 'extern template' patterns and "real" merging the tests in a file)..
Merging any modules doesn't help significantly. In "overall" compilation I saved roughly 1s (6:39 -> 6:38).
Compilation of single file 300x300 complex<double> takes 49s, when tests sharing the same template are merged.. 49s again (object file is ~33M, assembly file 180M in both cases).
Merging with other templates/tests cause the increases (1:08/47M/232M, previous file was 23s/22M). No "overall" change in building time here as well.

It seems that any task on merging won't help us in the time at all. Compilation time is not important at all (it is done in parallel), so only linking is "an issue", but we cannot do anything with it.

tests/eigen_svd/svd_cdouble_avx512.cpp
```diff
@@ -0,0 +1,56 @@
+/**
+ * @file
+ *
+ * @copyright
+ * Copyright 2022 Intel Corporation.
+ * SPDX-License-Identifier: Apache-2.0
+ *
+ * @test @b eigen_svd_cdouble
```
pwlazlyn, line 8: `eigen_svd_cdouble` -> `eigen_svd_cdouble_avx512`

pwlazlyn (changes requested):

tests/eigen_svd/svd_cdouble_avx512.cpp
```diff
@@ -0,0 +1,56 @@
...
+ * @parblock
+ * This piece of code aims to stress test the FMA execution units of
```
markdryan, line 10: Maybe we should mention that this is an AVX_512 version of eigen_svd_cdouble.  I guess it's obvious (once you've fixed the name) but probably worth spelling out.

markdryan (commented):
So this is mostly looking good to me.  In my local testing I can see that eigen_svd_cdouble is now consistent, in that it uses AVX2 instructions when built with both Eigen 3.3 and Eigen 3.4.  Without this patch, eigen_svd_cdouble uses AVX2 instructions when built with Eigen 3.3 and AVX-512 instructions when built with Eigen 3.4, i.e., you get a totally different test depending on your build environment.  

The reason for this incidentally, is that although Eigen 3.3 supports AVX-512 code generation, it needs to be explicitly enabled by defining EIGEN_VECTORIZE_AVX512, even when building with -march=skylake-avx512.  This requirement was dropped in Eigen 3.4.  So if you build with -march=skylake-avx512 when using Eigen 3.4 you automatically get AVX512 FMA, without needing to define EIGEN_VECTORIZE_AVX512.

I believe we should allow the new AVX512 test to be built with Eigen 3.3.  To get it to work we just need to define EIGEN_VECTORIZE_AVX512.

tests/eigen_svd/svd_cdouble_avx512.cpp
```diff
@@ -0,0 +1,62 @@
...
+ * precision, complex numbers input matrices and the (divide and
+ * conquer) bi-diagonalization SVD algorithm.
+ *
+ * @note This test requires at least 2 threads to run.
+ * @endparblock
+ */
+
+// Eigen 3.3.8 requires AVX512 vectorization to be enabled.
+// Eigent 3.4.0 has AVX512 enabled by default (if appropriate architecture is requested)
+#define EIGEN_ENABLE_AVX512
```
markdryan, line 36: I don't think that this will work with Eigen 3.3.x as it will expose a bug not fixed until 3.4

https://gitlab.com/libeigen/eigen/-/commit/354baa0fb1bb6c2d137288a61339e8cf0e702ad7

This needs to be EIGEN_VECTORIZE_AVX512

which will give us AVX512F but not AVX512DQ (which doesn't work < 3.4).

thiagomacieira:
I don't think we need this.

PR #137, opened 2022-08-03
FW: remove openssl version restriction

AthenasJimenez:

### DIFF
--- a/framework/sandstone_ssl.cpp
+++ b/framework/sandstone_ssl.cpp
@@ -41,7 +41,13 @@ void sandstone_ssl_init()
     };
 #else
     void *libcrypto;
+#ifdef SHLIB_VERSION_NUMBER
+    // For openssl 1.0 and 1.1
     libcrypto = dlopen("libcrypto.so." SHLIB_VERSION_NUMBER, RTLD_NOW);
+#else
+    // For openssl 3.0 and above
+    libcrypto = dlopen("libcrypto.so." SANDSTONE_STRINGIFY(OPENSSL_SHLIB_VERSION), RTLD_NOW);
+#endif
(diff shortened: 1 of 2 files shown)


AthenasJimenez:
As a side note for future reference, several low level APIs  have been deprecated in favor of high level ones on OpenSSL release 3.0. We may need to remove low level APIs support as well in sandstone_ssl.h in the future. For more details please refer to: https://www.openssl.org/docs/man3.0/man7/migration_guide.html

framework/sandstone_ssl.cpp
```diff
@@ -41,7 +41,7 @@ void sandstone_ssl_init()
     };
 #else
     void *libcrypto;
-    libcrypto = dlopen("libcrypto.so." SHLIB_VERSION_NUMBER, RTLD_NOW);
+    libcrypto = dlopen("libcrypto.so", RTLD_NOW);
```
thiagomacieira, line 44: This won't work because on most non-development systems, the .so symlink is not present. You need to keep the SHLIB_VERSION_NUMBER.

thiagomacieira (changes requested):

thiagomacieira (approved):

thiagomacieira:
Thanks.

thiagomacieira:
> As a side note for future reference, several low level APIs have been deprecated in favor of high level ones on OpenSSL release 3.0. We may need to remove low level APIs support as well in sandstone_ssl.h in the future. For more details please refer to: https://www.openssl.org/docs/man3.0/man7/migration_guide.html

Got it. I just listed everything that existed in 1.1 that we could possibly have an interest in using in tests. If there are better APIs to achieve the same objectives, let's use them.

PR #139, opened 2022-08-04
[Floats] Fixing #defs to keep consistent pattern over all types

jposwiata:
All #defs follows the same pattern of names (to avoid further confusion).
Remved "general NaN" from floats (to keep consistent with the definition).
Some common macros added.
Random functions kept in single place (alloc() moved a bit).

This is a part of '[rng_bits]' PR (https://github.com/opendcdiag/opendcdiag/pull/129), to keep the task tidy.

### DIFF
--- a/framework/sandstone.h
+++ b/framework/sandstone.h
@@ -145,6 +145,41 @@ extern "C" {
 
 #define END_DECLARE_TEST   };
 
+/// Variadic macros to count the number of arguments passed to other macro.
+/// @see OVERLOAD()
+#define NARGN(\
+            n1, n2, n3, n4, n5, n6, n7, n8, n9, n10,n11,n12,n13,n14,n15,n16,n17,n18,n19,n20,n21,n22,n23,n24,n25,n26,n27,n28,n29,n30,n31,\
+            n32,n33,n34,n35,n36,n37,n38,n39,n40,n41,n42,n43,n44,n45,n46,n47,n48,n49,n50,n51,n52,n53,n54,n55,n56,n57,n58,n59,n60,n61,n62,n63,\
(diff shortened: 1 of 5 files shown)


framework/sandstone.h
```diff
@@ -145,6 +145,42 @@ extern "C" {
...
+            63,62,61,60,59,58,57,56,55,54,53,52,51,50,49,48,47,46,45,44,43,42,41,40,39,38,37,36,35,34,33,32,\
+            31,30,29,28,27,26,25,24,23,22,21,20,19,18,17,16,15,14,13,12,11,10, 9, 8, 7, 6, 5, 4, 3, 2, 1, 0)
+
+#define OVERLOAD_(name,num) name##num
+#define OVERLOAD(name,num) OVERLOAD_(name,num)
+/** @} */
+
+#define IS_ALIGNED(ptr, a) ((((uint64_t) (ptr)) & ((a) - 1)) == 0)
+
+#define MEMBER_SIZE(type, member) sizeof(((type *)0)->member)
```
thiagomacieira, line 180: Not valid for C++, and doubtful it's valid even for C. But for C++ you would never write this, you'd just write `sizeof(type::member)`, so no problem.

thiagomacieira (approved):

thiagomacieira (approved):

framework/sandstone.h
jposwiata: It was intended to be used in C only (and sandstone.**h** suggest C language, while it is a mix of C/C++).
BTW. Removed, as is not used (was a part of RationalNumber used in matrix_inverse, which is not a part of any commit).

PR #140, opened 2022-08-12
Silencing false positive warnings from writev

pwlazlyn:
GCC 12 incorrectly reports reads outside boundry of an object when we
call writev with IoVec wrapper. We get warning as if it was a single
object and not an array:

    ../opendcdiag/framework/sandstone_iovec.h:39:18: warning: ‘ssize_t writev(int, const iovec*, int)’ reading 48 bytes from a region of size 16 [-Wstringop-overread]
       39 |     return writev(fd, vec, std::size(vec));
          |            ~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~
    ../opendcdiag/framework/sandstone_iovec.h:16:8: note: source object ‘{anonymous}::IoVec::<anonymous>’ of size 16
       16 | struct IoVec : iovec
          |        ^~~~~
    In file included from ../opendcdiag/framework/sandstone_p.h:24,
                     from ../opendcdiag/framework/sandstone.cpp:64:
    /usr/include/sys/uio.h:52:16: note: in a call to function ‘ssize_t writev(int, const iovec*, int)’ declared with attribute ‘access (read_only, 2, 3)’
       52 | extern ssize_t writev (int __fd, const struct iovec *__iovec, int __count)
          |                ^~~~~~

This commit provides a simple wrapper that ensures there is enough space
by using size deduction from builtin arrays and silences the warning on
the glibc boundary call.

### DIFF
--- a/framework/logging.cpp
+++ b/framework/logging.cpp
@@ -796,9 +796,9 @@ void logging_printf(int level, const char *fmt, ...)
     if (msg.empty())
         return;     // can happen if fmt was "%s" and the string ended up empty
 
-    IoVec vec[] = {
-        indent_spaces(),
-        std::string_view(msg)
+    iovec vec[] = {
+        IoVec(indent_spaces()),
+        IoVec(std::string_view(msg))
     };
 
     if (level <= sApp->verbosity && file_log_fd != real_stdout_fd) {
@@ -814,7 +814,7 @@ void logging_printf(int level, const char *fmt, ...)
     std::string timestamp;
     if (current_output_format() != SandstoneApplication::OutputFormat::yaml) {
         timestamp = log_timestamp();
-        vec[0] = std::string_view(timestamp);
+        vec[0] = IoVec(std::string_view(timestamp));
     }
 
     int fd = file_log_fd;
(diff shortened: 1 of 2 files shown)


thiagomacieira (changes requested):
There's a better way that doesn't involve `#pragma`. Let me find it.

thiagomacieira (changes requested):
See https://codereview.qt-project.org/c/qt/qtbase/+/409083/3..4

The trick was to use a function called `IoVec` that returns `struct iovec`, instead of a type called `IoVec`. 

You can copy the trick because it's copyright Intel (I wrote it). The similarity is not a coincidence.

pwlazlyn:
> See https://codereview.qt-project.org/c/qt/qtbase/+/409083/3..4

I thought about this, but It requires additional changes because we were using `IoVec` as a type to declare arrays so we could use ctor for nice syntax. I don't think we can omit the wrapping of each array element in `IoVec()` now.

That being said, I don't really care. Both works for me.

pwlazlyn:
> See https://codereview.qt-project.org/c/qt/qtbase/+/409083/3..4

btw, I think It's use-after-free, unless qt uses some garbage collection internally.
```cpp
static struct iovec IoVec(const char *str)
{
    return IoVec(std::string(str));
}
```

thiagomacieira:
> > See https://codereview.qt-project.org/c/qt/qtbase/+/409083/3..4
> 
> btw, I think It's use-after-free, unless qt uses some garbage collection internally.

That's why Qt uses:
```c++
static struct iovec IoVec(const char *str)
{
    struct iovec r = {};
    r.iov_base = const_cast<char *>(str);
    r.iov_len = strlen(str);
    return r;
}
```

pwlazlyn:
> > > See https://codereview.qt-project.org/c/qt/qtbase/+/409083/3..4
> > 
> > 
> > btw, I think It's use-after-free, unless qt uses some garbage collection internally.
> 
> That's why Qt uses:
> 
> ```c++
> static struct iovec IoVec(const char *str)
> {
>     struct iovec r = {};
>     r.iov_base = const_cast<char *>(str);
>     r.iov_len = strlen(str);
>     return r;
> }
> ```

We only take `const char*` and `std::string_view`, so is it ok to merge it as is?

thiagomacieira (approved):

PR #143, opened 2022-08-12
FW: sandstone_ssl add more EVP APIs

AthenasJimenez:


AthenasJimenez:
Seems I missed a bunch of EVP APIs, adding them for future tests.

thiagomacieira:
Looks like those functions are not present in the build in question. Either they're too new or the opposite: they're so old that they've got deprecated for security reasons and disabled. Do we need IDEA, MD2 and whatever that RC5 function was?

AthenasJimenez:
Yeah, I'll disable them for now. If we need them some day, I can figure out what to then.

lukasz-tuz (approved):

PR #145, opened 2022-09-05
Fixed random test selection issue with --1sec/2min/5min options

brad-kelly:
Test selection for some options like --1sec/--2min/--5min was happening
in a non-random order.  This will restore the randomness to the tests
chosen to execute when using these options.

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -1195,9 +1195,7 @@ Common command-line options are:
                   runs tests 6 through 10 from the file mytests.list.
      See User Guide for more details.
  --test-list-randomize
-     Randomizes the order in which tests specified with the --test-list-file
-     option are run.  This option is ignored if --test-list-file is not
-     provided.
+     Randomizes the order in which tests are executed.
  --test-delay <time in ms>
      Delay between individual test executions in milliseconds.
  --schedule-by <selection>
@@ -3386,21 +3384,25 @@ int main(int argc, char **argv)
             return EXIT_SUCCESS;
         case one_sec_option:
             test_list_randomize = true;
+            test_selection_strategy = Repeating;
             sApp->use_strict_runtime = true;
             sApp->endtime = calculate_wallclock_deadline(1s, &sApp->starttime);
             break;
         case thirty_sec_option:
             test_list_randomize = true;
+            test_selection_strategy = Repeating;
             sApp->use_strict_runtime = true;
             sApp->endtime = calculate_wallclock_deadline(30s, &sApp->starttime);
             break;
         case two_min_option:
             test_list_randomize = true;
+            test_selection_strategy = Repeating;
             sApp->use_strict_runtime = true;
             sApp->endtime = calculate_wallclock_deadline(2min, &sApp->starttime);
             break;
         case five_min_option:
             test_list_randomize = true;
+            test_selection_strategy = Repeating;
             sApp->use_strict_runtime = true;
             sApp->endtime = calculate_wallclock_deadline(5min, &sApp->starttime);
             break;
@@ -3430,6 +3432,13 @@ int main(int argc, char **argv)
             return EX_USAGE;
         }
     }
+
+    if (test_list_randomize) {
+        if ((test_selection_strategy == Ordered) || (test_selection_strategy == Alphabetical)) {
+            test_selection_strategy = NonRepeating;
+        }
+    }
+
 
     if (SandstoneConfig::RestrictedCommandLine) {
         // Default options for the simplified OpenDCDiag cmdline
@@ -3539,11 +3548,6 @@ int main(int argc, char **argv)
                                                        starting_test_number, ending_test_number,
                                                        test_list_randomize);
     } else {
-        if (test_list_randomize) {
-            logging_printf(LOG_LEVEL_QUIET, "# WARNING: --test-list-randomize used without "
-                                            "--test-list-file. Ignored.\n");
-        }
-
         if (use_builtin_test_list) {
             if (!SandstoneConfig::RestrictedCommandLine && test_list.size()) {
                 logging_printf(LOG_LEVEL_QUIET,
@@ -3554,10 +3558,11 @@ int main(int argc, char **argv)
         } else {
             generate_test_list(test_list);
         }
-
-        if (!test_selector)
+        if (!test_selector) {
+            weighted_run_info weights[] { {NULL} };
             test_selector = setup_test_selector(test_selection_strategy, weighted_testrunner_runtimes,
-                                                std::move(test_list), nullptr);
+                                                std::move(test_list), weights);
+        }
     }
 
 #if SANDSTONE_SSL_BUILD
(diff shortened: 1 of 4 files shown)


framework/sandstone.cpp
```diff
@@ -3386,21 +3386,25 @@ int main(int argc, char **argv)
             return EXIT_SUCCESS;
         case one_sec_option:
             test_list_randomize = true;
+            test_selection_strategy = Repeating;
             sApp->use_strict_runtime = true;
```
markdryan, line 3390: I guess these aren't needed any more as they will be set in the if statement below.

markdryan (dismissed):
I see a warning in the logs

```
 ./builddir/opendcdiag --1sec --test-list-randomize -o -
# WARNING: --test-list-randomize used without --test-list-file. Ignored.
command-line: 'opendcdiag --1sec --test-list-randomize -o -'
```

Presumably we want to remove that log now as using --test-list-randomize without --test-list-file is a valid thing to do.

markdryan:
./builddir/opendcdiag --test-list-randomize

doesn't run any tests, just mce_check.  Is that intended, or is that not a valid combination?

markdryan:
Also, the documentation is incorrect

```
 --test-list-randomize
     Randomizes the order in which tests specified with the --test-list-file
     option are run.  This option is ignored if --test-list-file is not
     provided.
```

When is it valid to use --test-list-randomize?

markdryan:
Note I also see the warning if I don't explicitly specify --test-list-randomize

```
./builddir/opendcdiag -v  --1sec
# WARNING: --test-list-randomize used without --test-list-file. Ignored.
command-line: 'opendcdiag -v --1sec'
```

which is even more confusing, as I didn't specify it.

markdryan:
./builddir/opendcdiag -v --test-list-file test-list.txt  --30sec

now runs the tests in the test file in random order.  Is this intended.  The documentation for --30sec doesn't mention random order of tests.

framework/sandstone.cpp
brad-kelly: This may appear redundant but the clause below is a guard clause to prevent ordered and alphabetical test ordering when the user explicitly asked for random.  The lines above are to determine what kind of random strategy will be used for these specific modes.  it is possible that we may change these modes to be NonRepeat while desiring the guard clause below to still be Repeat.

Now that I think about it, the default for test-list-randomize should probably be NonRepeat while --Xmin should probably be repeat.  this is based on the idea that --test-list-randomize would imply running the test list from start to finish in a random order, while the others are more of a weighted run list with no real end.

markdryan: I was referring to sApp->use_strict_runtime = true;

markdryan:
@brad-kelly I guess my concern here is that this change modifies the situations in which --test-list-randomize can be used without updating the documentation.

The documentation says

>  --test-list-randomize
     Randomizes the order in which tests specified with the --test-list-file
     option are run.  This option is ignored if --test-list-file is not
     provided.

And that's true in the existing code I think.  But now, with your change, --test-list-randomize will not be ignored if it is used without --test-list-file.  It will have an effect.  With this patch, strict_runtime will be set to true and the ordering will be changed from alphabetical to repeating.

There's also the issue with the erroneous log reported above, but this is partially a pre-existing bug.

framework/sandstone.cpp
```diff
@@ -3431,6 +3435,14 @@ int main(int argc, char **argv)
         }
     }
 
+    if (test_list_randomize) {
+        sApp->use_strict_runtime = true;
```
thiagomacieira, line 3439: This shouldn't be implied just because it randomises.

thiagomacieira (changes requested):

framework/sandstone.cpp
brad-kelly: This is still needed, otherwise the test list will continue to run to completion before it stops.
Without this option, the main loop will not check the duration often enough to end the run in the appointed time.

framework/sandstone.cpp
brad-kelly: Agreed - and now removed.

brad-kelly:
Mark - thanks for being so thorough :-)

Just did an update that:
- Removed the warning mark noticed
- fixes the case with only --test-list-randomize option (now runs all tests in random order)
- Updates the help message for thie --test-list-randomize option
- Removed warning with --1sec option
- 

Also, all the --1sec/30sec/2min/5min options are intended to run a subset of the tests in random order for the time requested.

brad-kelly:
For the record here is the matrix of behaviors that will change based on command line options:

<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:dt="uuid:C2F41010-65B3-11d1-A29F-00AA00C14882"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=OneNote.File>
<meta name=Generator content="Microsoft OneNote 15">
</head>

<body lang=en-US style='font-family:Calibri;font-size:11.0pt'>


<div style='direction:ltr'>



Options (-v   implied) | Previous behavior | New Behavior
-- | -- | --
--test-list-file   3tests.list --test-list-randomize | Random order.   Stops at end of   test list | NO CHANGE.
--test-list-file   3tests.list --test-list-randomize -T1m | Runs tests   repeatedly in same order.   same repeated   scrambled list until end.   Stops at end of   test list. | NO CHANGE.
--30sec/--1min/--5min | Runs tests in alphabetical order.   Ends at time   limit. | Runs random order.   Ends at time   limit.
-e adox -e   adcxlong -e x87 --test-list-randomize | Runs tests in alphabetical order.   Prints warning.   Stops at end of   list. | Runs random order.   Stops at end of   list.
-e adox -e   adcxlong -e x87 --test-list-randomize -T1m | Runs tests in alphabetical order repeatedly.   Prints warning.   Stops at ~1   minute. | Runs random order repeatedly.   Stops at ~1   minute.
--use-builtin-test-list   --test-list-randomize | Run alphabetical. | Random order.
-v   --test-list-randomize | Runs alphabetical. | Random order.



</div>


</body>

</html>

markdryan:
@brad-kelly Thanks for the table.  That really helps.  Looks good to me now but you'll need to sign off your commit to allow the CI to run.

thiagomacieira (commented):
Looks good. Not sure how to test the randomisation in BATS...

thiagomacieira (approved):

brad-kelly:
Already have a test ready for it.

I run the tests for 30 seconds (--30sec) then collect the test names run and make sure they are NOT sorted by sorting it and comparing.

extract_tests() {
    grep "\- test:" $1 | cut -d : -f 2 | uniq  | xargs
}

sort_words() {
    echo $1 | xargs -n1 | sort | xargs
}

@test "Ensure --30sec runs in 30 seconds and in random order" {
    run_sandstone_yaml $disables --30sec
    total=`extract_timing output.yaml`
    order=`extract_tests output.yaml`
    sorted=`sort_words "$order"`

    # [[ "$total" -lt 32000 ]]
    [[ "$order" != "$sorted" ]]
}

busykai (approved):
LGTM.

thiagomacieira:
This doesn't appear to work for the selftests:
```
$ ./opendcdiag -o - --selftests --test-list-randomize | yq '.tests[].test'
"mce_check"
```

PR #146, opened 2022-09-06
FW: use the non-coarse monotonic clock to calculate frequencies

thiagomacieira:
The coarse clock can return the same time twice:
```
(gdb) p before.ts
$1 = {tv_sec = 4093053, tv_nsec = 275360858}
(gdb) p after.ts
$2 = {tv_sec = 4093053, tv_nsec = 275360858}
```
Which would result in a division by zero (the other values in the structure prove that those are two different snapshots).

Additionally, save the time point in std::chrono form so we can more easily compare for equality, in case the Linux clock source did return the same value twice. Floating point divisions by zero on x86 as usually masked, so we'd get normally only get a NaN, which is why this hasn't been a problem.

[ChangeLog][Framework] Improved the precision of the calculation of the effective CPU frequency on x86.

### DIFF
--- a/framework/sysdeps/linux/effective_cpu_freq.hpp
+++ b/framework/sysdeps/linux/effective_cpu_freq.hpp
@@ -10,9 +10,9 @@
 #  include "../generic/effective_cpu_freq.hpp"
 #else
 
+#include "sandstone_p.h"
 #include <limits>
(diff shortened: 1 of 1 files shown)


thac0 (approved):
👍 lgtm

PR #150, opened 2022-09-09
Test selectors cleanup

thiagomacieira:
* Remove unused virtual function
* Replace macro with explicit C++ Code

### DIFF
--- a/framework/test_selectors/WeightedSelectorBase.cpp
+++ b/framework/test_selectors/WeightedSelectorBase.cpp
@@ -5,8 +5,6 @@
 
 #include "WeightedSelectorBase.h"
 
-#define KEY_EXISTS(m, key)    ( (m).find((key)) != (m).end())
-
 using namespace std;
 
 void WeightedTestrunSelector::general_setup_from_structs(weighted_run_info *weights, WeightedTestLength length_adjust) {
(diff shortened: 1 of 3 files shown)


framework/test_selectors/WeightedSelectorBase.cpp
```diff
@@ -22,8 +20,8 @@ void WeightedTestrunSelector::adjust_runinfo() {
 }
 
 int WeightedTestrunSelector::lookup_test_index(const char * test_id_string)  {
-    if (KEY_EXISTS(testid_to_index_map, test_id_string)) {
-        return testid_to_index_map[test_id_string];
+    if (auto it = testid_to_index_map.find(test_id_string); it != testid_to_index_map.end()) {
+        return it->second;
```
brad-kelly, line 24: This is a lot less readable.  I don't see how this is an improvement - besides avoiding one extra access to the structure.  To save 0.0000001 seconds of machine time you may end up costing a lot more human time.  

I prefer to optimize for readability unless you can prove that this is perf critical code since humans are the slow part that cost a lot of money.

thiagomacieira: It's not performance critical, but this is silly. Looking up twice means calling `strlen` twice to get the length of the `const char *`, then allocating memory inside the `std::string`, then hashing it, then looking up. Sorry, won't revert.

brad-kelly: Fine - I won't be the immovable object :-P

brad-kelly (approved):

thac0 (approved):
👍

PR #151, opened 2022-09-09
FW/Unix: fix fd leak in case we decide not to exec the fallback

thiagomacieira:


markdryan (approved):
LGTM

PR #152, opened 2022-09-09
FW: make --test-list-file ignore skip-quality tests

thiagomacieira:
Instead of printing an error. This allows us to be backwards-compatible with test list files that contain names we've retired. It also makes --test-file-list follow the same behaviour as -e.

Added tests for both.


brad-kelly (approved):
This all looks good to me.

You even cleaned up some of my code to make it more C++ ish and less pythonic :-)

busykai (approved):

PR #162, opened 2022-10-03
Update simd.conf

thiagomacieira:
* Added AMD Zen 3 and 4 (the latter of which has AVX512)
* Added Intel 13th Generation Core (Raptor Lake), Meteor Lake, Granite Rapids, Sierra Forest and Grand Ridge processors
* Added CPUID features `raoint`, `cmpccxadd`, `amxfp16`, `avxvnniint8`, `avxneconvert`,`avxifma` and `prefetchiti`
* Enabled detection of CPUID features `raoint`, `cmpccxadd`, `amxfp16`

Plus reorganised the detection to list the P-cores by name (the "Coves") going back to Cannon Lake / Palm Cove. So, for example, if you run `opendcdiag --dump-cpu-info` on a regular Alder Lake, you should see "Detected CPU: Gracemont", while the parts with AVX512 enabled will probably list "Golden Cove" (I have no way of testing).


thac0 (approved):
👍

PR #163, opened 2022-10-03
IFS test: Add support for multiple test images

AthenasJimenez:
Starting from kernel 6.2, In-field Scan (aka IFS) feature will support
multiple test images that can be load and run independently between them.
IFS test would load and run one of the available images on each core.

This commit also includes interactive options:

- test_file: Specify Image ID number. If ID does not exist, default ID (0x1)
  would be use. Values accepted: any unsigned integer.
- enforce_run: When previous IFS run failed, next IFS run would refuse to run,
  unless enforced. Values accepted: 1

### DIFF
--- a/tests/ifs/ifs.c
+++ b/tests/ifs/ifs.c
@@ -9,7 +9,7 @@
  *
  * Requires `ifs.ko` to be loaded in the Linux kernel, as well as
  * firmware test blob data in `/lib/firmware/...`. Supported since
- * 5.17.0.
+ * 6.2
  *
  */
 
@@ -28,6 +28,7 @@
 #include <unistd.h>
 
 #define PATH_SYS_IFS_BASE "/sys/devices/virtual/misc/"
+#define DEFAULT_TEST_ID 1
 
 #define BUFLEN 256 // kernel module prints at most a 64bit value
 
@@ -39,6 +40,11 @@
  */
 #define IFS_SW_TIMEOUT                          0xFD
 #define IFS_SW_PARTIAL_COMPLETION               0xFE
+
+typedef struct {
+    char image_id[BUFLEN];
+    char image_version[BUFLEN];
+} ifs_test_t;
 
 static bool is_result_code_skip(unsigned long long code)
 {
@@ -65,12 +71,8 @@ static bool write_file(int dfd, const char *filename, const char* value)
         return true;
 }
 
-static ssize_t read_file(int dfd, const char *filename, char buf[static restrict BUFLEN])
+static ssize_t read_file_fd(int fd, char buf[static restrict BUFLEN])
 {
-        int fd = openat(dfd, filename, O_RDONLY | O_CLOEXEC);
-        if (fd < 0)
-            return fd;
-
         ssize_t n = read(fd, buf, BUFLEN);
         close(fd);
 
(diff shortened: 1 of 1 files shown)


tests/ifs/ifs.c
```diff
@@ -28,6 +28,7 @@
 #include <unistd.h>
 
 #define PATH_SYS_IFS_BASE "/sys/devices/virtual/misc/"
+#define DEFAULT_TEST_ID 1
```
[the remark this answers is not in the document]
AthenasJimenez: It feels weird right? but no, batches goes from "1" to "N".

tests/ifs/ifs.c
```diff
@@ -82,8 +88,71 @@ static ssize_t read_file(int dfd, const char *filename, char buf[static restrict
...
+    sprintf(ifs_info->image_id, "%x", next_test);
+    if (write_file(dfd, "current_batch", ifs_info->image_id))
+    {
+        return true;
+    }
+    else if (errno == ENOENT)
+    {
+        /* when next blob does not exists, it will fail with
+         * ENOENT error. Then, start from the begining */
+        log_info("Test file %s, does not exists. Starting over from 0x01", ifs_info->image_id);
```
thiagomacieira, line 139: Unfortunately, this isn't defensive enough. We need to differentiate an ENOENT from openat() and the one from write(). If open fails, it means the sysfs entry doesn't exist (this ifs type doesn't support batches), while write() failing indicates the kernel driver does support batches but the kernel failed to find the required file (error passed through from `request_firmware()`).

tests/ifs/ifs.c
```diff
@@ -110,25 +179,49 @@ static int scan_init(struct test *test)
...
         if (ifs0 < 0 || faccessat(ifs0, "run_test", R_OK, 0) < 0 ||
-                faccessat(ifs0, "status", R_OK, 0) < 0 || faccessat(ifs0, "details", R_OK, 0) < 0) {
+                faccessat(ifs0, "status", R_OK, 0) < 0 ||
+                faccessat(ifs0, "details", R_OK, 0) < 0 ||
+                faccessat(ifs0, "image_version", R_OK, 0) < 0 ||
+                faccessat(ifs0, "current_batch", R_OK, 0) < 0) {
                 log_info("not supported (missing kernel module, firmware, or invalid HW)");
                 close(ifs0);
                 return -EOPNOTSUPP;
         }
```
thiagomacieira, line 191: Maybe remove the `faccessat` portions and leave only the check for `ifs0` file descriptor, which case remove the `close(ifs0)` too. Once the kernel API is set in stone, we're not going to see multiple versions of the driver any more.

tests/ifs/ifs.c
```diff
@@ -110,25 +179,49 @@ static int scan_init(struct test *test)
...
                 return -saved_errno;
         }
 
+        int batch_fd = openat(ifs0, "current_batch", O_WRONLY);
+        saved_errno = errno;
+        if (batch_fd < 0) {
+                log_info("could not open intel_ifs_0/current_batch for writing (not running as root?): %m");
+                close(batch_fd);
+                return -saved_errno;
+        }
```
thiagomacieira, line 208: Don't add this. Let's be flexible.

thiagomacieira (changes requested):
We probably need to change the `ifs_init` function to iterate over the paths in the sysfs and update the images for the `intel_ifs_1`, `intel_ifs_2`, etc.

tests/ifs/ifs.c
thiagomacieira: Changed my mind.

tests/ifs/ifs.c
```diff
@@ -110,25 +182,38 @@ static int scan_init(struct test *test)
...
+        close(run_fd);
+        close(batch_fd);
+
+
+        /* load test file */
+        if (!load_test_file(ifs0, test, ifs_info))
```
thiagomacieira, line 206: I'd say that since you've opened the batch file, you should simply pass the file descriptor here. You'd need to split `read_file` into two functions, one `read_file_fd` and the other that opens. There's no need for `write_file_fd`, that's just `write`

tests/ifs/ifs.c
```diff
@@ -82,8 +88,74 @@ static ssize_t read_file(int dfd, const char *filename, char buf[static restrict
...
+{
+    char current_buf[BUFLEN], status_buf[BUFLEN];
+    int next_test, current_test;
+
+    /* read both files status and current_batch */
+    read_file(dfd, "status", status_buf);
+    read_file(dfd, "current_batch", current_buf);
+
+    /* when previous run has a status of fail, skip test */
+    if (memcmp(status_buf, "fail", strlen("fail")) == 0)
```
thiagomacieira, line 101: I wonder: should we have a test knob to force running?

tests/ifs/ifs.c
```diff
@@ -82,8 +88,74 @@ static ssize_t read_file(int dfd, const char *filename, char buf[static restrict
...
+    sprintf(ifs_info->image_id, "%x", next_test);
+    if (write_file(dfd, "current_batch", ifs_info->image_id))
+    {
+        return true;
+    }
+    else if (errno == ENOENT)
+    {
+        /* when next blob does not exists, it will fail with
+         * ENOENT error. Then, start from the begining */
+        log_info("Test file %s, does not exists. Starting over from 0x01", ifs_info->image_id);
```
thiagomacieira, line 142: Quick suggestion: pass `DEFAULT_TEST_ID` as a parameter to `log_info` instead of hardcoding `0x01`

tests/ifs/ifs.c
```diff
@@ -82,8 +88,74 @@ static ssize_t read_file(int dfd, const char *filename, char buf[static restrict
...
+                log_info("Test file %s remains untested, so try again", current_buf);
+                next_test = current_test;
+            }
+            else
+                next_test = current_test + 1;
+        }
+    }
+
+    /* write next test file ID */
+    sprintf(ifs_info->image_id, "%x", next_test);
```
thiagomacieira, line 133: Either print in decimal (%d) or add the prefix (`0x%x` or `%#x`), otherwise the kernel will get confused for `next_test >= 10`. Repeats below, though below it's not harmful.

thiagomacieira (changes requested):

tests/ifs/ifs.c
AthenasJimenez: I think we should, do we have a get_testspecific_knob_value_bool() by any chance? Would be useful only add a "force" option

thiagomacieira: We can either add that or just use integers. `force=1`

Option only allowed if `$USER == "luke"` :-)

tests/ifs/ifs.c
```diff
@@ -82,8 +84,83 @@ static ssize_t read_file(int dfd, const char *filename, char buf[static restrict
...
+                next_test = current_test;
+            }
+            else
+                next_test = current_test + 1;
+        }
+    }
+
+    /* write next test file ID */
+    sprintf(ifs_info->image_id, "%#x", next_test);
+    if (write_file(dfd, "current_batch", ifs_info->image_id))
```
thiagomacieira, line 139: The call to `read_file_fd` closed the `batch_fd`, but you could have kept it open and used it here.

thiagomacieira (approved):

tests/ifs/ifs.c
AthenasJimenez: hahaha "I am one with the Force and the Force is with me".

I added the "enforce_run" knob so we can run, even if previous run fails.
Regarding to the write, I'd like to leave it for later and revamp the whole test.

thiagomacieira (approved):

PR #165, opened 2022-10-05
Switch to C++20 mode for C++ code

thiagomacieira:
This asks the C++ compiler to use C++20 for our code and it uses `std::span` replacing our home-grown copy.

libstdc++ got `std::span` for the GCC 10 release and libc++ got it for LLVM 7, though I cannot speak for the quality of those early implementations, prior to the 2020 edition to the standard was even published. This change was only tested with GCC 12.


thiagomacieira:
Yeah, I didn't test the unit tests.

pwlazlyn (commented):
`memory_order` changes make sense, but as you've said in the commit message - all of them can be relaxed, I think.

framework/sandstone.cpp
```diff
@@ -2352,7 +2352,7 @@ static void list_tests(int opt)
     auto groups = collate_test_groups();
     int i = 0;
 
-    for (struct test *test = test_set.begin(); test != test_set.end(); ++test) {
+    for (auto test = test_set.begin(); test != test_set.end(); ++test) {
```
pwlazlyn, line 2355: This could also be changed to range based for loop.

thiagomacieira: Indeed, but I went for the minimal change here. Changing to a for-range loop would change the body of the for to use `test.` instead of `test->`

PR #166, opened 2022-10-07
FW: kvm: Fix a few issues with the kvm framework code

markdryan:
These were noticed during a review of the code.

- Fix a potential file descriptor leak
- Don't madivse NULL
- Reset file descriptors and memory pointers to their zero state when we recreate the VM
- Fix an indentation issue
- Add a missing error check.


thac0 (approved):
👍

PR #167, opened 2022-10-11
Allow instrumenting TEST_LOOP

thiagomacieira:
This adds `test_loop_start` and `test_loop_end` symbols so one can extract the looping of a given OpenDCDiag-driven test. They complement `test_start` and `test_end`, but only work for tests that use `TEST_LOOP`.

All four functions now mark their existence by way of an expensive no-op of `CMPXCHG EAX, EAX`. The RAX register will contain the main type "TESTLOOP" or "TESTMAIN" string, and the EDX one will have the state ("STRT" or "END\0"). Both the instruction and the constants are very distinctive. Doing this was inspired on <valgrind.h>, which communicates with the Valgrind by an expensive NOP of rotating RDI left four times, adding up to 64 bits.

### DIFF
--- a/framework/sandstone.h
+++ b/framework/sandstone.h
@@ -104,15 +104,6 @@ extern "C" {
         auto __return_value__  = call;          \
         (void)__return_value__;                 \
     })
-/// to be used in the test_run function of a test.  TEST_LOOP executes its body continuously until it is
-/// asked to terminate by the OpenDCDiag framework. The second parameter, N, specifies the
-/// granularity of the loop.  On each iteration of the loop, the body is executed N times.  The
-/// framework then checks to see if the test's time slot has elapsed.  If it has, the loop terminates.
-/// If it has not, the body is executed another N times before another check is made.
-/// By convention, the second parameter to TEST_LOOP is always a power of two.
-#define TEST_LOOP(test, N)                      \
-    static_assert(N > 0, "N must be positive"); \
-    for (int _loop_i_ = 0; _loop_i_ < N || (_loop_i_ = 0, test_time_condition(test)); ++_loop_i_)
 
 #else
 #define IGNORE_RETVAL(call)                     \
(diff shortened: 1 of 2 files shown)


framework/sandstone.h
```diff
@@ -103,27 +104,27 @@ extern "C" {
...
+/// to be used in the test_run function of a test.  TEST_LOOP executes its body continuously until it is
+/// asked to terminate by the OpenDCDiag framework. The second parameter, N, specifies the
+/// granularity of the loop.  On each iteration of the loop, the body is executed N times.  The
+/// framework then checks to see if the test's time slot has elapsed.  If it has, the loop terminates.
+/// If it has not, the body is executed another N times before another check is made.
+/// By convention, the second parameter to TEST_LOOP is always a power of two.
+#define TEST_LOOP(test, N)                          \
+    static_assert(N > 0, "N must be positive");     \
+    test_loop_start();                              \
+    for (int _loop_i_ = 0; _loop_i_ == 0; test_loop_end(), ++_loop_i_)          \
```
thiagomacieira, line 125: This isn't right for a `break` in the inner loop.

markdryan:
Does it matter whether the instructions are executed or not or just whether they are in the binary?  I ask as test_loop_end probably won't get called if there's a test failure.  I think is different in behaviour to test_end which does get called on test failure, as far as I can tell (unless we ud2 on failure).  If execution of test_loop_end is important we might want to mention its behaviour on test failure in the comments in sandstone.h (or possibly arrange for it to be called if it makes sense to do so).

thiagomacieira:
> Does it matter whether the instructions are executed or not or just whether they are in the binary? I ask as test_loop_end probably won't get called if there's a test failure. I think is different in behaviour to test_end which does get called on test failure, as far as I can tell (unless we ud2 on failure). If execution of test_loop_end is important we might want to mention its behaviour on test failure in the comments in sandstone.h (or possibly arrange for it to be called if it makes sense to do so).

I used them as a marker, in case you're reading an instruction dump without executable symbols. That means they have to be executed. I'd hope that someone looking at the looks already knows whether the test passed or not. Will update the docs.

fenrus75 (approved):

PR #168, opened 2022-10-13
Fix the behavior of --test-delay option.

busykai:
Unlike background scan mode, --test-delay is an interactive option. If a test is skipped, the framework will constantly attempt to restart it without any delay. Consider the following command-line:

$ opendcdiag -vv -T 24h -e ifs --test-delay 30000

Instead of attempting to run the test every 5 minutes, the above will be constantly trying to restart the test (if it was skipped the first time or when it will be skipped later).

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -3599,8 +3599,8 @@ int main(int argc, char **argv)
             tc = 0;
             logging_print_iteration_start();
             initialize_smi_counts();  // used by smi_count test
-        } else if (lastTestResult != TestSkipped) {
-            if (sApp->service_background_scan)
+        } else {
+            if (sApp->service_background_scan && lastTestResult != TestSkipped)
                 background_scan_wait();
             else
                 wait_delay_between_tests();


thiagomacieira:
Like I said, I don't think this is right.

busykai:
> Like I said, I don't think this is right.

I understood that you made it so intentional, but I don't understand the why. I can see the argument for the background scan option, but not for when it's interactive.

AthenasJimenez:
While I agree that interactive/custom delay (when set) should be prioritized over background delay, this change looks a little weird to me. Instead I'd go with something like this:

`
// Assuming test_delay_option can be used here.
3602         } else if (lastTestResult != TestSkipped) {
3603             if (sApp->service_background_scan  && !test_delay_option )
3604                 background_scan_wait();
3605             else
3606                 wait_delay_between_tests();
3607         }
`

busykai:
@AthenasJimenez your code will retry the skipped test immediately, and will keep retrying without delay. Which kind of defeats the purpose (see description).

I am actually fine withdrawing this PR, since I don't have a good case (other than `ifs` test specifically specified alone on the command-line and one instance where such testing was needed). However, I don't see how it can hurt in the interactive mode...

AthenasJimenez:
I feel this is a big change for a particular case (IFS) and honesty, I don't think it is worthy.

busykai:
OK. Withdrawn!

PR #172, opened 2022-11-04
FW/Linux: set the malloc arena count to 2x the number of processors

thiagomacieira:
We do want to have an arena per thread that we will start, to avoid
contention in malloc() of every thread. Instead, it's much better to
simply allow each thread to have its own arena.

glibc's ptmalloc implementation creates a certain number of arenas
("M_ARENA_TEST", which defaults to 8 on 64-bit systems) before querying
the system config. Up until glibc 2.36, the query had a bug[1] that
failed to initialise some memory, so glibc concluded that the system had
lots of threads available, which albeit incorrect, was actually good for
us: OpenDCDiag sets the affinity for the worker threads to exactly 1
logical processor.

[ChangeLog][Framework] Fixed an issue encountered with glibc version
2.35.1 or later, which caused memory allocations in test threads to
contend with each other. This caused some tests to have reduced
efficiency in detecting SDEs.

[1] https://sourceware.org/bugzilla/show_bug.cgi?id=28850

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -26,7 +26,7 @@
 #include <getopt.h>
 #include <inttypes.h>
 #include <limits.h>
-#ifdef __GLIBC__
+#if __has_include(<malloc.h>)
 #  include <malloc.h>
 #endif
 #include <stdarg.h>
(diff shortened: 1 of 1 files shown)


thiagomacieira:
> @thiagomacieira, why 2 times the threads? I understand it guarantees a relief from potential contention on mem allocs, but why not 1.5 of # threads or # of threads + 10?

At least one of the internal tests does `pthread_create` in each thread. Due to timing issues, we can't guarantee that that `pthread_create` on thread 0 will happen after the start of thread N. I have a pending change to sync things up but the patch is broken.

busykai (approved):

thac0 (approved):
👍 Wow, good find.

PR #175, opened 2022-11-09
IFS: Add image version info to logging

AthenasJimenez:

### DIFF
--- a/tests/ifs/ifs.c
+++ b/tests/ifs/ifs.c
@@ -112,7 +112,9 @@ static int scan_init(struct test *test)
 
 
         if (ifs0 < 0 || faccessat(ifs0, "run_test", R_OK, 0) < 0 ||
-                faccessat(ifs0, "status", R_OK, 0) < 0 || faccessat(ifs0, "details", R_OK, 0) < 0) {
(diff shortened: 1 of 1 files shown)


thiagomacieira (dismissed):
Good. Since it's going to later conflict with #163, can you add there the printing of the version number and the chunk number to the `log_error`?

Would be nice to know if the IFS engineers prefer version numbers in hex.

AthenasJimenez:
Yes. I'll rebase this change in #163 and add the chunk number as well.

Regarding to use "log_error", I don't agree. As I see it, we should log image version and chunk number on info always. In other words, no matter if we have a pass or fail, version and chunk number are present in logs.

thiagomacieira:
> Yes. I'll rebase this change in #163 and add the chunk number as well.
> 
> Regarding to use "log_error", I don't agree. As I see it, we should log image version and chunk number on info always. In other words, no matter if we have a pass or fail, version and chunk number are present in logs.

Agreed then. But it should be on the error message too, because on lower-verbosity outputs, the `log_info` won't be present.

AthenasJimenez:
Got it. I just updated the log error to include image version.

AthenasJimenez:
This change is no longer needed. Changes were addressed by #163

PR #176, opened 2022-11-17
Adding system_idle_load implementation for Windows

pwlazlyn:
On Linux we are using 5 minute counter from /proc/loadavg to determine if the system was in idle state for the last 5 minutes. This patch implements similar counter but for Windows.

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -84,6 +84,7 @@ static constexpr struct test *ordered_test_list[] = { nullptr };
 #  include <ntstatus.h>
 #  include <shlwapi.h>
 #  include <windows.h>
+#  include <pdh.h>
 
 #  ifdef ftruncate
 // MinGW's ftruncate64 tries to check free disk space and that fails on Wine,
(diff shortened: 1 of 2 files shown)


framework/sandstone.cpp
```diff
@@ -2828,6 +2829,89 @@ static void warn_deprecated_opt(const char *opt)
...
+    //
+    // We also mutliply by number of cpus to make the metric behave more like the
+    // /proc/loadavg from Linux, so we get value from range (0.0;num_cpus()), where
+    // num_cpus() value means all cores at 100% utilization.
+    const double current_avg_cpu_usage = (vpt.doubleValue * num_cpus() / 100.0);
+    const double current_proc_queue    = vpql.doubleValue;
+    const double current_avg_load      = current_avg_cpu_usage + current_proc_queue;
+
+    double loadavg_ = loadavg.load(std::memory_order::memory_order_consume);
+    loadavg_ = (current_avg_load + ((TOTAL_5MIN_SAMPLES_COUNT - 1) * loadavg_)) / TOTAL_5MIN_SAMPLES_COUNT;
```
lukasz-tuz, line 2874: This equation won't calculate `loadavg` with the same scaling as Linux does. In Linux, when current average load is 1.0, 1 minute `loadavg` is ~0.62 after 60 seconds, and saturates at 1.0 after roughly 250 seconds.

Here, in the same conditions, after 60 seconds `loadavg_` is ~0.18, and saturates on 60 after ~1500 seconds.

```cpp
loadavg_ = loadavg_ + (SAMPLE_INTERVAL_SECONDS / TOTAL_5MIN_SAMPLES_COUNT) * (current_avg_load - loadavg_)
```
should yield results similar to `\proc\loadavg`. Otherwise we'd have to tune threshold values, and no one wants that :)

lukasz-tuz (changes requested):
Generally looks good, but some changes in how load is calculated and how errors are handled will be needed.

framework/sandstone.cpp
```diff
@@ -2847,6 +2931,8 @@ static float system_idle_load()
 
     if (ret == 1)
        return load_5;
+#elif defined(_WIN32)
```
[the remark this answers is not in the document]
pwlazlyn: Right. I think It's just a bug in previous implementation. I will create a separate PR for this. We should just return `0.f` - meaning that system is idle  if the opening of `loadavg` pseudo file fails.

framework/sandstone.cpp
```diff
@@ -2828,6 +2829,89 @@ static void warn_deprecated_opt(const char *opt)
...
+
+    loadavg.store(loadavg_, std::memory_order::memory_order_release);
+}
+
+static int setup_windows_loadavg_perf_counters()
+{
+    HQUERY load_query;
+    HANDLE load_event;
+
+    if (PdhOpenQueryA(NULL, 0, &load_query) != ERROR_SUCCESS)
```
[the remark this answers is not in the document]
pwlazlyn: We need the query to stay opened in the kernel, because we register callback with `PdhCollectQueryDataEx` that will be called when `load_event` fires associated with this query. It's closed by the kernel when the dcdiag exit.

framework/sandstone.cpp
lukasz-tuz: Equation I suggested should yield results that closely match behavior described in these articles:

* [https://www.brendangregg.com/blog/2017-08-08/linux-load-averages.html](https://www.brendangregg.com/blog/2017-08-08/linux-load-averages.html)
* [https://www.fortra.com/resources/guides/unix-load-average-part-1-how-it-works](https://www.fortra.com/resources/guides/unix-load-average-part-1-how-it-works)

1, 5, 15 min load averages for a single threaded app

But let's see what the measurements show.

pwlazlyn: opendcdiag_loadavg_plot

This is what I am getting with the code from PR. I have written two simple programs one for Linux and one for Windows.
Biggest difference can be seen when running more jobs than there are cores on the machine and Windows scheduler seem to choke right away, where Linux nicely saturates around expected value.

How to read the legend: 'L' is for Linux and 'W' is for Windows. Numbers with slash are `scheduled_jobs/available_cpus`.
For example `L32/24` means we scheduled 32 jobs on 24 cpu machine.

I was doing the experiment on 8 core Windows machine and 24 Linux machine as you can see, so the graphs will be different in terms of values, but trends are pretty similar and the speed and how they saturate. I think It's close enough and this metric can be useful on Windows for checking if cpu is idle or not, even though it behaves horribly when we have more jobs scheduled than we can run at a given time. I can experiment further with some scaling factor for PQL (\\System\\Processor Queue Length) counter so that it doesn't affect the counter that much. It shouldn't be that hard to do, It's the data collecting that takes most of the time honestly. I'll try to update this data tomorrow with the scaling factor for PQL.

lukasz-tuz: We definitely need to scale PQL a bit. Anyway, goal here is to ensure that on Windows idle state can be detected correctly, without tuning the thresholds. If current formula provides that (with some tuning to PQL), I'm ok with it.

pwlazlyn: Aight, got some more numbers.
But first I wanted to say that I misunderstood Linux implementation. For some reason I thought they are scaling nr_uninterruptible somehow, but they are not, It was just implementation of exponential moving average. Today I was analyzing it in more detail and came to this realization.


That being said, I checked how much moving average differs from exponential variant on our calculations. Below graph shows how randomly generated avgload data in two variants. We mock a sudden raise in cpu use and then a drop.
Legend: ma - moving average; ema - exponential moving average
opendcdiag_loadavg_stats2_2

Some more testing on real data on Windows with ma and ema variant. Exponential variant has "We" in marker name and "W" is with old implementation.
opendcdiag_loadavg_stats2_0

And some zoom-in on the data when we choke on Windows.
We can try bringing it closer to Linux with some scaling on PQL, but it will vary from system to system, depending on the amount of processes in the background. When we run a busy loop on all other cores, they just sit in the queue, starving. I don't think we can think about any reasonable scaling factor based on that. We would have to do some measurements on scale to know what we decided to use, works.
opendcdiag_loadavg_stats2_1

lukasz-tuz: For the purpose of detecting idle state this (EWMA) implementation will be sufficient. I was worried that regular running average would not accumulate value as fast as EWMA does, so we could - potentially - run background scan when we're not supposed to.

lukasz-tuz (approved):

pwlazlyn:
Rebased. @lukasz-tuz can I get it re-approved?

lukasz-tuz (approved):
Yessir.

PR #180, opened 2022-12-01
Fixed test runtime calculations.

jposwiata:
Test runtime (defined as current_test_startime -> current_test_endtime) is calculated as close the test as possible to avoid false positives with selftests (where it is expected that the test runs not shorter than given duration). In previous time 'endtime' included framework overhead, while 'starttime' didn't, what "randomly" resulted with CI failures

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -817,7 +817,6 @@ static void init_internal(const struct test *test)
     }
     sApp->current_slice_count = slice_count;
     sApp->current_max_threads = max_threads;
-    sApp->current_test_starttime = MonotonicTimePoint::clock::now();
 }
 
 static void initialize_smi_counts()
@@ -1174,7 +1173,7 @@ Common command-line options are:
      Display program version information.
  --1sec, --30sec, --2min, --5min
      Run for the specified amount of time in the option. In this mode, the program
-     prioritizes test execution based on prior detections. 
+     prioritizes test execution based on prior detections.
      These options are intended to drive coverage over multiple runs.
      Test priority is ignored when running in combination with the
      --test-list-file option.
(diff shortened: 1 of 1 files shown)


framework/sandstone.cpp
```diff
@@ -2120,7 +2118,12 @@ TestResult run_one_test(int *tc, const struct test *test, SandstoneApplication::
     /* First we go to do our -- possibly fractured -- normal run, we'll do retries after */
     for (sApp->current_iteration_count = 0;; ++sApp->current_iteration_count) {
         init_internal(test);
+
+        // calculate starttime->endtime, reduce the overhead to have better test runtime calculations
+        sApp->current_test_endtime = calculate_wallclock_deadline(sApp->current_test_duration - runtime, &sApp->current_test_starttime);
         state = run_one_test_once(tc, test);
+        runtime += MonotonicTimePoint::clock::now() - sApp->current_test_starttime;
```
jposwiata, line 2125: internal init/cleanup aren't included in the test runtime, but overall runtime includes all the overhead stuff.

thiagomacieira: Fair.

framework/sandstone.cpp
```diff
@@ -2170,8 +2173,8 @@ TestResult run_one_test(int *tc, const struct test *test, SandstoneApplication::
             if (sApp->total_retest_count > 0)
                 --sApp->total_retest_count;
             sApp->current_iteration_count = -iterations;
-            sApp->current_test_endtime = calculate_wallclock_deadline(sApp->current_test_duration);
             init_internal(test);
+            sApp->current_test_endtime = calculate_wallclock_deadline(sApp->current_test_duration, &sApp->current_test_starttime);
```
thiagomacieira, line 2177: Why are you passing the starttime here, when the previous code didn't?

thiagomacieira (approved):
Looks ok, so approving, but please answer the question in the comment.

framework/sandstone.cpp
jposwiata: This allows to set both from exactly same point of time (t.i. end_time := (start_time := now()) + duration). Having it split might provide some unintentional delays.

PR #182, opened 2022-12-02
argv[0] is now quoted on Windows before being passed to _spawn

pwlazlyn:
MSDN says we have to quote arguments passed in argv if htey contain space and we don't want them to get split:

	Spaces embedded in strings may cause unexpected behavior; for example,
	passing _spawn the string "hi there" will result in the new process
	getting two arguments, "hi" and "there". If the intent was to have the
	new process open a file named "hi there", the process would fail. You
	can avoid this by quoting the string: "\"hi there\"".

The lack of quoting were causing program to exit with error due to invalid argv[0] and thus program_invocation_name in child processes.

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -1831,9 +1831,16 @@ static int spawn_child(const struct test *test, intptr_t *hpid, int shmempipefd)
     snprintf(statefdstr.begin(), statefdstr.size(), "%u", unsigned(statefd));
 
     std::string random_seed = random_format_seed();
+
+#ifdef _WIN32
+    // _spawn on Windows requires argv elements to be quoted if they contain space.
+    const char * const argv0 = SANDSTONE_EXECUTABLE_NAME ".exe";
+#else
+    const char * const argv0 = SANDSTONE_EXECUTABLE_NAME;
+#endif
     const char *argv[] = {
         // argument order must match exec_mode_run()
-        program_invocation_name, "-x", test->id, random_seed.c_str(),
+        argv0, "-x", test->id, random_seed.c_str(),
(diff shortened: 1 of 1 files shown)


framework/sandstone.cpp
```diff
@@ -1831,9 +1831,21 @@ static int spawn_child(const struct test *test, intptr_t *hpid, int shmempipefd)
...
     std::string random_seed = random_format_seed();
+#ifdef _WIN32
+    // _spawn on Windows requires argv elements to be quoted if they contain space.
+    std::string argv0_str;
+    argv0_str.reserve(strlen(program_invocation_name) + 2);
+    argv0_str += '\"';
+    argv0_str += program_invocation_name;
+    argv0_str += '\"';
+
+    const char * const argv0 = argv0_str.c_str();
```
thiagomacieira, line 1842: Alternatively, since we know that the child process will make no decisions based on the path, we could pass the portion after the last slash, or even a fixed string. For example, just set
```c++
    static const char argv0[] = SANDSTONE_EXECUTABLE_NAME ".exe";
```

thiagomacieira (approved):
Approving, your choice to change again.

framework/sandstone.cpp
pwlazlyn: > Alternatively, since we know that the child process will make no decisions based on the path

I had similar thing in mind for a second, but thought brining this "dependency" to the code might not be right.
It's just hard to express in code that we are doing it and why. On the other hand I can't think of anything that will require opendcdiag to rely on the directory it was run from or argv[0] being equal to argv[0] of It's parent.
I will just leave a comment why we are doing it and stick with your suggestion.

thiagomacieira (approved):

PR #184, opened 2022-12-09
Windows: Move temp files to Temp directory.

dawid-sabat:
For Windows compilations of OpenDCDiag, temp files were created in directory where the .exe is placed. The application doesn't always have access rights to do that.

Changes here move all temp files to Windows designated area: C:\Users\%username\AppData\Local\Temp

Signed-off-by: dawid-sabat <dawid.sabat@intel.com>


framework/sysdeps/windows/tmpfile.c
```diff
@@ -43,19 +43,30 @@ static unsigned next_random()
...
+
     for (int i = 0; hFile == INVALID_HANDLE_VALUE && i < 16; ++i) {
         wcscat(_ultow(next_random(), tmpname, 36), L".tmp");            // yes, base 36
         DWORD access = GENERIC_READ | GENERIC_WRITE;
         DWORD sharemode = FILE_SHARE_DELETE | FILE_SHARE_READ | FILE_SHARE_WRITE;
         DWORD creation = CREATE_NEW;
         DWORD flags = FILE_ATTRIBUTE_TEMPORARY | FILE_FLAG_DELETE_ON_CLOSE;
-        hFile = CreateFileW(tmpname, access, sharemode, &sa, creation, flags, NULL);
+
+        if ((wcslen(tmppath) + wcslen(tmpname)) > MAX_PATH)
```
pwlazlyn, line 66: Shouldn't it be `>= MAX_PATH`?

dawid-sabat: If I count the sizes right we have a buffer (tmppath) sized MAX_PATH. If we put something with MAX_PATH length string we shouldn't have a buffer overflow.

pwlazlyn: But we need one more extra for the null terminator, don't we?

lukasz-tuz: `CreateFileW` limits the filename to `MAX_PATH`, but spec is vague on details if this includes `<NUL>` or not. `wcslen` handles non-null terminated strings, so `(wcslen(tmppath) + wcslen(tmpname)) == MAX_PATH` case should technically be fine, but I'd feel better about this if we kept the one extra slot for null termination.

lukasz-tuz (changes requested):

framework/sysdeps/windows/tmpfile.c
pwlazlyn: Also, from https://en.cppreference.com/w/cpp/string/wide/wcscat:
```
The behavior is undefined if the destination array is not large enough for the contents of
both str and dest and the terminating null wide character.
```

dawid-sabat: Done

dawid-sabat:
I noticed that this change breaks selftests in a way I don't completely understand. I suspect there might something wrong with wine configuration but I'm not sure.

```
not ok 1 Verify that the python yaml module can import regular output
# (from function `run_sandstone_yaml' in file bats/sanity-check/../testenv.bash, line 87,
#  in test file bats/sanity-check/[10](https://github.com/opendcdiag/opendcdiag/actions/runs/3657877805/jobs/6181966319#step:10:11)-yaml-validate.bats, line 9)
#   `run_sandstone_yaml --retest-on-failure=0 --selftest --timeout=60s --quick -e @positive' failed
# + wine builddir-windows-cross/opendcdiag.exe --on-crash=core --on-hang=kill -Y -o - --retest-on-failure=0 --selftest --timeout=60s --quick -e @positive
# internal error creating temporary file: File too large
```

Any ideas?

thiagomacieira (commented):
Hadn't seen this before I wrote #186

framework/sysdeps/windows/tmpfile.c
```diff
@@ -43,19 +43,30 @@ static unsigned next_random()
...
+
     for (int i = 0; hFile == INVALID_HANDLE_VALUE && i < 16; ++i) {
         wcscat(_ultow(next_random(), tmpname, 36), L".tmp");            // yes, base 36
         DWORD access = GENERIC_READ | GENERIC_WRITE;
         DWORD sharemode = FILE_SHARE_DELETE | FILE_SHARE_READ | FILE_SHARE_WRITE;
         DWORD creation = CREATE_NEW;
         DWORD flags = FILE_ATTRIBUTE_TEMPORARY | FILE_FLAG_DELETE_ON_CLOSE;
-        hFile = CreateFileW(tmpname, access, sharemode, &sa, creation, flags, NULL);
+
+        if ((wcslen(tmppath) + wcslen(tmpname)) >= MAX_PATH)
```
thiagomacieira, line 66: That doesn't happen for all file names. The shortest file name we'll generate is "0.tmp" and the longest is "1z141z4.tmp". So there's a possibility that some paths will work.

Either way, either we handle GetTempPathW returning too long a string before the loop or we simply `continue` here. Or both.

thiagomacieira (changes requested):

dawid-sabat:
@lukasz-tuz  and I made some experiments about opendcdiag selftests failing:

- tried to recreate the same conditions as CI (the compiller, libs etc) locally
- from my end I noticed that selftests and opendcdiag.exe works with wine outside the docker container.
- Lukasz has been debugging on his end and he's going to describe his effort.

thiagomacieira:
> @lukasz-tuz and I made some experiments about opendcdiag selftests failing:
> 
> * tried to recreate the same conditions as CI (the compiller, libs etc) locally
> * from my end I noticed that selftests and opendcdiag.exe works with wine outside the docker container.
> * Lukasz has been debugging on his end and he's going to describe his effort.

I recreated the conditions on Wine by simply doing a `cd /` before launching it. For the BATS testing, we need to check that the output log file (if not stdout) has a full path). We should do that for Unix too.

lukasz-tuz:
Lack of full path to output log file is only part of this (albeit needs to be fixed). However:

- default Wine prefix doesn't seem to have `%TEMP%` or `%TMP%` variables; they need to be added manually, and need to be added to the registry file in the Wine prefix being used. Passing them as shell environment variables is not sufficient (see [Wine User Guide](https://wiki.winehq.org/Wine_User%27s_Guide#Setting_Windows.2FDOS_environment_variables)).
- Need to ensure that directory pointed to by `%TEMP%` has write permissions set
- even with the steps above completed, `opendcdiag.exe` would still fail when trying to truncate the temp file with `ftruncate`. As @dawid-sabat observed, this does not happen with more up-to-date version of Wine (I used 7.22 in my tests). Older version (7.2) fails with `EFBIG` error, although length passed to `ftruncate` is just 65 kB.

Going to run a targeted test, where previously failing environment gets a Wine update (and just Wine, to rule out any other dependencies).

thiagomacieira:
I don't know where the environment came from in my ~/.wine/user.reg, but it does show:
```ini
[Environment] 1632151948
#time=1d7ae34b87983e8
"PATH"="C:\\windows\\system32;C:\\windows;C:\\windows\\system32\\wbem;C:\\windows\\system32\\WindowsPowershell\\v1.0;C:\\MinGW\\bin;C:\\Qt\\qtbase\\bin"
"QT_FORCE_STDERR_LOGGING"="1"
"TEMP"="C:\\users\\tjmaciei\\Temp"
"TMP"="C:\\users\\tjmaciei\\Temp"
```

I don't remember setting them and I probably wouldn't have set them to the path that is shown there (I'd choose "Z:\\tmp"), but the other two variables are definitely me.

Another thing is that tracing its execution shows just how Wine executes GetTempPathW:
```
1515859.354:00dc:00e0:Call KERNEL32.GetTempPathW(000000f9,0081e6a0) ret=140020632
1515859.354:00dc:00e0:trace:process:GetEnvironmentVariableW (L"TMP" 000000000081E350 260)
1515859.354:00dc:00e0:Call ntdll.RtlInitUnicodeString(0081e2d0,7b08bd14 L"TMP") ret=7b05c2d7
1515859.354:00dc:00e0:Ret  ntdll.RtlInitUnicodeString() retval=00000008 ret=7b05c2d7
1515859.354:00dc:00e0:Call ntdll.RtlQueryEnvironmentVariable_U(00000000,0081e2d0,0081e2e0) ret=7b05c305
1515859.354:00dc:00e0:trace:environ:RtlQueryEnvironmentVariable_U 0000000000000000 L"TMP" 000000000081E2E0
1515859.354:00dc:00e0:Ret  ntdll.RtlQueryEnvironmentVariable_U() retval=00000000 ret=7b05c305
1515859.354:00dc:00e0:Call ntdll.RtlGetFullPathName_U(0081e350 L"C:\\users\\tjmaciei\\Temp",00000208,0081e350,00000000) ret=7b01d1e4
1515859.354:00dc:00e0:trace:file:RtlGetFullPathName_U (L"C:\\users\\tjmaciei\\Temp" 520 000000000081E350 0000000000000000)
1515859.354:00dc:00e0:Ret  ntdll.RtlGetFullPathName_U() retval=0000002c ret=7b01d1e4
1515859.354:00dc:00e0:Call ntdll.memset(0081e6d0,00000000,000001c2) ret=7b01d390
1515859.354:00dc:00e0:Ret  ntdll.memset() retval=0081e6d0 ret=7b01d390
1515859.354:00dc:00e0:trace:file:GetTempPathW returning 23, L"C:\\users\\tjmaciei\\Temp\\"
```

That is, the darned thing does a `GetEnvironmentVariableW`! I don't know what real Windows does, but we can assume it's maybe smarter than this.

thiagomacieira:
I've just tested this patch with Wine 7.22 on Arch Linux, with and without WINE_ROOT set. The selftests launch and (mostly) pass. I get a BATS failure on `selftest_fastfail` but that seems unrelated (it's expecting code 0xC0000005 but getting 0xc0000409).

lukasz-tuz:
> the darned thing does a GetEnvironmentVariableW! I don't know what real Windows does, but we can assume it's maybe smarter than this.

Well, it probably isn't. Quoting the documentation for [GetTempPathW](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-gettemppathw):

> The GetTempPath function checks for the existence of environment variables in the following order and uses the first path found:
> 
> 1. The path specified by the TMP environment variable.
> 2. The path specified by the TEMP environment variable.
> 3. The path specified by the USERPROFILE environment variable.
> 4. The Windows directory.
> 
> Note that the function does not verify that the path exists, nor does it test to see if the current process has any kind of access rights to the path. 

The last one probably just looks for `%WINDIR%`. 

> get a BATS failure on selftest_fastfail but that seems unrelated (it's expecting code 0xC0000005 but getting 0xc0000409)

Same result here.

lukasz-tuz:
Ah, the problem was right there, staring me in the face:

```c
// MinGW's ftruncate64 tries to check free disk space and that fails on Wine,
// so use the the 32-bit offset version (which calls _chsize)
```

And the following typo:

```c
#    undef fruncate
```

After changing to 

```c
#    undef ftruncate
```

`opendcdiag` stops failing when trying to truncate the temp file with `ftruncate`.

thiagomacieira (approved):
Approving to unblock, but my comment on tmpfile.c is unaddressed.

framework/sysdeps/windows/tmpfile.c
dawid-sabat: Should be done now

busykai (approved):
LGTM.

PR #186, opened 2022-12-13
FW/Win32: find a directory for our temporary files other than "."

thiagomacieira:
When the tool is installed system-wide (e.g., C:\Program Files), it will not have write access to its current-working directory. So we try, in order:
 - GetTempPath()
 - %TEMP% environment variable
 - %TMP% environment variable
 - C:\Temp
 - the current working directory

We do a similar search in sysdeps/unix/tmpfile.c.

### DIFF
--- a/framework/sysdeps/windows/tmpfile.c
+++ b/framework/sysdeps/windows/tmpfile.c
@@ -9,6 +9,7 @@
 #include <errno.h>
 #include <fcntl.h>
 #include <inttypes.h>
+#include <io.h>
 #include <stdatomic.h>
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


dawid-sabat:
I like this implementation more than #184 .
I covered only GetTempPath() case in my PR.

framework/sysdeps/windows/tmpfile.c
```diff
@@ -40,9 +41,68 @@ static unsigned next_random()
...
+        return dirname;
+
+    dirname = _wgetenv(L"TEMP");
+    if (is_dir_writable(dirname))
+        return dirname;
+
+    dirname = _wgetenv(L"TMP");
+    if (is_dir_writable(dirname))
+        return dirname;
+
```
pwlazlyn, line 75: Shouldn't we set `dirname` back to `NULL` if `is_dir_writable(L"TMP")` check fails?

framework/sysdeps/windows/tmpfile.c
```diff
@@ -40,9 +41,68 @@ static unsigned next_random()
...
+        return false;
+
+    size_t len = wcslen(dir);
+    if (!len)
+        return false;
+    if (dir[len - 1] == L'\\')
+        return _waccess(dir, W_OK) == 0;
+
+    // append a backslash so we confirm it's a directory
+    wchar_t path[MAX_PATH];
```
pwlazlyn, line 56: `MAX_PATH` might not be enough to hold the contents of result of `_wgetenv()`

pwlazlyn (commented):
There are couple of places where we append to the buffer of `MAX_PATH` length and don't check if we have enough space to do that.

framework/sysdeps/windows/tmpfile.c
```diff
@@ -40,9 +41,68 @@ static unsigned next_random()
...
+    return NULL;
+}
+
+// fills in the temporary directory in tmpname and returns the length of the path
+// (including the terminating backslash)
+static size_t fill_tmpdir(wchar_t tmpname[MAX_PATH])
+{
+    size_t len = GetTempPathW(MAX_PATH, tmpname);
+    if (len) {
+        wcscat(tmpname, L"\\");
```
lukasz-tuz, line 85: `GetTempPathW` always returns a string that ends with a backslash; this step is not really needed. Also, if string with length equal to `MAX_PATH` was returned, this step would overflow `tmpname`.

framework/sysdeps/windows/tmpfile.c
```diff
@@ -40,9 +41,68 @@ static unsigned next_random()
...
+    wcscpy(tmpname, othertmp);
+    wcscat(tmpname, L"\\");
+    return wcslen(tmpname);
+}
+
 int open_memfd(enum MemfdCloexecFlag flag)
 {
-    wchar_t tmpname[sizeof SANDSTONE_STRINGIFY(UINT_MAX) ".tmp"];
+    wchar_t tmppath[MAX_PATH];
+    wchar_t *tmpname = tmppath + fill_tmpdir(tmppath);
```
lukasz-tuz, line 105: If `fill_tmpdir`, and its underlying API calls, return string with length longer than `MAX_PATH`, then `tmpname` pointer would exceed boundaries of `tmppath` array. 
We're trying to append name of a temp file few lines below without checking if the name will fit into the buffer.
Would be best to reserve at least 12 characters in the `tmppath` and - if total length of temp directory path + temp file name - exceeded `MAX_PATH - 12` - convert name of the file to 8.3 format.

lukasz-tuz (changes requested):

framework/sysdeps/windows/tmpfile.c
thiagomacieira: Didn't notice that it returned a string ending in backslash in the documentation, but it's right there. Thanks, adapting.

framework/sysdeps/windows/tmpfile.c
thiagomacieira: Fixing.

thiagomacieira (commented):
I don't think any of this is necessary or will work. From [_waccess](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/access-waccess?view=msvc-170):

> This function only checks whether the file and directory are read-only or not, it doesn't check the filesystem security settings.

So let's go with #184.

thiagomacieira:
Abandoning. The only way to know if we can create files in a directory is to attempt it. So let's go with the simpler #184

PR #187, opened 2023-01-19
cpu_features.h::x86_architectures sorted properly

jposwiata:
Natural sorting of 'x86_architectures' array to keep binary images identical from compilation to compilation. CPU generations are alphanumerically sorted by features (t.i. list containing number of enabled features as '%02d' and names of these features).

### DIFF
--- a/framework/scripts/x86simd_generate.pl
+++ b/framework/scripts/x86simd_generate.pl
@@ -283,7 +283,7 @@ struct X86Architecture
 };
 
 static const struct X86Architecture x86_architectures[] = {|;
-for (sort { $b <=> $a } keys %sorted_archs) {
+for (sort keys %sorted_archs) {
     my $arch = $sorted_archs{$_};
     next if $arch->{base} eq "<>";
     printf "    { cpu_%s, \"%s\" },\n", $arch->{id}, $arch->{prettyname};


pwlazlyn:
I'd like to see this change go in, but I don't know perl enough to review the code change, so I am not approving.

jposwiata:
Keep it not merged until v130 release.

framework/scripts/x86simd_generate.pl
```diff
@@ -283,7 +283,7 @@
 };
 
 static const struct X86Architecture x86_architectures[] = {|;
-for (sort { $b <=> $a } keys %sorted_archs) {
+for (sort keys %sorted_archs) {
```
lukasz-tuz, line 286: I think that sorting these keys lexically rather than numerically is better, so this change is OK. But there's an additional problem, I think: in line 275 there's a potential key collision:

```perl
my $key = sprintf "%02d_%s", scalar(@{$arch->{allfeatures}}), join(',', @{$arch->{allfeatures}});
```

If we ever get inputs such as that `allfeatures` are equal in two separate architectures, the former would be overwritten by the latter.

Now, if we generated keys using simple incrementing integer rather than getting array length through
```perl
scalar(@{$arch->{allfeatures}})
```
we'd get key values that do not collide and can be sorted numerically.

jposwiata: I think, the intention was to sort the list by the lengths (to have the one with longest list of features at the end). So lists longer than 100 features will not work for us ('%02d' format specifier is used to add count before features).
Just incrementing will be disaster as hash map has no specified order at all (and all post 5.18 will give "random" order, different from run to run).
Should we use '%03d' (as `cpuid` has multiple registers/possible combinations)

lukasz-tuz: Gotcha. Let's review the implicit inheritances - to ensure we don't accidentally miss a generation - in a separate PR. This PR looks good and seems safe with the input data we have.

lukasz-tuz (approved):

PR #190, opened 2023-01-25
Skip feature: add "Skip Reason" framework changes

dshivaraintel:
These changes will enable tests to add reason that whey skip a test. Skips might be due to cpu that doesn't support the test, kernel not supported, device not found etc.

Truncated output from one of the tests that will skip:

skip in main:
```
- test: lock_in_mispredicted_branches
  result: skip
  skip-category: RESOURCE_UNAVAILABLE
  skip-reason: This test requires atl east 2 cpus to run.
```
skip in run:

```
- test: lock_in_mispredicted_branches
  result: skip
  skip-category: RUNTIME_SKIP
  skip-reason: Skipped in Run. Check individual thread messages
```

The yaml format will escape special characters in the skip message but not new line since we are printing the skip message in a single line.

### DIFF
--- a/framework/logging.cpp
+++ b/framework/logging.cpp
@@ -110,6 +110,7 @@ enum LogTypes {
     UserMessages = 0,
     Preformatted = 1,
     UsedKnobValue = 2,
+    SkipMessages = 3,
 };
 
 struct ThreadLog
@@ -149,7 +150,7 @@ private:
     const char *stdout_terminator = nullptr;
 
     void maybe_print_yaml_marker(int fd);
-    void print_thread_messages();
+    void print_thread_messages(ChildExitStatus status);
     void print_thread_header(int fd, int cpu, int verbosity);
     void print_child_stderr();
     static std::string format_status_code(ChildExitStatus status);
@@ -176,7 +177,7 @@ private:
     void maybe_print_messages_header(int fd);
     void print_thread_header(int fd, int cpu, int verbosity);
     static int print_test_knobs(int fd, mmap_region r);
-    static int print_one_thread_messages(int fd, mmap_region r, int level);
+    static int print_one_thread_messages(int fd, mmap_region r, int level, ChildExitStatus status);
     void print_result_line(ChildExitStatus status);
 
     enum TestHeaderTime { AtStart, OnFirstFail };
@@ -192,14 +193,14 @@ public:
         prepare_line_prefix();
     }
 
-    void print(int tc);
+    void print(int tc, ChildExitStatus status);
 
 private:
     std::string timestamp_prefix;
 
     void prepare_line_prefix();
     void print_thread_header(int fd, int cpu, const char *prefix);
-    void print_thread_messages();
+    void print_thread_messages(ChildExitStatus status);
     void print_child_stderr();
 };
 
@@ -230,7 +231,7 @@ static const char *strnchr(const char *buffer, char c, size_t len)
 
 static uint8_t message_code(enum LogTypes logType, int level)
 {
-    assert((int)logType < 3);
+    assert((int)logType < 4);
     unsigned code = ((unsigned)logType + 1) << 4;
     code |= (level & 0xf);
     return (uint8_t)code;
@@ -438,6 +439,32 @@ static uint8_t status_level(char letter)
 
     log_warning("got improper status log message '%c'", letter);
     return 2;
+}
+
+static const char *char_to_skip_category(int val)
+{
+    switch (val) {
+    case SkipCategory(1):
+        return "ResourceIssueSkipCategory";
+    case SkipCategory(2):
+        return "CpuNotSupportedSkipCategory";
+    case SkipCategory(3):
+        return "DeviceNotFoundSkipCategory";
+    case SkipCategory(4):
+        return "DeviceNotConfiguredSkipCategory";
+    case SkipCategory(5):
+        return "UnknownSkipCategory";
+    case SkipCategory(6):
+        return "RuntimeSkipCategory";
+    case SkipCategory(7):
+        return "SelftestSkipCategory";
+    case SkipCategory(8):
+        return "OsNotSupportedSkipCategory";
+    case SkipCategory(9):
+        return "ThreadIssueSkipCategory";
+    }
+
+    return "NO CATEGORY PRESENT";
 }
 
 static std::vector<ThreadLog> &all_thread_logs() noexcept
@@ -1221,6 +1248,30 @@ void log_platform_message(const char *fmt, ...)
     log_message_preformatted(-1, msg);
 }
 
+#undef log_message_skip
+void log_message_skip(int thread_num, SkipCategory category, const char *fmt, ...)
+{
+    if (current_output_format() == SandstoneApplication::OutputFormat::no_output)
+        return;
+
+    va_list va;
+    va_start(va, fmt);
+    std::string msg = create_filtered_message_string(fmt, va);
+    va_end(va);
+    
+    msg.insert(msg.begin(), category);
+
+    if (msg[msg.size() - 1] == '\n')
+        msg.pop_back(); // remove trailing newline
+             
+    int level = LOG_LEVEL_QUIET;
+    FILE *log = log_for_thread(thread_num).log;
+    fflush(log);
+    fputc(message_code(SkipMessages, level), log);
+    fwrite(msg.c_str(), 1, msg.size(), log);
+    logging_stream_close(log);
+}
+
 #undef log_message
 void log_message(int thread_num, const char *fmt, ...)
 {
@@ -1545,50 +1596,93 @@ static void print_content_single_line(int fd, std::string_view before,
     writeln(fd, before, escape_for_single_line(message, escaped), after);
 }
 
-static void format_and_print_message(int fd, int message_level, std::string_view message)
+static void format_and_print_message(int fd, int message_level, std::string_view message, bool from_thread_message)
 {
-    const char *levels[] = { "error", "warning", "info", "debug" }; //levels for yaml
+    const char *levels[] = { "error", "warning", "info", "debug", "skip" }; //levels for yaml
 
     if (message.find('\n') != std::string_view::npos) {
         /* multi line */
         if (current_output_format() == SandstoneApplication::OutputFormat::yaml) {
-            if (message_level > int(std::size(levels)))
-                message_level = std::size(levels) - 1;
+            if (from_thread_message) {
+                 if (message_level > int(std::size(levels)))
+                    message_level = std::size(levels) - 1;
 
-            // trim a trailing newline, if any (just one)
-            if (message[message.size() - 1] == '\n')
-                message.remove_suffix(1);
-            writeln(fd, indent_spaces(), "    - level: ", levels[message_level]);
-            writeln(fd, indent_spaces(), "      text: |1");
-            print_content_indented(fd, "       ", message);
+                // trim a trailing newline, if any (just one)
+                if (message[message.size() - 1] == '\n')
+                    message.remove_suffix(1);
+
+                writeln(fd, indent_spaces(), "    - level: ", levels[message_level]);
+                writeln(fd, indent_spaces(), "      text: |1");
+                print_content_indented(fd, "       ", message);
+            } else {
+                writeln(fd, indent_spaces(), "  skip-reason: |1");
+                print_content_indented(fd, "   ", message);
+            }   
         } else {
-            writeln(fd, "  - |");
-            print_content_indented(fd, "    ", message);
+            if (from_thread_message) // are you writing individual thread's message?
+                writeln(fd, "  - |");
+            else                     // are you writing init thread's message?
+                writeln(fd, "\n", indent_spaces(), " - |");
+            print_content_indented(fd, "    ", message); 
         }
     } else {
         /* single line */
         if (current_output_format() == SandstoneApplication::OutputFormat::yaml) {
-            iovec vec[] = { IoVec(indent_spaces()), IoVec("    - { level: "), IoVec(levels[message_level]) };
-            IGNORE_RETVAL(writev(fd, vec, std::size(vec)));
-            print_content_single_line(fd, ", text: '", message, "' }");
+            if (from_thread_message) {
+                iovec vec[] = { IoVec(indent_spaces()), IoVec("    - { level: "), IoVec(levels[message_level]) };
+                IGNORE_RETVAL(writev(fd, vec, std::size(vec)));
+                print_content_single_line(fd, ", text: '", message, "' }");
+            } else {
+                iovec vec[] = { IoVec(indent_spaces()) };
+                IGNORE_RETVAL(writev(fd, vec, std::size(vec)));
+                print_content_single_line(fd, "  skip-reason: '", message, "'");
+            }
         } else {
-            char c = '\'';
-            print_content_single_line(fd, "   - '", message, std::string_view(&c, 1));
+            if (from_thread_message) { // are you writing individual thread's message?
+                char c = '\'';
+                print_content_single_line(fd, "   - '", message, std::string_view(&c, 1));
+            } else { // are you writing init thread's message?
+                print_content_single_line(fd, "'", message, "'"); 
+            }
         }
     }
+}
 
+static std::string get_skip_message(int thread_num)
+{
+    std::string skip_message;
+    auto log = log_for_thread(thread_num);
+    struct mmap_region r = mmap_file(log.log_fd);
+    auto ptr = static_cast<const char *>(r.base);
+    const char *end = ptr + r.size;
+    const char *delim;
+
+    for ( ; ptr < end && (delim = strnchr(ptr, '\0', end - ptr)) != nullptr; ptr = delim + 1) {
+        uint8_t code = (uint8_t)*ptr++;
+        if (log_type_from_code(code) == LogTypes::SkipMessages) {
+            skip_message.assign(ptr, delim);
+            break;
+        }
+    }
+    munmap_file(r);
+    return skip_message;
+}
+
+static inline void format_skip_message(std::string &skip_message, std::string_view message)
+{
+    skip_message += message.substr(1, message.size());
 }
 
 /// Returns the lowest priority found
 /// (this function is shared between the TAP and key-value pair loggers)
-static int print_one_thread_messages(int fd, struct per_thread_data *data, struct mmap_region r, int level)
+static int print_one_thread_messages(int fd, struct per_thread_data *data, struct mmap_region r, int level, ChildExitStatus status)
 {
     int lowest_level = INT_MAX;
     auto ptr = static_cast<const char *>(r.base);
     const char *end = ptr + r.size;
     const char *delim;
 
-    for ( ; ptr < end && (delim = strnchr(ptr, '\0', end - ptr)) != NULL; ptr = delim + 1) {
+    for ( ; ptr < end && (delim = strnchr(ptr, '\0', end - ptr)) != nullptr; ptr = delim + 1) {
         uint8_t code = (uint8_t)*ptr++;
         int message_level = level_from_code(code);
 
@@ -1598,11 +1692,19 @@ static int print_one_thread_messages(int fd, struct per_thread_data *data, struc
         std::string_view message(ptr, delim - ptr);
         switch (log_type_from_code(code)) {
         case UserMessages:
-            format_and_print_message(fd, -1, message);
+            format_and_print_message(fd, -1, message, true);
             break;
 
         case Preformatted:
             IGNORE_RETVAL(write(fd, ptr, delim - ptr));
+            break;
+        
+        case SkipMessages: 
+            if (status.result != TestSkipped) { // If test skipped in init, no need to display as it's already displayed in print_result_line
+                std::string skip_message;
+                format_skip_message(skip_message, message);
+                format_and_print_message(fd, -1, std::string_view{&skip_message[0], skip_message.size()}, true);
+            }
             break;
 
         case UsedKnobValue: {
@@ -1667,8 +1769,7 @@ inline AbstractLogger::AbstractLogger(const struct test *test, TestResult state_
         all_logs[0] = reopen_log(-1);   // main thread
 
     if (state == TestSkipped)
-        return;         // no threads were started
-
+        return;         // no threads were started    
     for (int i = 0; i < num_cpus(); ++i) {
         struct per_thread_data *data = cpu_data_for_thread(i);
         ThreadState thr_state = data->thread_state.load(std::memory_order_relaxed);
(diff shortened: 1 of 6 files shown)


framework/sandstone.h
```diff
@@ -76,6 +77,19 @@ extern "C" {
...
+typedef enum SkipCategory{
+    RESOURCE_UNAVAILABLE=65,
+    CPU_NOT_SUPPORTED=66,
+    DEVICE_NOT_FOUND=67,
+    DEVICE_NOT_CONFIGURED=68,
+    OTHERS=69,
+    RUNTIME_SKIP=70,
+}SkipCategory;
+
+#define log_skip(skip_category,...)           log_message_skip(thread_num, skip_category,SANDSTONE_LOG_SKIP __VA_ARGS__)
```
lukasz-tuz, line 91: Nitpicking: there should be a space before `SANDSTONE_LOG_SKIP`. Also, would be cool to have some comments here, similar to other log macros. Done nitpicking :)

framework/logging.cpp
```diff
@@ -1223,6 +1266,27 @@ void log_platform_message(const char *fmt, ...)
...
+        return;         /* no Debug in non-debug build */
+    va_list va;
+    va_start(va, fmt);
+    std::string msg = create_filtered_message_string(fmt, va);
+    va_end(va);
+    std::string tmp_s;
+    char ch=skip_category_to_char(category);
+    tmp_s+=SANDSTONE_LOG_SKIP;
+    tmp_s.push_back(ch);
+    tmp_s+=msg.substr(3,msg.size());
```
lukasz-tuz, line 1286: Please add spaces around operators to maintain coding style consistency.

framework/logging.cpp
```diff
@@ -1571,7 +1653,13 @@ static int print_one_thread_messages(int fd, struct per_thread_data *data, struc
...
                 /* single line */
                 char c = '\'';  // I'm lazy
-                print_content_single_line(fd, "   - '", message, std::string_view(&c, 1));
+                if(message[0]=='S'){
+                    std::string message_string=std::string(message);
+                    std::string skip_message=message_string.substr(0,3)+message_string.substr(4,message_string.size());
+                    std::string_view skip_message_view(&skip_message[0],skip_message.size());
+                    print_content_single_line(fd, "   - '", skip_message_view, std::string_view(&c, 1));
+                }else
+                    print_content_single_line(fd, "   - '", message, std::string_view(&c, 1));
```
lukasz-tuz, line 1662: This part calls for some comments. Why 'S'? I'm assuming that's how skipped tests' log message begins (with `S>`). Which characters are the `substr` meant to remove?

lukasz-tuz (changes requested):
Please adjust formatting (spaces after commas and around operators, etc.) so that it's consistent with the rest of the code. Otherwise looks good, although if there's a way to reduce code repeatability (the same, long sequences involving `substr`, etc.), it'd be great to take it. Maybe macros or inline functions?

framework/logging.cpp
dshivaraintel: the skip messages always start with `S> ` but there is a skip category that the framework needs to encode along with the skip message. For e.g., if the skip category is `RESOURCE_UNAVAILABLE` which corresponds to letter `A`, this needs to be part of the skip message shared to the main sandstone process. When decoding on the main process side, we need to exclude it when logging. 

For e.g.,

Encoded message: `S> ATest skipped due to resource unavailability`
Decode message:   `S> Test skipped due to resource unavailability` No `A` here.

lukasz-tuz (approved):
Looks good!

dawid-sabat:
I'd suggest squashing the two commits in this PR and reapproving it. The second commit is about review fixes and I think we can join both of them.

dshivaraintel:
> I'd suggest squashing the two commits in this PR and reapproving it. The second commit is about review fixes and I think we can join both of them.

@dawid-sabat  I just rebased it. I am guessing this should be fine now?

dawid-sabat (approved):
Good for me

dshivaraintel:
Does this require another approval from @lukasz-tuz ? Merging is still blocked, not sure what the next steps are.

lukasz-tuz (dismissed):

framework/logging.cpp
```diff
@@ -1571,7 +1658,13 @@ static int print_one_thread_messages(int fd, struct per_thread_data *data, struc
...
+                if(message[0] == 'S'){
+                    std::string skip_message;
+                    FORMAT_SKIP_MESSAGE(skip_message, message);
+                    std::string_view skip_message_view(&skip_message[0], skip_message.size());
+                    print_content_single_line(fd, "   - '", skip_message_view, std::string_view(&c, 1));
+                }else
+                    print_content_single_line(fd, "   - '", message, std::string_view(&c, 1));
             } else {
                 /* multi line */
                 writeln(fd, "  - |");
```
thiagomacieira, line 1670: Needs to be handled in the newline case too.

framework/logging.cpp
```diff
@@ -1684,11 +1777,29 @@ void KeyValuePairLogger::prepare_line_prefix()
...
     logging_printf(LOG_LEVEL_QUIET, "%s_result = %s\n", test->id,
-                   state == TestSkipped ? "skip" :
+                   state == TestSkipped ? skip_message.c_str() :
                    state == TestFailed ? "fail" : "pass");
+    
+    if(status.result == TestPassed && state == TestSkipped){ // if test passed in init and skipped on all threads in run
+        logging_printf(LOG_LEVEL_QUIET, "%s_skip_category = %s\n", test->id, "RUNTIME_SKIP");
+        logging_printf(LOG_LEVEL_QUIET, "%s_skip_reason = %s\n", test->id, "Skipped in Run. Check individual thread messages");
+    }
+    else{
```
thiagomacieira, line 1794: Coding style: same line, spaces around braces.

framework/logging.cpp
```diff
@@ -2212,8 +2339,24 @@ void YamlLogger::print_result_line(ChildExitStatus status)
...
+        case TestSkipped:{
+            logging_printf(loglevel, "  result: skip\n");
+            if(status.result == TestPassed && state == TestSkipped){ // if test passed in init and skipped on all threads in run
+                logging_printf(loglevel, "  skip-category: %s\n", "RUNTIME_SKIP");
+                return logging_printf(loglevel, "  skip-reason: %s\n", "Skipped in Run. Check individual thread messages");
+            }else{
+                std::string skip_message_from_main_thread;
+                get_skip_message(-1, skip_message_from_main_thread);
+                std::string storage;
+                escape_for_single_line(skip_message_from_main_thread, storage);
```
thiagomacieira, line 2351: Doesn't work if the message contains a newline.

framework/sandstone.h
```diff
@@ -82,6 +83,21 @@ extern "C" {
 #  define log_debug( ...)       (void)0
 #endif
 
+// skip categories
+typedef enum SkipCategory{
+    RESOURCE_UNAVAILABLE=65,
```
thiagomacieira, line 88: Why start on 65? If it's important, add a comment here.

thiagomacieira (changes requested):
Also needs coding style fixes. Please make sure there are spaces before opening parentheses and braces, after closing parentheses and braces, that `else` are on the same line as the closing brace.

Opening braces are on the same line, except for top-level functions.

thiagomacieira (commented):
Also needs selftests. Please add tests that:
* skip in the main thread
* skip in all run threads
* skip in some run threads but pass in others
* skip with a newline in the message anywhere but the last character

I'd like BATS tests for this. I can help you write them, but they'll need those selftests above.

framework/logging.cpp
thiagomacieira: The skip message could be another entry in the log type enum:
https://github.com/opendcdiag/opendcdiag/blob/2ac0d930ae7f3472794d4e05fbfdfd014c928e44/framework/logging.cpp#L109-L113

framework/logging.cpp
```diff
@@ -1198,7 +1245,7 @@ FILE *logging_stream_open(int thread_num, int level)
 static inline void assert_log_message(const char *fmt)
 {
     assert(fmt);
-    assert(fmt[0] == 'd' || fmt[0] == 'I' || fmt[0] == 'W' || fmt[0] == 'E');
+    assert(fmt[0] == 'd' || fmt[0] == 'I' || fmt[0] == 'W' || fmt[0] == 'E' || fmt[0] == 'S');
```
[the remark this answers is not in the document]
dshivaraintel: They are not. I just wanted to keep the design as is in doing a sanity check that the skip message starts with `S> ` consistent with other messages(d, I, W, E).

framework/logging.cpp
```diff
@@ -1223,6 +1270,28 @@ void log_platform_message(const char *fmt, ...)
...
 
+#undef log_message_skip
+void log_message_skip(int thread_num, SkipCategory category, const char *fmt, ...){
+
+    if (current_output_format() == SandstoneApplication::OutputFormat::no_output)
+        return;
+
+    assert_log_message(fmt);
+    if (!SandstoneConfig::Debug && *fmt == 'd')
+        return;         /* no Debug in non-debug build */
```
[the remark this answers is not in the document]
dshivaraintel: I can remove the debug check, this is not necessary. Regarding the prefixes, i just wanted to be consistent with the other log messages which start with `D|E|W|I|d> `

framework/logging.cpp
```diff
@@ -435,12 +439,55 @@ static uint8_t status_level(char letter)
...
+        return 67;
+    case SkipCategory::DEVICE_NOT_CONFIGURED:
+        return 68;
+    case SkipCategory::OTHERS:
+        return 69;
+    case SkipCategory::RUNTIME_SKIP:
+        return 70;
+    }
+
+    log_warning("No category present, please select one from the predefined ones");
```
busykai, line 467: "Please select" does not apply to this situation, it's not interactive.

framework/sandstone.h
```diff
@@ -82,6 +83,21 @@ extern "C" {
...
+    CPU_NOT_SUPPORTED=66,
+    DEVICE_NOT_FOUND=67,
+    DEVICE_NOT_CONFIGURED=68,
+    OTHERS=69,
```
busykai, line 92: What `OTHERS` is supposed to mean? First of all, it should be singular. Secondly, is `UNKNOWN` is a better name for this?

framework/sandstone.h
```diff
@@ -82,6 +83,21 @@ extern "C" {
 #  define log_debug( ...)       (void)0
 #endif
 
+// skip categories
+typedef enum SkipCategory{
```
busykai, line 87: Nitpick: add space before `{`.

framework/sandstone.h
```diff
@@ -82,6 +83,21 @@ extern "C" {
...
+    RESOURCE_UNAVAILABLE=65,
+    CPU_NOT_SUPPORTED=66,
+    DEVICE_NOT_FOUND=67,
+    DEVICE_NOT_CONFIGURED=68,
+    OTHERS=69,
+    RUNTIME_SKIP=70,
+}SkipCategory;
```
busykai, line 94: Nitpick: space.

framework/logging.cpp
```diff
@@ -1547,6 +1616,24 @@ static void print_content_single_line(int fd, std::string_view before,
...
+    auto log = log_for_thread(thread_num);
+    struct mmap_region r = mmap_file(log.log_fd);
+    auto ptr = static_cast<const char *>(r.base);
+    const char *end = ptr + r.size;
+    const char *delim;
+
+    for ( ; ptr < end && (delim = strnchr(ptr, '\0', end - ptr)) != NULL; ptr = delim + 1) {
+        ptr++;
+        if(*ptr == 'S'){
+            skip_message.assign(ptr+3, delim-(ptr+3));
```
busykai, line 1629: This may go beyond `end`.

framework/logging.cpp
```diff
@@ -1684,11 +1777,29 @@ void KeyValuePairLogger::prepare_line_prefix()
...
+    get_skip_message(-1, skip_message_from_main_thread);
+
     logging_printf(LOG_LEVEL_QUIET, "%s_result = %s\n", test->id,
-                   state == TestSkipped ? "skip" :
+                   state == TestSkipped ? skip_message.c_str() :
                    state == TestFailed ? "fail" : "pass");
+    
+    if(status.result == TestPassed && state == TestSkipped){ // if test passed in init and skipped on all threads in run
+        logging_printf(LOG_LEVEL_QUIET, "%s_skip_category = %s\n", test->id, "RUNTIME_SKIP");
+        logging_printf(LOG_LEVEL_QUIET, "%s_skip_reason = %s\n", test->id, "Skipped in Run. Check individual thread messages");
```
busykai, line 1792: I really dislike "Skipped in Run" message. What about "All CPUs skipped during execution, check log for details."

framework/logging.cpp
```diff
@@ -1684,11 +1777,29 @@ void KeyValuePairLogger::prepare_line_prefix()
...
+    }
+    else{
+        if(skip_message_from_main_thread.size() > 0){
```
busykai, line 1795: This is misleading name and concept. A test skips on "main thread" when it skips in the `init()` function which is executed single-threaded before threads are launched and execute `run()`. So this should be called `init_skip_message`. The entire concept of "main thread" should be removed from the code.

busykai (changes requested):
In addition to the comments I left, please reformat the code to:
1. Have spaces before the conditional statement and the parenthesis (e.g. `if (`, not `if(`)
2. Space before the braces (e.g. `if () {`, not `if(){`)
3. Functions should have opening brace on the new line.

framework/logging.cpp
dshivaraintel: Working on this. The new version doesn't have this function at all.

framework/sandstone.h
dshivaraintel: Will change accordingly. Changing this too which will have singular values.

framework/logging.cpp
dshivaraintel: makes sense. I also want to replace the word "execution" with "run" or "run phase" because execution can also mean the init or run or cleanup functions.

framework/logging.cpp
dshivaraintel: This part collects message from thread0. Thread0 can either be the "main" thread in `init` function or `cpu=0` in `run` function. Skip can happen in `init` or all cpus in `run` function can skip. In both these cases, we are getting the skip message from thread 0. Thread 0 is called main thread in the `init` function where as just `Thread: 0` in the `run`. perhaps i should change to `skip_message_from_thread0`?

I feel the `init_skip_message` doesn't capture the essence of all thread skips in the `run` function.

busykai: this actually code treats only init skips. but alright, you're getting it using `get_skip_message(-1, ...)` somewhere else. so it just should be called `skip_message` -- it's the entire tests "skip message".

dshivaraintel: Sorry, i was wrong. Just jumped to the conclusion too early without much understanding. You are right it's only skip message from `init` function. Let me change it `init_skip_message` itself.

framework/logging.cpp
busykai: how about "All CPUs skipped while executing 'test_run()' function, check log for details"?

We refer to `test_run()` function in the test writing guide.

dshivaraintel: Done 

```
result: skip
skip-category: RUNTIME_SKIP
skip-reason: All CPUs skipped while executing 'test_run()' function, check log for details
time-at-end:   { elapsed:     10.000, now: !!timestamp '2023-02-16T22:55:42Z' }
```

framework/logging.cpp
dshivaraintel: Can you explain more on this? Cannot think of a scenario.

I thought the `delim` will ensure it doesn't go over the end. For e.g. message can be only of this format

`I> This is info\0W> This is warning\0E> This is error\0D> This is debug\0S> This is skip\0`

busykai: In theory it could. If the message is less than 3 chars (e.g. writing ended abruptly due to a crash), it only has `ptr < end` as the loop condition. Than the expression `ptr + 3` will go beyond.

dshivaraintel: This has been handled in the second commit which i have pushed.

framework/logging.cpp
```diff
@@ -1221,6 +1246,33 @@ void log_platform_message(const char *fmt, ...)
...
+        return;
+
+    assert_log_message(fmt);
+    va_list va;
+    va_start(va, fmt);
+    std::string msg = create_filtered_message_string(fmt, va);
+    va_end(va);
+
+    std::string tmp_s;
+    tmp_s += SANDSTONE_LOG_SKIP;
```
thiagomacieira, line 1262: Please remove the call to `assert_log_message` and starting with `SANDSTONE_LOG_SKIP`.

framework/logging.cpp
```diff
@@ -1221,6 +1246,33 @@ void log_platform_message(const char *fmt, ...)
...
+    tmp_s.push_back(category);
+    tmp_s += msg.substr(3, msg.size());
+    if (tmp_s[tmp_s.size() - 1] == '\n')
+        tmp_s.pop_back();           // remove trailing newline
+
+    int level = status_level(tmp_s[0]);
```
thiagomacieira, line 1268: Note that `tmp_s[0]` was 'S', so this was wrong.

framework/logging.cpp
```diff
@@ -1545,43 +1597,86 @@ static void print_content_single_line(int fd, std::string_view before,
...
+            } else {
+                writeln(fd, indent_spaces(), "  skip-reason: |1");
+                print_content_indented(fd, "   ", message);
+            }   
         } else {
-            writeln(fd, "  - |");
+            if (from_thread_message)
+                writeln(fd, "  - |");
+            else
+                writeln(fd, indent_spaces(), "\n  - |");
```
thiagomacieira, line 1626: `indent_spaces()` followed by a newline sounds wrong. Did you mean `"\n", indent_spaces(), "  - |"`?

framework/logging.cpp
```diff
@@ -1545,43 +1597,86 @@ static void print_content_single_line(int fd, std::string_view before,
...
 
+static std::string_view get_skip_message(int thread_num, std::string &skip_message)
+{
+    auto log = log_for_thread(thread_num);
+    struct mmap_region r = mmap_file(log.log_fd);
+    auto ptr = static_cast<const char *>(r.base);
+    const char *end = ptr + r.size;
+    const char *delim;
+
+    for ( ; ptr < end && (delim = strnchr(ptr, '\0', end - ptr)) != NULL; ptr = delim + 1) {
```
thiagomacieira, line 1660: This is C++, please use `nullptr`, not `NULL`.

framework/logging.cpp
```diff
@@ -1545,43 +1597,86 @@ static void print_content_single_line(int fd, std::string_view before,
...
+        ptr++;
+        if (*ptr == 'S') {
```
thiagomacieira, line 1662: Instead of checking for the letter 'S' in the message prefix, please use the message code that is in the byte you skipped.

framework/logging.cpp
```diff
@@ -1545,43 +1597,86 @@ static void print_content_single_line(int fd, std::string_view before,
...
+            break;
+        }
+    }
+    munmap_file(r);
+    return skip_message;
+}
+
+static inline void format_skip_message(std::string &skip_message, std::string_view message)
+{
+    skip_message += message.substr(0, 3);
```
thiagomacieira, line 1673: Remove the 'S>' prefix.

framework/logging.cpp
```diff
@@ -1598,12 +1693,23 @@ static int print_one_thread_messages(int fd, struct per_thread_data *data, struc
         std::string_view message(ptr, delim - ptr);
         switch (log_type_from_code(code)) {
         case UserMessages:
-            format_and_print_message(fd, -1, message);
+            format_and_print_message(fd, -1, message, true);
```
thiagomacieira, line 1696: Boolean trap... sorry, this was my idea to use the same function, but it didn't end up looking good. I'll figure out a way to fix this.

framework/logging.cpp
```diff
@@ -1598,12 +1693,23 @@ static int print_one_thread_messages(int fd, struct per_thread_data *data, struc
...
             break;
 
         case Preformatted:
             IGNORE_RETVAL(write(fd, ptr, delim - ptr));
             break;
+        
+        case SkipMessages: {
```
thiagomacieira, line 1703: Nitpick: no need for braces.

framework/logging.cpp
```diff
@@ -1668,7 +1774,7 @@ inline AbstractLogger::AbstractLogger(const struct test *test, TestResult state_
 
     if (state == TestSkipped)
         return;         // no threads were started
-
+
```
thiagomacieira, line 1777: Nitpick: blank but not empty line.

framework/logging.cpp
```diff
@@ -2169,13 +2305,24 @@ inline int YamlLogger::print_one_thread_messages(int fd, mmap_region r, int leve
...
-            format_and_print_message(fd, message_level, message);
+            format_and_print_message(fd, message_level, message, true);
             break;
 
         case Preformatted:
             IGNORE_RETVAL(write(fd, message.data(), message.size()));
             break;
-
+        
+        case SkipMessages: {
```
thiagomacieira, line 2315: Nitpick: braces not required.

framework/logging.cpp
```diff
@@ -2205,8 +2352,27 @@ void YamlLogger::print_result_line(ChildExitStatus status)
     case TestPassed:
         // recheck, as status.result does not take failing threads into account
         switch (state) {
-        case TestSkipped:
-            return logging_printf(loglevel, "  result: skip\n");
+        case TestSkipped: {
```
thiagomacieira, line 2355: Ditto.

framework/sandstone.h
```diff
@@ -84,6 +85,22 @@ extern "C" {
...
+    DEVICE_NOT_CONFIGURED,
+    UNKNOWN,
+    RUNTIME_SKIP,
+    SELFTEST_SKIP,
+} SkipCategory;
+
+/// logs a skip message to the logfile. log_skip accepts the category to which the skip belongs to
+/// and accepts a constant format string followed by 0 or more arguments that provide data for the 
+/// format string.
+#define log_skip(skip_category, ...)          log_message_skip(thread_num, skip_category, SANDSTONE_LOG_SKIP __VA_ARGS__)
```
thiagomacieira, line 102: Remove `SANDSTONE_LOG_SKIP` from here.

thiagomacieira (changes requested):
Hello Dilip. This is looking good. I'd like you to fix the nitpicks and rename the SkipCategory enumeration's enumerators, then squash so it can be submitted.

I want to get rid of the "S>", but we can do that in a separate PR to avoid having to rework everything (unless you think you can do this easily).

framework/logging.cpp
dshivaraintel: Done.

framework/logging.cpp
dshivaraintel: Yes, corrected this.

framework/selftest.cpp
```diff
@@ -1067,7 +1136,7 @@ FOREACH_DATATYPE(DATACOMPARE_TEST)
     .test_kvm_config = selftest_kvm_config_real_16bit_fail,
     .desired_duration = -1,
     .flags = test_type_kvm,
-},
+}
```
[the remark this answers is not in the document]
dshivaraintel: No i did not get warning, i deleted this line. It's no longer part of the new commit.

dshivaraintel:
@thiagomacieira removed the `S> ` part and made the suggested changes. Please give it a look one more time and i can go ahead and squash commits.

thiagomacieira (approved):

busykai (approved):

PR #191, opened 2023-02-08
Fixing default values for average cpu load

pwlazlyn:
In case of failure in the process of reading CPU avereage load we should return the system is busy.


lukasz-tuz (approved):
👍🏼

Issue #153, opened 2022-09-09
Possible race condition somewhere in log_platform_message


thiagomacieira:
```yaml
THIS IS AN UNOPTIMIZED BUILD: DON'T TRUST TEST TIMING!
command-line: 'opendcdiag --selftests -e selftest_log_platform --quick -o -'
version: opendcdiag-861692fe19b0-dirty
os: Linux 5.19.2-1-default, glibc 2.36
timing: { duration: 1000.000, timeout: 300000.000 }
cpu-info:
  0: { logical: 0, package: 0, core: 0, thread: 0, family: 6, model: 0x8c, stepping: 1, microcode: 0xa4, ppin: null }
  1: { logical: 1, package: 0, core: 1, thread: 0, family: 6, model: 0x8c, stepping: 1, microcode: 0xa4, ppin: null }
  2: { logical: 2, package: 0, core: 2, thread: 0, family: 6, model: 0x8c, stepping: 1, microcode: 0xa4, ppin: null }
  3: { logical: 3, package: 0, core: 3, thread: 0, family: 6, model: 0x8c, stepping: 1, microcode: 0xa4, ppin: null }
  4: { logical: 4, package: 0, core: 0, thread: 1, family: 6, model: 0x8c, stepping: 1, microcode: 0xa4, ppin: null }
  5: { logical: 5, package: 0, core: 1, thread: 1, family: 6, model: 0x8c, stepping: 1, microcode: 0xa4, ppin: null }
  6: { logical: 6, package: 0, core: 2, thread: 1, family: 6, model: 0x8c, stepping: 1, microcode: 0xa4, ppin: null }
  7: { logical: 7, package: 0, core: 3, thread: 1, family: 6, model: 0x8c, stepping: 1, microcode: 0xa4, ppin: null }
tests:
- test: selftest_log_platform
  details: { quality: beta, description: "Logs platform messages (not related to the test)" }
  state: { seed: 'AES:3eda399208fbab11b49dfa85ce79e91dc125c66df70454ee4b62057a318616e2', iteration: 0, retry: false }
  time-at-start: { elapsed:      0.000, now: !!timestamp '2022-09-09T21:16:48Z' }
  result: pass
  time-at-end:   { elapsed:     16.000, now: !!timestamp '2022-09-09T21:16:48Z' }
  test-runtime: 15.493
  threads:
  - thread: main
    messages:
    - { level: info, text: 'BI> Platform issue: This is an informational platform message from CPU 0' }
[...]

PR #194, opened 2023-02-22
FW/selftests: don't call log_platform_message in run function

thiagomacieira:
log_platform_message() is not thread-safe. See #153, which is closed as "won't fix".


dshivaraintel:
@thiagomacieira looks good to me.

thac0 (approved):
👍

PR #198, opened 2023-02-27
Athenas/ifs/ifs refactor

AthenasJimenez:
Code refactor to make it more flexible and ready for other IFS tests.

### DIFF
--- a/tests/ifs/ifs.c
+++ b/tests/ifs/ifs.c
@@ -93,19 +93,19 @@ static bool load_test_file(int dfd, int batch_fd, struct test *test, ifs_test_t 
     return false;
 }
 
-static int scan_common_init(struct test *test, const char *sys_ifs_dir)
+static int scan_common_init(struct test *test)
 {
-        /* allocate info struct */
+        /* Get info struct */
         ifs_test_t *ifs_info = test->data;
 
         /* see if driver is loaded */
-        int ifs_fd = kernel_driver_is_loaded(sys_ifs_dir);
+        int ifs_fd = kernel_driver_is_loaded(ifs_info->sys_path);
 
         /* see if we can open run_test for writing */
         int run_fd = openat(ifs_fd, "run_test", O_WRONLY);
         int saved_errno = errno;
         if (run_fd < 0) {
-                log_info("could not open %s/run_test for writing (not running as root?): %m", sys_ifs_dir);
+                log_info("could not open %s/run_test for writing (not running as root?): %m", ifs_info->sys_dir);
                 close(run_fd);
                 return -saved_errno;
         }
@@ -117,7 +117,7 @@ static int scan_common_init(struct test *test, const char *sys_ifs_dir)
             int batch_fd = openat(ifs_fd, "current_batch", O_RDWR);
             saved_errno = errno;
             if (batch_fd < 0) {
-                    log_info("could not open %s/current_batch for writing (not running as root?): %m", sys_ifs_dir);
+                    log_info("could not open %s/current_batch for writing (not running as root?): %m", ifs_info->sys_dir);
                     close(batch_fd);
                     return -saved_errno;
             }
@@ -216,10 +216,13 @@ static int scan_run(struct test *test, int cpu)
 static int scan_saf_init(struct test *test)
 {
     ifs_test_t *data = (ifs_test_t *) malloc(sizeof(ifs_test_t));
-    data->image_support = true;
-    test->data = data;
 
-    return scan_common_init(test, "intel_ifs_0");
+    data->sys_dir = "intel_ifs_0";
+    data->sys_path = PATH_SYS_IFS_BASE "intel_ifs_0";
+    data->image_support = true;
+
+    test->data = data;
+    return scan_common_init(test);
 }
 
 DECLARE_TEST(ifs, "Intel In-Field Scan (IFS) hardware selftest")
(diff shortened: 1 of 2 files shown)


tests/ifs/sandstone_ifs.h
```diff
@@ -67,11 +67,16 @@ static ssize_t read_file(int dfd, const char *filename, char buf[static restrict
...
 {
+        /* */
+        char sys_ifs_path[50];
+        strcpy(sys_ifs_path, PATH_SYS_IFS_BASE);
+        strcat(sys_ifs_path, dir);
+
         /* see if driver is loaded, otherwise try to load it */
-        int ifs0 = open(PATH_SYS_IFS_BASE "intel_ifs_0", O_DIRECTORY | O_PATH | O_CLOEXEC);
-        if (ifs0 < 0) {
+        int sys_ifs_fd = open(sys_ifs_path, O_DIRECTORY | O_PATH | O_CLOEXEC);
```
thiagomacieira: One trick here is to use `openat` on the base dir, but I think this is fine.

Please use `strncpy` and `strncat`. Or also
```c
    int n = snprintf(sys_ifs_path, sizeof(sys_ifs_path), PATH_SYS_IFS_BASE "%s", dir);
    assert(n < sizeof(sys_ifs_path));
```

tests/ifs/sandstone_ifs.h
```diff
@@ -16,6 +16,8 @@
 #define IFS_SW_PARTIAL_COMPLETION               0xFE
 
 typedef struct {
+    const char *sys_dir;
+    const char *sys_path;
```
thiagomacieira, line 20: Shouldn't be needed to keep both, as the base is constant.

thiagomacieira (commented):
Looking good.

BTW, you can refer to  the Array BIST patches in https://lore.kernel.org/all/20230214234426.344960-1-jithu.joseph@intel.com/

tests/ifs/sandstone_ifs.h
AthenasJimenez: nice solution, added!

tests/ifs/sandstone_ifs.h
AthenasJimenez: right, I was just being lazy. Fixed now as well. :)

tests/ifs/sandstone_ifs.h
```diff
@@ -0,0 +1,104 @@
...
+static ssize_t read_file(int dfd, const char *filename, char buf[static restrict BUFLEN])
+{
+        int fd = openat(dfd, filename, O_RDONLY | O_CLOEXEC);
+        if (fd < 0)
+            return fd;
+
+        return read_file_fd(fd, buf);
+}
+
+static int kernel_driver_is_loaded(const char *sys_path)
```
thiagomacieira, line 73: Minor:  `is_kernel_driver_loaded`, but better yet: `open_sysfs_ifs_base`

tests/ifs/ifs.c
```diff
@@ -156,149 +94,154 @@ static bool load_test_file(int dfd, int batch_fd, struct test *test, ifs_test_t
...
+        ifs_test_t *ifs_info = test->data;
 
-                /* try opening again now that we've potentially modprobe'd */
-                ifs0 = open(PATH_SYS_IFS_BASE "intel_ifs_0", O_DIRECTORY | O_PATH | O_CLOEXEC);
-        }
+        /* see if driver is loaded */
+        char sys_path[PATH_LEN];
+        int n = snprintf(sys_path, PATH_LEN, PATH_SYS_IFS_BASE "%s", ifs_info->sys_dir);
+        assert(n < sizeof(sys_path));
+        int ifs_fd = kernel_driver_is_loaded(sys_path);
```
thiagomacieira, line 106: Need to check here that it didn't return -1, otherwise the openat() below will fail and we'll get a cryptic "Could not open .../run_test for writing (not running as root?): Bad file descriptor".

So, please insert here:
```c
if (ifs_fd < 0) {
        saved_errno = errno;
        log_info("could not find IFS control files in %s: either IFS is not supported on this system"
                 " or this kernel does not support IFS (%m)", ifs_info->sys_dir);
        return -saved_errno;
}
```

I might add the modprobe error message later.

thiagomacieira (commented):
Looks good. Please remove the Array BIST content from this PR so we can focus on the clean up and get it in ASAP.

tests/ifs/sandstone_ifs.h
AthenasJimenez: Fixed

tests/ifs/ifs.c
AthenasJimenez: Fixed.

tests/ifs/ifs.c
```diff
@@ -106,7 +44,7 @@ static bool load_test_file(int dfd, int batch_fd, struct test *test, ifs_test_t
     enforce_run = get_testspecific_knob_value_uint(test, "enforce_run", -1);
     if (memcmp(status_buf, "fail", strlen("fail")) == 0 && enforce_run != 1 )
     {
-        log_warning("Previous run failure found! Refusing to run");
+        log_warning("Previous run failure found! This test will skip until enforced adding flag: -O ifs.enforce_run=1");
```
AthenasJimenez, line 47: @thiagomacieira I mentioned the enforce_run knob in the log_warning, otherwise it could be tricky to know what to do next.

thiagomacieira: Good idea.

thiagomacieira (approved):

PR #204, opened 2023-03-10
FW/logging: split the crash / OS error result from the result: line

thiagomacieira:
The other result types were always a string, so some parsers failed to
properly extract the details here and ended up blank.

[ChangeLog][Framework] Changed how the YAML output format logs crashes
and other unexpected operating system errors. Previously, they were a
YAML object in the "results" entry, instead of the "pass" or "fail"
string. Now, those will print either "crash" or "operating system error"
for that entry, with the details moved to the "result-details" entry.

### DIFF
--- a/framework/logging.cpp
+++ b/framework/logging.cpp
@@ -1923,7 +1923,7 @@ std::string TapFormatLogger::fail_info_details()
     case STATUS_STACK_BUFFER_OVERRUN:
         return "Stack buffer overrun";
     }
-    return nullptr;
+    return "Unknown";
 #else
     return strsignal(status.extra);
 #endif
@@ -2202,6 +2202,10 @@ void YamlLogger::print_result_line(ChildExitStatus status)
         writeln(real_stdout_fd, indent_spaces(), "- test: ", test->id);
     }
 
+    bool crashed = false;
+    bool coredumped = false;
+    std::string reason;
+
     switch (status.result) {
     case TestSkipped:   // can only be "result: skip"...
     case TestPassed:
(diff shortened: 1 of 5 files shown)


framework/logging.cpp
```diff
@@ -1923,7 +1923,7 @@ std::string TapFormatLogger::fail_info_details()
     case STATUS_STACK_BUFFER_OVERRUN:
         return "Stack buffer overrun";
     }
-    return nullptr;
+    return "unknown";
```
busykai, line 1926: Could you capitalize this?

framework/logging.cpp
```diff
@@ -1948,6 +1949,7 @@ std::string TapFormatLogger::fail_info_details()
     case EX_PROTOCOL:   return "remote error in protocol";
     case EX_NOPERM:     return "permission denied";
     case EX_CONFIG:     return "configuration error";
+    case EXIT_MEMORY:   return "Failed to perform an action due to memory shortage.";
```
busykai, line 1952: Strings are capitalized inconsistently (new ones are, the existing are not).

framework/simd.conf
```diff
@@ -169,7 +169,7 @@ arch=CNL    	PLC
 arch=ICL	SNC
 arch=TGL	WLC
 arch=ADL	GRT
-#arch=RPL	GRT
+arch=RPL	GRT
```
busykai, line 172: Why this change is part of this commit?

framework/logging.cpp
thiagomacieira: They come from different sources... Fixed.

busykai (approved):
Look good!

PR #205, opened 2023-04-07
Added WeightedPickerTestFixture to sandstone_test_utils

brad-kelly:
This commit adds a structure used by tests to select from a list of possible values based on a weighted distribution.  Simply provide a map of value => weights to the constructor and call pick() to get a vlaue from the list.

```
Example:
    WeightedPicker<std::string> picker({
        {"value_a": 1},
        {"value_b": 5},  // This value should come up 5X more
        {"value_c": 0},  // This value should never come up
    });

    ...

    std::string value = picker.pick();
```

### DIFF
--- a/framework/sandstone_test_utils.h
+++ b/framework/sandstone_test_utils.h
@@ -19,6 +19,7 @@
 #define SANDSTONE_TEST_UTILS_H
 #include <sys/mman.h>
 #include <cstdint>
+#include <map>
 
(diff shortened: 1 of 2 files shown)


thac0 (commented):
+1 -- If @busykai or @lukasz-tuz seconds, I'll also approve.

lukasz-tuz:
+1, looks good to me.

thac0 (approved):

PR #211, opened 2023-05-03
Add support for multiple embedded test lists

busykai:
This adds an option to specify multiple test lists to be embedded in the binary using build-time option to meson `-Dbuiltin_test_list` in the format: `<id>:<file>,<id>:<file>`. For example:

```
-Dbuiltin_test_list="default:data/default_test_list.txt,skx:data/cpx_test_list.txt"
```
will add two test lists: "default" and "skx". They can be picked when running OpenDCDiag by specifying an argument to `--use-builtin-test-list`, e.g. `--use-builtin-test-list=skx`. The first list specified is also the default (and named "default" by convention) -- it will be used if no argument to `--use-builtin-test-list` is specified.

The test list file should follow opendcdiag test list file format -- at the very minimum, one test name per line.


framework/scripts/generate_test_list.py
```diff
@@ -0,0 +1,99 @@
...
+            all_tests[l] = 1
+
+    # generate the .cpp file
+    with open(cpp_file, "w") as f:
+        f.write(f'#include "{h_file_basename}"\n\n')
+        decls = [ f"extern struct test _test_{test};" for test in sorted(all_tests.keys()) ]
+        f.write("\n".join(decls))
+        f.write("\n\n")
+        for name, tests in test_lists.items():
+            f.write(f"static constexpr struct test *{name}_test_list[] __attribute__((section(\"{name}_test_list\"))) = {{")
```
thiagomacieira, line 78: Is the section necessary? Especially one without a dot, which will remain in the final executable. I don't think it's a good idea to even consider binary-editing it -- it's a list of pointers.

framework/scripts/generate_test_list.py
```diff
@@ -0,0 +1,99 @@
...
+            test_list = [ f"    &_test_{test}" for test in tests ]
+            f.write(",\n".join(test_list))
+            f.write("\n};\n\n")
+        # write selector function
+        f.write("const struct test * *get_test_list(const char *platform, size_t *size) {\n")
```
thiagomacieira, line 83: Move `const` to between the pointers. Our `struct test` are modifiable, but the test list itself shouldn't be. That should solve the need for the cast you have below.

I'd much prefer a `std::span` here, to avoid an output argument, but I won't insist. x86-64 can return two pointer-sized numeric values in registers.

framework/scripts/generate_test_list.py
```diff
@@ -0,0 +1,99 @@
...
+        if default_name is not None:
+            f.write(f'    if (!platform) {{ *size = sizeof({default_name}_test_list)/sizeof(struct test *); return (const struct test **)&{default_name}_test_list; }}\n')
```
thiagomacieira, line 85: How about:
```
    if (!platform)
        platform = "{default_name}";
```

framework/sandstone.cpp
```diff
@@ -3777,8 +3783,14 @@ int main(int argc, char **argv)
                     logging_printf(LOG_LEVEL_QUIET, "# WARNING: test list is not empty while built-in test list provided.\n");
                 }
             }
-            for (auto& test : ordered_test_list) {
-                add_test(test_list, test);
+            builtin_test_list = get_test_list(builtin_test_list_name, &builtin_test_list_sz);
+            if (!builtin_test_list) {
+                logging_printf(LOG_LEVEL_QUIET,
+                        "# ERROR: no such list '%s'\n", builtin_test_list_name);
```
thiagomacieira, line 3789: Please say '--use-builtin-test-list' somewhere in this message so we can understand what the tool is complaining about when it prints the error.

thiagomacieira (changes requested):
Minor things; architecture looks good.

framework/scripts/generate_test_list.py
busykai: Technically, it's not necessary, but I would like to have them in their separate sections. It's not supposed to be binary editable, but I would like to keep it replaceable (with `--update-section`). I will add a dot to the section name.

framework/scripts/generate_test_list.py
busykai: I had a `std::span` there, but removed it since it wouldn't allow me to place it in a section -- it's initialized via a constructor. I will make changes to `const`.

thiagomacieira: But why do we need the section? This a list of pointers, not of names. It's not something one would want to edit in the first place and I don't think we should support someone attempting to do that either.

framework/scripts/generate_test_list.py
thiagomacieira: If you add a dot, then it won't show up in the binary.

thiagomacieira (dismissed):

framework/sandstone.cpp
```diff
@@ -102,6 +96,15 @@ struct HandleCloser
...
 #endif
 
+#include "sandstone_test_lists.h"
+static size_t builtin_test_list_sz;
+static struct test * const *builtin_test_list
+#if SANDSTONE_BUILTIN_TEST_LIST
+        ; // do nothing here
+#else
+        = { nullptr }; // statically initialize to empty list
+#endif
```
thiagomacieira, line 106: Minor nit: these two variables don't need to be static globals. They're only used when processing the requested test list.

thiagomacieira (approved):
I'm approving so we can test that this works with the internal build of the tool. We can clean the minor things later, possibly when switching to `std::span`

PR #212, opened 2023-05-08
athenas/ifs/unittests

AthenasJimenez:
Adding unittests for IFS test content.

### DIFF
--- a/tests/ifs/unit/ifs_unittests.cpp
+++ b/tests/ifs/unit/ifs_unittests.cpp
@@ -1,0 +1,134 @@
+/*
+ * Copyright 2023 Intel Corporation.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#if defined(__x86_64__) && defined(__linux__)
+
+#include "gtest/gtest.h"
+#include "sandstone.h"
+
+#undef log_warning
+#define log_warning
+#undef log_error
+#define log_error
+#undef log_info
+#define log_info
+#undef log_debug
+#define log_debug
+
+#include "../sandstone_ifs.c"
+
+#undef PATH_SYS_IFS_BASE
+#define PATH_SYS_IFS_BASE
+
+#include "../ifs.c"
+#include "ifs_test_cases.h"
+#include "ifs_unit_utils.h"
+
+struct cpu_info *cpu_info = nullptr;
+int num_cpus() { return IFS_UNIT_NUM_CPUS; }
+
+/*
+ * @brief Setup all structs, directory and files needed by test
+ */
+void *test_setup(ifs_unit setup_data)
+{
+    // Setup dummy sysfs tree
+    setup_sysfs_directory(setup_data);
+
+    // Setup dummy test_t struct
+    test *test_t = (test *) malloc(sizeof(test));
+    test_t->id = setup_data.name;
+
+    // Setup data
+    ifs_test_t *ifs_info = (ifs_test_t *) malloc(sizeof(ifs_test_t));
+    ifs_info->sys_dir = setup_data.name;
+    test_t->data = ifs_info;
+
+    return test_t;
+}
+
+/*
+ * @brief Remove files, directory and free structs memory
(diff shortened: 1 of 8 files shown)


tests/ifs/unit/ifs_unittests.cpp
```diff
@@ -0,0 +1,377 @@
...
+#define PATH_SYS_IFS_BASE
+
+#include "../ifs.c"
+#include "ifs_test_cases.h"
+#include "ifs_unit_utils.h"
+
+/*
+ * @brief Setup all structs, directory and files needed by test
+ */
+void *test_setup(ifs_unit setup_data)
```
lukasz-tuz, line 32: Wondering, is there a reason for which `ifs_unit` is passed as argument instead of `ifs_unit *`?

tests/ifs/unit/ifs_unittests.cpp
```diff
@@ -0,0 +1,377 @@
...
+
+/*
+ * @brief Remove files, directory and free structs memory
+ */
+void test_cleanup(test *test_t, ifs_test_t *ifs_info, ifs_unit setup_data)
+{
+    // Remove files first
+    for (size_t i = 0; i < setup_data.files_sz; i++)
+    {
+        char tmp_path[256];
```
lukasz-tuz, line 57: Maybe use `NAME_MAX` or `PATH_MAX` instead of a number?

lukasz-tuz (commented):
Generally, looks good!

lukasz-tuz:
One additional thought: it'd be great to somehow ensure that all test cases defined in `ifs_test_cases.h` are covered by tests implemented in `ifs_unittests.cpp`. I don't know how to achieve that, yet, but I think it'd be helpful.

tests/ifs/unit/ifs_unittests.cpp
AthenasJimenez: Not really, if you think its important I can do the change.

AthenasJimenez:
> One additional thought: it'd be great to somehow ensure that all test cases defined in `ifs_test_cases.h` are covered by tests implemented in `ifs_unittests.cpp`. I don't know how to achieve that, yet, but I think it'd be helpful.

Uhm... I don't see how that would be useful. Simply counting the number of test cases and the test run should suffice.

tests/ifs/unit/ifs_unittests.cpp
```diff
@@ -0,0 +1,386 @@
...
+ * @test The current_batch file is present, image loading is supported.
+ */
+TEST(IFSRequirements, CurrentBatchFound)
+{
+    // Setup dummy test_t struct
+    test *test_t = (test *) test_setup(reqs_test2);
+    ifs_test_t *ifs_info = (ifs_test_t *) test_t->data;
+
+    // Clean errno, before running
+    errno = 0;
```
AthenasJimenez, line 104: This is needed to prevent errors from previous runs to impact next runs. The thing is, I'm not sure this is the right way to handle it. Any thoughts @lukasz-tuz @thiagomacieira ?

tests/ifs/unit/ifs_unittests.cpp
```diff
@@ -0,0 +1,386 @@
...
+#include "../sandstone_ifs.c"
+
+#undef PATH_SYS_IFS_BASE
+#define PATH_SYS_IFS_BASE
+
+#include "../ifs.c"
+#include "ifs_test_cases.h"
+#include "ifs_unit_utils.h"
+
+struct cpu_info *cpu_info = nullptr;
```
AthenasJimenez, line 29: I'm defining those 2 here, but in the future they could interfere with other unittests that may defined them as well. Suggestions @lukasz-tuz @thiagomacieira ?

thiagomacieira: I don't know. We can deal with that if we have a problem.

thiagomacieira (approved):

PR #217, opened 2023-05-16
Clean up and better interfaces for multi-list support.

busykai:
1. Change C array to std::span.
2. Clean up docs and code.

### DIFF
--- a/framework/scripts/generate_test_list.py
+++ b/framework/scripts/generate_test_list.py
@@ -81,25 +81,25 @@ def main():
         f.write('\n'.join(decls))
         f.write('\n\n')
         for name, tests in test_lists.items():
-            f.write(f'static constexpr struct test * const {name}_test_list_[] = {{')
+            f.write(f'static constexpr struct test * const {name}_test_list[] = {{')
             test_list = [ f'    &_test_{test}' for test in tests ]
             f.write(',\n'.join(test_list))
             f.write('\n};\n\n')
-            f.write(f'static constinit const span<struct test * const> {name}_test_list({name}_test_list_);\n\n');
         # write selector function
-        f.write('const span<struct test * const> *get_test_list(const char *test_list_name) {\n')
+        f.write('optional<const span<struct test * const>> get_test_list(const char *test_list_name) {\n')
         if default_name is not None:
-            f.write(f'    if (!test_list_name) return &{default_name}_test_list;\n')
+            f.write(f'    if (!test_list_name) return {default_name}_test_list;\n')
         for name in test_lists.keys():
-            f.write(f'    if  (!strcmp(test_list_name, "{name}")) return &{name}_test_list;\n')
-        f.write('    return nullptr;\n')
+            f.write(f'    if  (!strcmp(test_list_name, "{name}")) return {name}_test_list;\n')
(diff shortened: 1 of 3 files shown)


framework/scripts/generate_test_list.py
```diff
@@ -74,28 +74,32 @@ def main():
...
         for name, tests in test_lists.items():
-            f.write(f'static constexpr struct test *{name}_test_list[] = {{')
+            f.write(f'static constexpr struct test * const {name}_test_list_[] = {{')
             test_list = [ f'    &_test_{test}' for test in tests ]
             f.write(',\n'.join(test_list))
             f.write('\n};\n\n')
+            f.write(f'static constinit const span<struct test * const> {name}_test_list({name}_test_list_);\n\n');
         # write selector function
-        f.write('struct test * const *get_test_list(const char *platform, size_t *size) {\n')
+        f.write('const span<struct test * const> *get_test_list(const char *test_list_name) {\n')
```
thiagomacieira, line 90: Don't return a pointer to the span. Just return the span.

The list array itself, above, can remain the unchanged. I think you simply return the array in the generated test function code and `std::span` will automatically get its size.

thiagomacieira (changes requested):

framework/scripts/generate_test_list.py
busykai: I need to know if the function returns nothing, hence returning the nullptr. What would you suggest instead?

busykai: Did it with optional which is basically the same as a pointer, but in more characters. :)

framework/scripts/generate_test_list.py
```diff
@@ -55,45 +59,50 @@ def main():
...
-        f.write("struct test * const *get_test_list(const char *platform, size_t *size) {\n")
+        f.write('optional<const span<struct test * const>> get_test_list(const char *test_list_name) {\n')
         if default_name is not None:
-            f.write(f'    if (!platform) {{ *size = sizeof({default_name}_test_list)/sizeof(struct test *); return (struct test * const *)&{default_name}_test_list; }}\n')
+            f.write(f'    if (!test_list_name) return {default_name}_test_list;\n')
         for name in test_lists.keys():
-            f.write(f'    if  (!strcmp(platform, "{name}")) {{ *size = sizeof({name}_test_list)/sizeof(struct test *); return (struct test * const *)&{name}_test_list; }}\n')
-        f.write("    return nullptr;\n")
-        f.write("}\n")
+            f.write(f'    if  (!strcmp(test_list_name, "{name}")) return {name}_test_list;\n')
```
thiagomacieira, line 94: Minor coding style suggestion: don't declare the spans as `static constinit` variables. Just return the actual array directly in the `return` statement.

framework/scripts/generate_test_list.py
```diff
@@ -55,45 +59,50 @@ def main():
...
+        f.write('    return {};\n')
```
thiagomacieira, line 95:
```c++
    return std::nullopt;
```

thiagomacieira (approved):

thiagomacieira (approved):

PR #226, opened 2023-05-26
Tests: Update skip reason message for the below tests:

dshivaraintel:
tests/eigen_sparse/eigen_sparse.cpp
tests/ifs/ifs.c
tests/openssl/openssl_sha.cpp

### DIFF
--- a/tests/ifs/ifs.c
+++ b/tests/ifs/ifs.c
@@ -123,7 +123,7 @@ static int scan_common_init(struct test *test)
         int ifs_fd = open_sysfs_ifs_base(sys_path);
         if (ifs_fd < 0) {
             int saved_errno = errno;
-            log_info("could not find IFS control files in %s: either IFS is not supported on this system"
+            log_skip(OsNotSupportedSkipCategory, "could not find IFS control files in %s: either IFS is not supported on this system"
                      " or this kernel does not support IFS (%m)", ifs_info->sys_dir);
             return -saved_errno;
         }
@@ -134,7 +134,7 @@ static int scan_common_init(struct test *test)
         int enforce_run = get_testspecific_knob_value_uint(test, "enforce_run", -1);
         if (memcmp(status_buf, "fail", strlen("fail")) == 0 && enforce_run != 1 )
         {
-            log_warning("Previous run failure found! This test will skip until enforced adding flag: "
+            log_skip(ResourceIssueSkipCategory, "Previous run failure found! This test will skip until enforced adding flag: "
                         "-O %s.enforce_run=1", test->id);
             return EXIT_SKIP;
         }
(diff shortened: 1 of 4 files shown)


tests/ifs/ifs.c
```diff
@@ -277,6 +279,7 @@ static int scan_run(struct test *test, int cpu)
         int scan_ret = scan_run_helper(test, i);
         if (scan_ret == IFS_EXIT_CANNOT_START)
         {
+            log_skip(CpuNotSupportedSkipCategory, "Skipping. Not supported on core %d", i);
```
thiagomacieira, line 282: You can remove the "Skipping", as it's implied. And we don't need to print the thread number because it comes from the logging header, so just "Not supported on this CPU".

tests/openssl/openssl_sha.cpp
```diff
@@ -112,6 +112,7 @@ static int ssl_sha_init(struct test* test)
         return EXIT_SUCCESS;
     }
     else {
+        log_skip(ResourceIssueSkipCategory, "s_EVP_DigestInit_ex/s_EVP_DigestUpdate/s_EVP_DigestFinal_ex/s_EVP_get_digestbyname one or more of these have failed");
```
thiagomacieira, line 115: Do we get an error string from OpenSSL? @AthenasJimenez

thiagomacieira (changes requested):

thiagomacieira (approved):

thiagomacieira:
Unit test building failed. You need to apply the `#define log_message_skip` trick to make skips go away.

dshivaraintel:
@thiagomacieira please let me know if changes to sandstone_unittests_util.h is the actual change you were expecting. Need your approval again.

tests/openssl/openssl_sha.cpp
AthenasJimenez: No, we don't. But this skip only happen if Sandstone could not load openssl functions. So for me the skip reason should look like:
```
log_skip(ResourceIssueSkipCategory, "Openssl library is not available or the current version is not supported")
```

thiagomacieira: Ack. Please spell "OpenSSL".

thiagomacieira:
> @thiagomacieira please let me know if changes to sandstone_unittests_util.h is the actual change you were expecting. Need your approval again.

It was. Please implement Athenas'  suggestion on the skip message.

tests/openssl/openssl_sha.cpp
dshivaraintel: Done

AthenasJimenez (approved):
LGTM

Issue #213, opened 2023-05-09
disable option does not remove tests previously added to command line with -e option


brad-kelly:
If I use the -e option to specify some tests to run, andthen explicitly disable some subset with the --disable option I would expect those test to be removed, but they are not.

Simple example:
```
% opendcdiag -v --output-format=tap -e zlib --disable=zlib
# opendcdiag -v --output-format=tap -e zlib --disable=zlib
# Operating system: Linux 5.15.0-46-generic, glibc 2.35
ok   1 zlib                     # (alpha test)   <-- marked alpha??
...
```

Why is this important (simple use case):
If I want to run all the eigen tests except those with "double" in the name, I'd expect to do something like:
`opendcdiag -v --output-format=tap -e eigen* --disable=*double*`

Unfortunately this does not work:
```
% opendcdiag -v  --output-format=tap -e eigen* --disable=*double*
# opendcdiag -v  --output-format=tap -e eigen* --disable=*double*
# Operating system: Linux 5.15.0-46-generic, glibc 2.35
ok   1 eigen_gemm_double14      # (alpha test) 
ok   2 eigen_gemm_cdouble_dynamic_square# (alpha test) 
ok   3 eigen_gemm_double_dynamic_square# (alpha test) 
ok   4 eigen_gemm_float_dynamic_square
ok   5 eigen_sparse
ok   6 eigen_svd
```

busykai:
So right now what `--disable` does is not actually remove the test from the prepared list of tests to run, but [sets its quality level to SKIP](https://github.com/opendcdiag/opendcdiag/blob/main/framework/sandstone.cpp#L2474). This results on the `(alpha test)` remark in the output above. While this may work in some cases, it does not work in many others, such as when `--disable` tries to remove a test from the test list file, or exclude some of the tests are enabled using wildcard, or in a simple case when `--enable` and `--disable` specify the same test.

busykai:
This was fixed as a side-effect of test management refactoring and (maybe) command-line option processing.

```
$ opendcdiag -v  --output-format=tap -e eigen* --disable=*double*
THIS IS AN UNOPTIMIZED BUILD: DON'T TRUST TEST TIMING!
# opendcdiag -v --output-format=tap -e eigen* --disable=*double*
# opendcdiag-5f30fb7cc88a; Operating system: Linux 6.9.12-amd64, glibc 2.41
ok   1 eigen_gemm_float_dynamic_square
ok   2 eigen_sparse
[...]

PR #227, opened 2023-05-26
Fix test listing behavior and enabling/disabling tests

busykai:
The tests listing (options -l, --list-tests, --list-groups) are not 
taking into account the set of tests specified on the command line,
either test list, built in tests list, or --enable/--disable options.

Before any option modifying the actual test list were ignored by the 
list options. With this change, the actual tests that would be executed
are listed. However, the default behavior of 'opendcdiag -l' stays the 
same.

Some examples illustrating the new behavior:
```
$ ./builddir/opendcdiag -vv -l -e eigen_svd -e zstd1
1 eigen_svd            "Eigen SVD (Singular Value Decomposition) solving payload, which issues a bunch of matrix multiplies underneath"
2 zstd1                "ZStandard compression test - ZStandard compression and decompression with random data (level 1)" 

Groups:
@compression           "Tests that drive compression routines in various libraries"
  zstd1
@math                  "Tests that perform math using, e.g., Eigen"
  eigen_svd
$ ./builddir/opendcdiag -vv -l -e eigen_svd -e zstd1 --disable=zstd1
1 eigen_svd            "Eigen SVD (Singular Value Decomposition) solving payload, which issues a bunch of matrix multiplies underneath"

Groups:
@compression           "Tests that drive compression routines in various libraries"
@math                  "Tests that perform math using, e.g., Eigen"
  eigen_svd
$ cat testlist
eigen_svd
zstd1
$ ./builddir/opendcdiag -vv -l --test-list-file=testlist
1 eigen_svd            "Eigen SVD (Singular Value Decomposition) solving payload, which issues a bunch of matrix multiplies underneath"
2 zstd1                "ZStandard compression test - ZStandard compression and decompression with random data (level 1)" 

Groups:
@compression           "Tests that drive compression routines in various libraries"
  zstd1
@math                  "Tests that perform math using, e.g., Eigen"
  eigen_svd
$ ./builddir/opendcdiag -vv -l --test-list-file=testlist --disable=eigen_svd
1 zstd1                "ZStandard compression test - ZStandard compression and decompression with random data (level 1)" 

Groups:
@compression           "Tests that drive compression routines in various libraries"
  zstd1
```

Additionally, issue #213 is fixed.

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -2592,6 +2592,8 @@ static struct test *get_next_test(int tc)
     }
 
     auto next_test = test_selector->get_next_test();
+    while (next_test && next_test->quality_level < sApp->requested_quality)
+        next_test = test_selector->get_next_test();
 
     if (next_test == nullptr){
         return get_next_test_iteration();
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -2323,16 +2323,17 @@ static auto collate_test_groups()
         std::vector<const struct test *> entries;
     };
     std::map<std::string_view, Group> groups;
-    for (struct test &test : test_set) {
-        for (auto ptr = test.groups; ptr && *ptr; ++ptr) {
+    for (auto test = test_selector->get_next_test(); test; test = test_selector->get_next_test()) {
+        for (auto ptr = test->groups; ptr && *ptr; ++ptr) {
             Group &g = groups[(*ptr)->id];
             g.definition = *ptr;
-            g.entries.push_back(&test);
+            g.entries.push_back(test);
         }
     }
 
     return groups;
 }
+
 
 static void list_tests(int opt)
 {
@@ -2343,7 +2344,7 @@ static void list_tests(int opt)
     auto groups = collate_test_groups();
     int i = 0;
 
-    for (auto test = test_set.begin(); test != test_set.end(); ++test) {
+    for (auto test = test_selector->get_next_test(); test; test = test_selector->get_next_test()) {
         if (test->quality_level >= sApp->requested_quality) {
             if (include_tests) {
                 if (include_descriptions) {
@@ -3240,6 +3241,8 @@ int main(int argc, char **argv)
     const char *on_crash_arg = nullptr;
 
     // test selection
+    int list_tests_action = 0; /* if we were asked to list tests instead of execute, remember what exactly were requested. */
+    const char *list_tests_action_arg = nullptr;
     const char *test_list_file_path = nullptr;
     bool test_list_randomize = false;
     bool use_builtin_test_list = false;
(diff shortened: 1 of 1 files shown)


framework/sandstone.cpp
```diff
@@ -3707,19 +3709,6 @@ int main(int argc, char **argv)
     if (sApp->total_retest_count < -1 || sApp->retest_count == 0)
         sApp->total_retest_count = 10 * sApp->retest_count; // by default, 100
 
-    load_cpu_info(sApp->enabled_cpus);
```
thiagomacieira, line 3710: (GitHub doesn't let me comment on a line more than 3 lines away from a line you modified)
The thermal monitor and the MCE counter are now initialised before the CPU info is loaded. That's not an issue today, but I'd rather that was kept.

framework/sandstone.cpp
```diff
@@ -3792,6 +3777,29 @@ int main(int argc, char **argv)
...
+    init_shmem(UseSharedMemory);
+    debug_init_global(on_hang_arg, on_crash_arg);
+    pthread_setcancelstate(PTHREAD_CANCEL_DISABLE, nullptr);
+
+    print_application_banner();
+    logging_init_global();
+    cpu_specific_init();
+    random_global_init(seed);
+    background_scan_init();
+
```
thiagomacieira, line 3803: Nitpick: unnecessary extra line. In fact, remove both and let the SSL init be merged into this section.

framework/sandstone.cpp
```diff
@@ -2592,6 +2592,8 @@ static struct test *get_next_test(int tc)
     }
 
     auto next_test = test_selector->get_next_test();
+    while (next_test && next_test->quality_level < sApp->requested_quality)
```
thiagomacieira, line 2595: We already have this code somewhere...

thiagomacieira (dismissed):

framework/sandstone.cpp
busykai: Apparently we don't. Otherwise #213 would not have existed. We have similar code in test listings...

framework/sandstone.cpp
busykai: I generally moved everything related to the execution past modifying the test lists, basically to avoid executing anythin in vain. However, yes, we can keep topology initialization at the beginning.

mrkz:
Actually, this also fixes one of the itches I noticed previously about listing `--beta` tests (i.e: currently, If the `--beta` flag is provided after any `-l` `--list*` option, it's ignored), which is something I was going to send a PR to fix too (but then I noticed this change and it's the exact same change (saving options from the big `switch` and process them afterward), so looking forward to getting this one merged instead.

busykai:
This is a stale PR, closing. The entire test management and option management code has been refactored.

PR #228, opened 2023-05-26
Fix file descriptor insufficiency on Linux

thiagomacieira:
Along with other cleanups to the signal handling.

Linux defaults to 1024 file descriptors by default (a legacy caused by
FD_SETSIZE being 1024). The logging framework needs 1 file descriptor
for each thread for the logging file. If a test required two file
descriptors per thread, as in the example of creating a pipe, we were
limited to 341 processors before starting to get unexpected OS errors.

There are many 8-socket systems commercially available that exceed those
limits. An 8-socket 28-core system of the Cascade Lake generation would
add up to 448 logical processors. An 8-socket 60-core Sapphire Rapids
racks up to 960 logical processors.

This commit assumes that 4 file descriptors per thread will be enough
for all tests, plus a reasonable, baseline overhead. This means we won't
touch the Linux default for anything under 240 logical processors.

This can be tested by forcing the hard limit to less than 256:
```
 $ (ulimit -Hn 255; strace -e prlimit64 ./opendcdiag --selftests -e selftest_pass)
 prlimit64(0, RLIMIT_STACK, NULL, {rlim_cur=8192*1024, rlim_max=RLIM64_INFINITY}) = 0
 prlimit64(0, RLIMIT_NOFILE, NULL, {rlim_cur=255, rlim_max=255}) = 0
 prlimit64(0, RLIMIT_NOFILE, {rlim_cur=256, rlim_max=256}, NULL) = -1 EPERM (Operation not permitted)
 ./opendcdiag: failed to increase file descriptor limit: Operation not permitted
 The number of file descriptors for this system (soft 255, hard 255) is too low for the number of CPUs being tested (8).
 Either increase the number of allowed file descriptors to 256 or decrease the number of CPUs being tested, using taskset(1), or schedtool(1), or the --cpuset command-line option.
 If using bash, type 'help ulimit' for information on increasing the file descriptor limit.
 +++ exited with 4 +++
```

### DIFF
--- a/framework/sysdeps/unix/resource.cpp
+++ b/framework/sysdeps/unix/resource.cpp
@@ -1,0 +1,53 @@
+/*
+ * Copyright 2023 Intel Corporation.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <sandstone_system.h>
+#include <sandstone_config.h>
+#include <sandstone_p.h>
+
+#include <algorithm>
+
+#include <stdlib.h>
+#include <stdio.h>
+#include <sys/resource.h>
+
+void resource_init_global()
+{
+    // rounded up so it doesn't look precise
+    static const int FileDescriptorOverhead = 64;
+    static const int FileDescriptorsPerThread = 4;
+    rlim_t desired_fd_count = num_cpus() * FileDescriptorsPerThread + FileDescriptorOverhead;
+    desired_fd_count = ROUND_UP_TO(desired_fd_count, 256);
+
+    struct rlimit oldlimit;
+    if (getrlimit(RLIMIT_NOFILE, &oldlimit) < 0) {
+        perror("getrlimit");
+        exit(EX_OSERR);
+    }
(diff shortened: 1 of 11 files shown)


framework/sysdeps/unix/resource.cpp
```diff
@@ -0,0 +1,53 @@
...
+            fprintf(stderr, "The number of file descriptors for this system (soft %zu, hard %zu) is "
+                            "too low for the number of CPUs being tested (%d).\n"
+                            "Either increase the number of allowed file descriptors to %zu or "
+                            "decrease the number of CPUs being tested, using taskset(1)"
+#if SANDSTONE_RESTRICTED_CMDLINE
+                            " or schedtool(1).\n"
+#else
+                            ", or schedtool(1), or the --cpuset command-line option.\n"
+#endif
+                            "If using bash, type 'help ulimit' for information on increasing the "
```
busykai, line 47: I think this is excessive and unnecessary.

busykai (approved):
Looks good to me. With a minor comment.

PR #238, opened 2023-06-02
CI/Bats: fix the backtrace selftest again

thiagomacieira:
Looks like on one of our self-hosted runners, even though gdb was able to start debugging /bin/true, it still couldn't attach to the OpenDCDiag child process. So instead of trying to guess whether GDB will work, simply accept it might have failed and detect at runtime when it might have.

### DIFF
--- a/bats/sanity-check/20-selftests.bats
+++ b/bats/sanity-check/20-selftests.bats
@@ -863,10 +863,8 @@ selftest_crash_common() {
     fi
 
     # Check that gdb works
-    have_working_gdb=false
     on_crash_arg=
     if gdb -batch -ex 'python 1' -ex run /bin/true 2>/dev/null; then
-        have_working_gdb=true
         on_crash_arg=--on-crash=backtrace
     fi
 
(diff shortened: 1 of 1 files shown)


AthenasJimenez (approved):

busykai (approved):

busykai:
For the record, I think this modification beats the purpose of the test. It may or may not get the backtrace, we're fine either way now.

thiagomacieira:
> For the record, I think this modification beats the purpose of the test. It may or may not get the backtrace, we're fine either way now.

Fair. Ideally we'd be able to confirm that gdb works first. Have you managed to get to one of the machines that are affected? Can you try this:
```sh
sleep 1h
gdb -batch -pid $! -ex 'python 1' -ex kill
```

thiagomacieira:
This might work. If Y try that on a process that is already being traced, I get:
```
warning: process 755980 is already traced by process 755977
ptrace: Operação não permitida.
The program is not being run.
```

PR #239, opened 2023-06-03
CI: Partially revert "CI/Bats: fix the backtrace selftest again"

thiagomacieira:
This partially reverts commit efbb4e0c454125f3030e618190e3e0bd154abc62 (PR #238), restoring the requirement to test gdb if gdb is found to be working, but changing what we test of gdb. Instead of trying to start and run a process, we instead ask gdb to attach to a running process, which is what our child_debug.cpp does.

### DIFF
--- a/bats/sanity-check/20-selftests.bats
+++ b/bats/sanity-check/20-selftests.bats
@@ -862,10 +862,16 @@ selftest_crash_common() {
         skip "Not executing directly (executing '$SANDSTONE')"
     fi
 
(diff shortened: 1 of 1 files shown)


thiagomacieira:
@busykai can you test that this correctly detects whether gdb will work inside child_debug.cpp?

busykai (approved):
Trying to attach works as opposed to executing the debug program. So this is good.

PR #245, opened 2023-06-16
FW/Mac: fix build on macOS: it doesn't consider itself Unix

thiagomacieira:
Apple doesn't want to be part of the cool kids?

### DIFF
--- a/framework/logging.cpp
+++ b/framework/logging.cpp
@@ -38,7 +38,7 @@
 #include <sys/types.h>
 #include <unistd.h>
 
-#ifdef __unix__
+#ifndef _WIN32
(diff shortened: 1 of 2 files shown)


AthenasJimenez (dismissed):
Not a big issue, but looks more congruent to keep the same condition on both places.
```
#if defined(__unix__) || defined(__APPLE__)
```

thiagomacieira:
> Not a big issue, but looks more congruent to keep the same condition on both places.
> 
> ```
> #if defined(__unix__) || defined(__APPLE__)
> ```

Sure, using the same for both, though it'll be the !Win32.

AthenasJimenez (approved):

PR #246, opened 2023-06-16
FW/logging: use atomics for the message count and data size

thiagomacieira:
log_data() is probably only used by tests, so there's no race problem there. But log_message() may be used on the main thread's messaging, which could cause a race condition.


busykai (approved):
Looks good to me.

PR #248, opened 2023-06-16
FW: introduce test_schedule_sequential to run threads sequentially

thiagomacieira:
Not all tests need to run in parallel. In fact, some of them can't run in parallel (the Intel In-Field Scan ones, for example, due to limitation of the kernel API), so they were marked as sequential.

The `mce_test` test is another case.

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -1201,25 +1201,51 @@ static void restart_init(int iterations)
 {
 }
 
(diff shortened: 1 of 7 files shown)


AthenasJimenez (approved):
LGTM, specially because it removes all the ugliness of scan_run()/scan_run_helper() of IFS test.

PR #252, opened 2023-06-20
FW/child_debug: let the child inform its PID to the parent

thiagomacieira:
Instead of having the parent remember what the child's PID is. This allows grandchild processes to crash and still get debugged.

This is incomplete: if more than one child crashes at the same time,
they will all be stopped on read() at

```c++
        // now wait for the parent process to be done with us
        char c;
        IGNORE_RETVAL(read(crashsocket, &c, sizeof(c)));
```

It's unspecified which one we wake up when write()ing back to them. The next commit will fix that.


thiagomacieira:
I'm thinking of a wholly-different approach for this, to using a pipe + futex.

fenrus75 (approved):

PR #259, opened 2023-06-23
FW/child_debug: add --on-crash=context to dump only the context

thiagomacieira:
It was a portion of "backtrace", but we may want to get the context only without incurring the cost of launching gdb.

### DIFF
--- a/framework/sysdeps/unix/child_debug.cpp
+++ b/framework/sysdeps/unix/child_debug.cpp
@@ -178,7 +178,7 @@ enum CrashPipe {
 
 #if SANDSTONE_CHILD_BACKTRACE
 static uint8_t on_hang_action = print_ps_on_hang;
-static uint8_t on_crash_action;
+static uint8_t on_crash_action = context_on_crash;
 #else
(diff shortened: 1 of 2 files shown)


framework/sysdeps/unix/child_debug.cpp
```diff
@@ -902,6 +902,10 @@ void debug_init_global(const char *on_hang_arg, const char *on_crash_arg)
     if (on_crash_arg) {
         if (strcmp(on_crash_arg, "gdb") == 0) {
             on_crash_action = attach_gdb_on_crash;
+        } else if (strcmp(on_crash_arg, "context") == 0) {
+            on_crash_action = context_on_crash;
+        } else if (strcmp(on_crash_arg, "context+core") == 0 || strcmp(on_crash_arg, "core+context")) {
```
thiagomacieira, line 907: Missing `== 0`, causing:
```
builddir/opendcdiag: --on-crash=core requires gdb but we couldn't find it.
```

fenrus75 (approved):

PR #263, opened 2023-06-26
FW/logging: use std::string concatenation in log_data_common()

thiagomacieira:
Not the most efficient code, but it removes the direct use of stdio.h, in preparation for its removal from logging.cpp. In any event, it's already used in logging_format_data(), the other part of `logging_report_mismatched_data()`.

### DIFF
--- a/framework/logging.cpp
+++ b/framework/logging.cpp
@@ -1323,35 +1323,39 @@ static void log_data_common(const char *message, const uint8_t *ptr, size_t size
     fputc(message_code(Preformatted, LOG_LEVEL_VERBOSE(2)), log);
 
     std::string spaces;
+    std::string buffer;
 
     switch (current_output_format()) {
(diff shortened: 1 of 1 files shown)


framework/logging.cpp
```diff
@@ -1360,20 +1364,21 @@ static void log_data_common(const char *message, const uint8_t *ptr, size_t size
...
+                buffer += ' ';
         }
-        fprintf(log, " %02x", (unsigned)ptr[i]);
+        buffer += stdprintf(" %02x", (unsigned)ptr[i]);
     }
 
-    fputc('\n', log);
-    fputc(0, log);
+    buffer += std::string_view("\n", 2);    // include the null
+    fwrite(buffer.data(), 1, msg.size(), log);
```
thiagomacieira, line 1381: That was `buffer`. I never compiled this commit, because this line went away in another commit of mine.

PR #266, opened 2023-06-27
CI/Bats: add a test to confirm that we are using the cpuset threads

thiagomacieira:
To confirm that we got scheduled where we expected to get scheduled, we ask the OS where we are.

This doesn't work on Windows systems with more than 64 logical processors due how it divides the processors into [Processor Groups](https://learn.microsoft.com/en-us/windows/win32/procthread/processor-groups), as the number we print from the tool does not match anything meaningful in the OS above 64.


fenrus75 (approved):

thiagomacieira:
> Needs more work. In one of our self-hosted machines:
> 
> ```
> Measured: one=97% half=5972% all=10966%
> ```

Rewritten so it doesn't do CPU timing.

pwlazlyn (approved):

PR #272, opened 2023-06-28
FW: Remove the ability to run without shared memory

thiagomacieira:
I added this back in February 2021 to support running with secure enclaves, where each process gets its own memory encryption and thus can't share memory with other processes. It was later used for the early Windows port, before we discovered how to have the child process inherit file descriptors from the parent (you have to use [`_spawnv()`](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/spawnv-wspawnv), not directly [`CreateProcessW()`](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw)), but isn't needed any more there.

Given the upcoming changes to support multiple child processes, this would be a complexity I could live without, so I'm removing. If the need arises, we can bring it back later, adapting as necessary.

[ChangeLog][Framework] Removed the ability to run without shared memory
between the multiple tool processes. This ability existed to support
process isolation in some secure enclave environments; it is now
necessary to configure such environments to keep all the processes in
the same enclave so they can share memory.

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -1230,124 +1230,6 @@ static void run_threads(const struct test *test)
     pthread_attr_destroy(&thread_attr);
 }
 
-namespace {
-struct SharedMemorySynchronization
-{
-    Pipe pipe;
-    static bool needsWork()
-    {
-        return !__builtin_expect(sApp->shared_memory_is_shared, true);
-    }
-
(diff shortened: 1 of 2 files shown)


thiagomacieira:
Will need rebasing on top of #247 and #271

fenrus75 (approved):

fenrus75 (approved):

PR #275, opened 2023-06-28
FW/child_debug: don't decrease the Unix socket buffer sizes

thiagomacieira:
I'm not entirely sure how they account for the two sizes, since Unix socket datagrams are delivered immediately to the peer, so there can be only one buffer size total. From what I can read from the Linux kernel sources (net/unix/af_unix.c), there's a limit on the number of datagrams pending and the sender's SNDBUF-32 specifies the maximum datagram length. The number of pending datagrams appears to be set from the net.unix.max_dgram_qlen sysctl at socket creation and can't be modified.

In any case, fail the whole procedure if we can't raise the limit from unsuitable defaults. The size of the XSAVE state including the AMX is about 10 kB, so it isn't likely to be denied.

### DIFF
--- a/framework/sysdeps/unix/child_debug.cpp
+++ b/framework/sysdeps/unix/child_debug.cpp
@@ -367,25 +367,43 @@ static bool check_gdb_available()
     return false;
 }
 
-static void create_crash_pipe(int xsave_size)
+static bool create_crash_pipe(int xsave_size)
 {
     int socktype = SOCK_DGRAM;
 #ifdef SOCK_CLOEXEC
     socktype |= SOCK_CLOEXEC;
 #endif
     if (socketpair(AF_UNIX, socktype, 0, crashpipe) == -1)
-        return;
+        return false;
 
     // set the buffer sizes
     xsave_size += sizeof(CrashContext::Fixed) + sizeof(CrashContext::mc);
+    xsave_size += 256;      // add some headroom; the Linux kernel subtracts 32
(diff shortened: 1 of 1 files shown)


framework/sysdeps/unix/child_debug.cpp
```diff
@@ -367,25 +367,43 @@ static bool check_gdb_available()
...
+
+    int rcvbuf;
+    socklen_t size = sizeof(rcvbuf);
+    if (getsockopt(crashpipe[CrashPipeParent], SOL_SOCKET, SO_RCVBUF, &rcvbuf, &size) == 0) {
+        if (rcvbuf >= xsave_size)
+            xsave_size = 0;
+    }
+    if (xsave_size) {
+        bool ok = setsockopt(crashpipe[CrashPipeChild], SOL_SOCKET, SO_SNDBUF, &xsave_size, sizeof(xsave_size)) == 0;
+        ok = ok && setsockopt(crashpipe[CrashPipeParent], SOL_SOCKET, SO_RCVBUF, &xsave_size, sizeof(xsave_size)) == 0;
```
fenrus75, line 392: creative but... really?

fenrus75 (approved):

framework/sysdeps/unix/child_debug.cpp
thiagomacieira: The `if` line was too long, so I split.

I don't think this will EVER fail, at least on Linux. If you can't ask for a measly 10kB skb, then you've got bigger problems.

PR #279, opened 2023-06-29
FW/SharedMemory: refactor the per-thread shared memory structures and their initialisation

thiagomacieira:
Instead of doing a simple `memset()` in the parent process, move that to the child process and use dedicated functions to do so. This is slightly worse for now, exactly because it's not a `memset()`: that can do 32, 64, or even 128 bytes per cycle. However, once we have multiple children and the work is divided between them, this work will become parallel.

I've also split the main thread's (to become main threads') data structure from the test threads' data structure, so we can have different fields there. An upcoming field will be the test thread data's offset in the shared memory segment.

### DIFF
--- a/framework/logging.cpp
+++ b/framework/logging.cpp
@@ -1013,7 +1013,7 @@ void logging_flush(void)
     // we don't need to fflush() because all our log files are opened without
     // buffering
     do_flush(file_log_fd);
-    do_flush(cpu_data_for_thread(-1)->log_fd);
+    do_flush(sApp->main_thread_data()->log_fd);
 }
 
 void logging_init(const struct test *test)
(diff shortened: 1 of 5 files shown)


fenrus75 (dismissed):

thiagomacieira:
Waiting for #286.

framework/logging.cpp
```diff
@@ -1624,8 +1621,7 @@ inline AbstractLogger::AbstractLogger(const struct test *test, ChildExitStatus s
         PerThreadData::Common *data = sApp->thread_data(i);
         ThreadState thr_state = data->thread_state.load(std::memory_order_relaxed);
         if (data->has_failed()) {
-            if (data->fail_time != 0 && data->fail_time < earliest_fail)
-                earliest_fail = data->fail_time;
+            earliest_fail = std::min(earliest_fail, data->fail_time);
```
pwlazlyn, line 1624: It uses the fact that "invalid_time_point" is `MonotonicTimePoint::max()`. Can we define a constant for "invalid_time_point" and static_assert that our assumption holds?
This might be painful to debug later if we ever change the invalid time point value.
If we `static_assert` it here we explicitly state the assumption made.

pwlazlyn (approved):
Small thing, maybe I am just paranoid. Feel free to ignore or create a separate PR with it.
Otherwise the whole set looks good.

framework/logging.cpp
thiagomacieira: I can define a constant, but I don't see a way to `static_assert` it.

PR #288, opened 2023-06-30
FW/SharedMemory: don't pass state variables to the child that the child doesn't need

thiagomacieira:
`current_test_duration` was used in the sliding-window slicing code, which was removed in 8f59e10b04ad82724d3bf2975d827f0aeb1ca022 (PR #271).

The other variables don't appear to ever have been needed in the child.

### DIFF
--- a/framework/sandstone_p.h
+++ b/framework/sandstone_p.h
@@ -431,7 +431,6 @@ private:
     F(ud_on_failure)                        \
     F(current_max_loop_count)               \
     F(current_test_endtime)                 \
-    F(current_test_duration)                \
     F(test_time)                            \
     F(max_test_time)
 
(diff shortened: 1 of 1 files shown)


thiagomacieira:
Windows failure looks real. Investigating.

thiagomacieira:
I can reproduce by forcing `-f exec`:
```
$ SANDSTONE="$objdir/opendcdiag -f exec" bats -t bats/sanity-check -f timedpass
1..3
not ok 1 selftest_timedpass -t 250
...
# Numeric test failed:
# query:      /tests/0/test-runtime
# value =     11.793
# expression: value >= 250
```

It ran a 250 millisecond test in less than 12. [ECRAY](https://www.gnu.org/fun/jokes/errno.2.en.html) error?

thiagomacieira:
Won't work.

pwlazlyn (approved):

PR #297, opened 2023-07-07
FW/Topology: rewrite reorder_cpus() with a simple std::sort()

thiagomacieira:
And we can use std::tuple's operator< to perform the comparison for us.

### DIFF
--- a/framework/topology.cpp
+++ b/framework/topology.cpp
@@ -6,6 +6,7 @@
 #include "topology.h"
 #include "sandstone_p.h"
 
+#include <algorithm>
 #include <map>
 #include <string>
 #include <vector>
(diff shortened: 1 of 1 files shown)


framework/topology.cpp
```diff
@@ -152,49 +153,15 @@ static linux_cpu_info &proc_cpuinfo()
...
 static void reorder_cpus()
 {
-    auto find_cpu_idx_on_same_core_and_packet = [](int cpu_idx) {
-        for (int idx=0; idx < num_cpus(); ++idx)
-            if (cpu_info[idx].cpu_number != cpu_info[cpu_idx].cpu_number &&
-                cpu_info[idx].core_id    == cpu_info[cpu_idx].core_id    &&
-                cpu_info[idx].package_id == cpu_info[cpu_idx].package_id)
-                return idx;
-        return -1;
+    static auto tie = [](const struct cpu_info &cpu) {
```
busykai, line 156: Please name it something like `cpu_tie`.

framework/topology.cpp
```diff
@@ -152,49 +153,15 @@ static linux_cpu_info &proc_cpuinfo()
...
-            fprintf(stderr, "unable to find unassigned cpu while generating cpu order");
-            exit(EX_USAGE);
-        }
-        cpu_idx_added_to_reorder_list[next_cpu_idx] = 1;
-        cpu_info_reorder[i] = cpu_info[next_cpu_idx];
-        ++i;
-    }
-    delete[] cpu_info;
-    cpu_info = cpu_info_reorder;
+    std::sort(cpu_info, cpu_info + sApp->thread_count, cpu_compare);
```
busykai, line 164: Please use `num_cpus()` here.

busykai (changes requested):

framework/topology.cpp
thiagomacieira: Done.

framework/topology.cpp
thiagomacieira: Done.

busykai (approved):

PR #298, opened 2023-07-07
FW: make TestResult an enum class and sort according to severity

thiagomacieira:
The enum was only used in the C++ side, so I changed it to an `enum class` to ensure there wasn't an accidental cast to/from integers (which there was). Therefore, I introduced an explicit mapping between those, avoiding unnecessarily clever math. Good compilers are perfectly capable of noticing when your switch labels make no transformation or identical ones for some case labels (see [1]).

Then I re-ordered the entries by severity, with the most severe ones having higher values. When we have multiple child processes, the test's condition will be the one with highest severity:
* anything beats a Skip
* any failure beats a Pass
* any crash causes the failed test to be reported as a crash
* core dumps deserve more attention than crashes that don't dump core
* OOM and OS error conditions mean the tool cannot proceed at all
* timeouts beat even OS problems
* interruptions are because the user pressed Ctrl+C

[1] https://gcc.godbolt.org/z/cr9M1xeM9

Note: TimedOut cannot have value 3, because of the Windows runtimes' `abort()` function behaviour. This is enforced by the switch in `test_result_from_exit_code()`.

### DIFF
--- a/framework/logging.cpp
+++ b/framework/logging.cpp
@@ -121,7 +121,7 @@ public:
     const struct test *test;
     MonotonicTimePoint earliest_fail = MonotonicTimePoint::max();
     ChildExitStatus childExitStatus;
-    TestResult testResult = TestPassed;
+    TestResult testResult = TestResult::Passed;
     int pc = 0;
 };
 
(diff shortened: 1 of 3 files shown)


thiagomacieira:
> > I'll make it an enum class to enforce that.
> 
> Why not right away as a part of this PR/commit?

Because it conflicted with #294 (2fcf24f4fc874aa4c2cb9e719b36a113adb24471). Rebasing now allows me to push this.

fenrus75 (approved):

PR #299, opened 2023-07-08
FW/Topology: improve sorting algorithm by sorting the fields

thiagomacieira:
This is optimised for little-endian CPUs: place the least-significant fields first. This way, the compiler could treat these four 32-bit fields as a single 128-bit integer, provided they were all unsigned. Since they aren't, we help it by making it so.

That means we sort -1 last instead of first, but hopefully there isn't a VM that provides this bad topology data (the tool won't be very effective inside it anyway!).

With GCC, we now generate for the comparison:
```asm
        movq    24(%rsi), %rdx
        movq    24(%rdi), %rcx
        cmpq    %rdx, %rcx
        setb    %al
        je      .L5
        ret
.L5:
        movq    16(%rsi), %rdx
        movq    16(%rdi), %rcx
        cmpq    %rdx, %rcx
        setb    %al
        setne   %dl
        andl    %edx, %eax
        ret
```

Similar for Clang, only without a branch:
```asm
        movq    16(%rdi), %rax
        movq    24(%rdi), %rcx
        xorl    %edx, %edx
        cmpq    16(%rsi), %rax
        setb    %dl
        xorl    %eax, %eax
        cmpq    24(%rsi), %rcx
        setb    %al
        cmovel  %edx, %eax
        retq
```

Which are close enough to the ideal 128-bit comparison:
```asm
        movq    16(%rsi), %rdx
        movq    24(%rdi), %rax
        cmpq    %rdx, 16(%rdi)
        sbbq    24(%rsi), %rax
        setc    %al
        ret
```

To get there, we could do:
```c++
        return (uint128(h) << 64) | l;
```
but meh.

### DIFF
--- a/framework/sandstone.h
+++ b/framework/sandstone.h
@@ -324,9 +324,9 @@ struct cpu_info {
     uint64_t ppin;          ///! Processor ID read from MSR
     uint64_t microcode;     ///! Microcode version read from /sys
     int cpu_number;         ///! Logical processor number as seen by OS
-    int package_id;         ///! Topology info from APIC
-    int core_id;            ///! Topology info from APIC
     int thread_id;          ///! Topology info from APIC
+    int core_id;            ///! Topology info from APIC
+    int package_id;         ///! Topology info from APIC
     struct cache_info cache[3]; ///! Cache info from OS
     uint8_t family;         ///! CPU family (usually 6)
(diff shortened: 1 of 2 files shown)


busykai (approved):
We don't allow any of the ids to remain negative 1, so this should be fine.

PR #300, opened 2023-07-10
FW/child_debug: fix --on-crash=context without gdb

thiagomacieira:
This option was introduced so we could get a crash context without gdb, but we still printed:
```
--on-crash=context requires gdb but we couldn't find it.
```

I developed this functionality in the socket-separation branch, probably after commit 704f1415097aacb73d33b74f56b6f4ea74084e1d which changed the enum. So this commit backports that change too.

### DIFF
--- a/framework/sysdeps/unix/child_debug.cpp
+++ b/framework/sysdeps/unix/child_debug.cpp
@@ -133,11 +133,11 @@ private:
 }
 
 enum CrashAction : uint8_t {
+    kill_on_crash           = 0x00,
     coredump_on_crash       = 0x01,
     context_on_crash        = 0x02,
-    backtrace_on_crash      = 0x06,
     attach_gdb_on_crash     = 0x10,
-    kill_on_crash           = 0x20,
+    backtrace_on_crash      = context_on_crash | attach_gdb_on_crash,
 };
 
 enum HangAction {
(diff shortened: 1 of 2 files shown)


thiagomacieira:
The other issue is that these tests don't run on the GitHub CI because somehow debugging doesn't work.

framework/sysdeps/unix/child_debug.cpp
```diff
@@ -133,11 +133,11 @@ struct CrashContext
 }
 
 enum CrashAction : uint8_t {
+    kill_on_crash           = 0x00,
```
pwlazlyn, line 136: Is there a good reason to define it as 0? Now it will become the only field that cannot be tested with
`if (crash_action & kill_on_crash) { ... }` and has to be tested with `==`.
I am worried that we might shoot ourselves in the foot at some point.

pwlazlyn (commented):
Looks fine otherwise

framework/sysdeps/unix/child_debug.cpp
thiagomacieira: Because "kill" is the absence of any other action. We don't let it dump core, we don't ask for a context, we don't ask for a backtrace and we won't attach gdb.

pwlazlyn: Yeah... `kill_on_crash` was already handled uniquely, so blocking this particular PR doesn't really make it better.

pwlazlyn (approved):

PR #304, opened 2023-07-11
FW/child_debug: stop using ATOMIC_FLAG_INIT in C++

thiagomacieira:
We still use ATOMIC_VAR_INIT in C (tmpfile.c, forkfd.c), but in C++ it's deprecated. Found by Clang/libc++:

```
child_debug.cpp:207:48: warning: macro 'ATOMIC_FLAG_INIT' has been marked as deprecated [-Wdeprecated-pragma]
```


pwlazlyn (approved):

thiagomacieira:
Looks like LWG un-deprecated it. But that doesn't change the fact that Clang 15 is complaining.

PR #317, opened 2023-07-13
FW: make LogicalProcessorSet dynamically-sized

thiagomacieira:
This allows us to go beyond the 1024 logical processor limit that the tool currently has. That number is very close to what an 8-socket Sapphire Rapids may have (896 for 56 cores per socket, 960 for 60 cores per socket) and is exceeded for servers with more than 8 sockets.

Plus, update the `ambient_logical_processor_set()` for Linux and FreeBSD not to use a variable sized to `CPU_SETSIZE` (1024), but instead also allocate dynamically (on the stack).

Last `MAX_THREADS` user bites the dust!

### DIFF
--- a/framework/sysdeps/linux/cpu_affinity.cpp
+++ b/framework/sysdeps/linux/cpu_affinity.cpp
@@ -7,8 +7,7 @@
 
 #include <sched.h>
 #include <stdio.h>
-
-static_assert(CPU_SETSIZE >= LogicalProcessorSet::Size);
+#include <sysexits.h>
 
 #ifdef __linux__
 #include <sys/prctl.h>
@@ -22,9 +21,22 @@ static void set_thread_name(const char *thread_name)
 
 LogicalProcessorSet ambient_logical_processor_set()
 {
-    LogicalProcessorSet result;
-    if (sched_getaffinity(0, sizeof(result.array), reinterpret_cast<cpu_set_t *>(result.array)) != 0)
-        result.clear();
+    int size = CPU_SETSIZE;
+    LogicalProcessorSet result(size);
+    int r = sched_getaffinity(0, result.size_bytes(), reinterpret_cast<cpu_set_t *>(result.array.data()));
+
+    while (r != 0 && errno == EINVAL) {
+        // increase the set size until it stops failing
+        size *= 2;
+        result.unset(LogicalProcessor(size - 1));
+        r = sched_getaffinity(0, result.size_bytes(), reinterpret_cast<cpu_set_t *>(result.array.data()));
+    }
+
+    if (r != 0) {
+        perror("could not get the ambient CPU set with sched_getaffinity()");
+        exit(EX_OSERR);
+    }
+
     return result;
 }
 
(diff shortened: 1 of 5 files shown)


thiagomacieira:
This depends on "FW/SharedMemory: move the enabled_cpus bitset off of the shared memory" which I haven't pushed because it requires some topology work, which I can't push because it depends on some triage work. And that depends on #311.

framework/sysdeps/linux/cpu_affinity.cpp
```diff
@@ -22,9 +21,22 @@ static void set_thread_name(const char *thread_name)
...
 {
-    LogicalProcessorSet result;
-    if (sched_getaffinity(0, sizeof(result.array), reinterpret_cast<cpu_set_t *>(result.array)) != 0)
-        result.clear();
+    int size = CPU_SETSIZE;
+    LogicalProcessorSet result(size);
+    int r = sched_getaffinity(0, result.size_bytes(), reinterpret_cast<cpu_set_t *>(result.array.data()));
+
+    while (r != 0 && errno == EINVAL) {
+        // increase the set size until it works
```
busykai, line 29: typo: "while it works" or "until it does not work"

framework/sysdeps/linux/cpu_affinity.cpp
```diff
@@ -22,9 +21,22 @@ static void set_thread_name(const char *thread_name)
...
+        size *= 2;
+        result.set(LogicalProcessor(size - 1));
```
busykai, line 31: This seems broken to me. It sets the last processor in the set and then returns it as ambient set?.. The rest of the bits are left unset -- it's probing the size, but not ambient set...

framework/sysdeps/linux/cpu_affinity.cpp
thiagomacieira: No, this is correct: we're failing, so we increase until it works. But I'll rephrase.

framework/sysdeps/linux/cpu_affinity.cpp
thiagomacieira: `sched_getaffinity` will overwrite everything. This `.set` call simply enlarges the buffer. But to be clear I'll change it to `.unset`

framework/sysdeps/linux/cpu_affinity.cpp
```diff
@@ -22,9 +21,22 @@ static void set_thread_name(const char *thread_name)
...
+
+    while (r != 0 && errno == EINVAL) {
+        // increase the set size until it stops failing
+        size *= 2;
+        result.unset(LogicalProcessor(size - 1));
+        r = sched_getaffinity(0, result.size_bytes(), reinterpret_cast<cpu_set_t *>(result.array.data()));
+    }
+
+    if (r != 0) {
+        perror("could not get the ambient CPU set with sched_getaffinity()");
```
busykai, line 36: mention of `sched_getaffinity()` is excessive here.

busykai (approved):
Looks good with on minor comment.

framework/sysdeps/linux/cpu_affinity.cpp
thiagomacieira: I like to include the name of the system call or library function that failed. Helps with debugging.

PR #321, opened 2023-07-14
FW/Topology: deprecate the --schedule-by option and enforce it

thiagomacieira:
This is needed because we need our internal threads to be grouped together and we definitely want to run the same child on both logical processors. This does not affect Windows or FreeBSD, because those already sorted the logical processors this way, and it's how the ACPI numbering worked. Raymond Chen even had a [blog post](https://devblogs.microsoft.com/oldnewthing/20230620-00/?p=108358) about this a couple of weeks ago.

[ChangeLog][Framework] The sorting order for logical processors on Linux has changed. Previously, the tool kept the order provided by the OS, but now it will group all cores together, side-by-side. This change is noticeable when the framework prints warnings and errors from the tests: users who may be parsing the output of the tool must be aware that "thread 1" is likely no longer "Linux logical processor 1". This change does not affect Windows nor systems without hyperthreading.

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -3200,16 +3200,6 @@ int main(int argc, char **argv)
             if (sApp->shmem->max_messages_per_thread <= 0)
                 sApp->shmem->max_messages_per_thread = INT_MAX;
             break;
-        case schedule_by_option:
-            if (strcmp(optarg, "thread") == 0) {
-                sApp->schedule_by = SandstoneApplication::ScheduleBy::Thread;
-            } else if (strcmp(optarg, "core") == 0) {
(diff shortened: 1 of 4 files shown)


pwlazlyn (approved):

framework/sandstone.cpp
```diff
@@ -3200,16 +3200,6 @@ int main(int argc, char **argv)
             if (sApp->shmem->max_messages_per_thread <= 0)
                 sApp->shmem->max_messages_per_thread = INT_MAX;
             break;
-        case schedule_by_option:
-            if (strcmp(optarg, "thread") == 0) {
-                sApp->schedule_by = SandstoneApplication::ScheduleBy::Thread;
```
thiagomacieira, line 3205: I forgot to remove the enum.

PR #323, opened 2023-07-14
FW/FreeBSD: delegate the common cpu_affinity.cpp to the Linux version

thiagomacieira:
The two OSes have very similar API, so keep the implementation in one place. I even wrote wrapper functions, but it turns out they weren't needed because the FreeBSD libc provides the wrappers for us, including translating the ERANGE error of "insufficient buffer" for `cpuset_getaffinity()` to Linux's EINVAL for `sched_getaffinity()`.

https://github.com/freebsd/freebsd-src/blob/main/lib/libc/gen/sched_getaffinity.c
https://github.com/freebsd/freebsd-src/blob/main/lib/libc/gen/sched_setaffinity.c

### DIFF
--- a/framework/sysdeps/freebsd/cpu_affinity.cpp
+++ b/framework/sysdeps/freebsd/cpu_affinity.cpp
@@ -3,37 +3,16 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-#include <topology.h>
-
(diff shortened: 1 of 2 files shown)


thiagomacieira:
Self-approving (TrustMe™).

PR #332, opened 2023-07-18
FW: move the between-test checks on MCE and thermal events to mce_check

thiagomacieira:
This removes the parsing of /proc/interrupts on Linux between each test and when the "mce_check" test is run. In any case, the product of those checks was only comments, not properly structured data (because we introduced them for the TAP output format, before we had YAML), which means many tools never caught or stored them at all. Though as a consequence the test running in a child process, its update of the counts won't be committed, so when the test is run again, it'll compare against the original numbers from the tool's start.

The problem with /proc/interrupts is that it increases in size linearly with the number of cores.

This improves start-up time a bit if `--disable=mce_check` is passed. For `perf stat -r500 $objdir/opendcdiag --disable=mce_check --selftests -o - -e selftest_skip`:

| Machine     | Before (ms) | After (ms) |
|-------------|-------------|------------|
| 14c SKX     |       3.94  |      3.46  |
| 2x28c CLX   |      36.40  |     27.83  |
| 2x48c SPR   |     123.86  |     80.00  |

And 1000 runs of `selftest_pass` (not including the start up time above):

| Machine     | Before (ms) | After (ms) |
|-------------|-------------|------------|
| 14c SKX     |       6.83  |      6.52  |
| 2x28c CLX   |      21.45  |     13.91  |
| 2x48c SPR   |      63.73  |     17.23  |

[ChangeLog][mce_check] This test now has a "sticky" behavior: if a problem is found, it is reported every time the test is run, instead of only the first time the problem was detected.

### DIFF
--- a/tests/mce_check/mce_check.cpp
+++ b/tests/mce_check/mce_check.cpp
@@ -62,7 +62,8 @@ static int mce_check_run(struct test *test, int cpu)
         assert(cpu < differences.size());
 
         if (differences[i] != 0) {
-            log_message(i, SANDSTONE_LOG_ERROR "MCE detected");
+            log_message(i, SANDSTONE_LOG_ERROR "MCE detected (%u interrupts since start)",
+                        differences[i]);
             differences[i] = 0;
             ++errorcount;
         }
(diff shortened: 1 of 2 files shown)


tests/mce_check/mce_check.cpp
```diff
@@ -78,6 +78,13 @@ static int mce_check_run(struct test *test, int cpu)
 
     if (errorcount)
         log_platform_message(SANDSTONE_LOG_ERROR "MCE interrupts detected on %d CPUs", errorcount);
+
+    uint64_t thermal_now = sApp->count_thermal_events();
+    if (thermal_now != sApp->last_thermal_event_count) {
+        sApp->last_thermal_event_count = thermal_now;
+        log_platform_message(SANDSTONE_LOG_WARNING "Thermal events detected.");
```
busykai, line 85: Does it make sense to also output the number of events detected? It would give a sense of how severe the overheating was...

thiagomacieira: Well, the number of events tells us little, because the ratio at which they come is not something obvious. But printing the number might help the user knowing which loop of execution those events happened in, so adding that. For MCE too.

busykai (approved):

PR #336, opened 2023-07-20
FW/SharedMemory: remove hardcoded limit on MAX_THREADS

thiagomacieira:
This moves the `cpu_info` array into the shared memory segment as a dynamically-sized C99 Flexible Array Member. That way, the children processes (in exec mode) won't have to re-inspect the system again to find the topology.

In order to insert it, we moved the per-thread data to a dynamic offset in the segment, aligned to a page. This layout will change again, once we support multiple main threads. But the advantage of making it dynamic is that we don't have to depend on `MAX_THREADS` any more, thereby we remove the second-to-last instance of this use (see #317 for the last).

This moves the `enabled_cpus` bitset off the shared memory segment, as we don't need it any more.

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -1027,7 +1027,16 @@ static void attach_shmem(int fd)
 {
     assert(sApp->current_fork_mode() == SandstoneApplication::child_exec_each_test);
 
-    int size = ROUND_UP_TO_PAGE(sizeof(SandstoneApplication::SharedMemory));
+    size_t size;
+    if (struct stat st; fstat(fd, &st) >= 0) {
+        size = st.st_size;
+        assert(size == ROUND_UP_TO_PAGE(size));
+    } else {
+        fprintf(stderr, "internal error: could not get the size of shared memory (fd = %d): %m\n",
+                fd);
+        exit(EX_IOERR);
+    }
+
     void *base = mmap(NULL, size, PROT_READ | PROT_WRITE, MAP_SHARED, fd, 0);
     if (base == MAP_FAILED) {
         perror("internal error: could not map the shared memory file to memory");
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -1076,6 +1076,14 @@ static void attach_shmem(int fd)
     sApp->main_thread_data()->thread_state.exchange(thread_not_started, std::memory_order_acquire);
 }
 
+static void protect_shmem()
+{
+    size_t protected_len = sApp->shmem->thread_data_offset;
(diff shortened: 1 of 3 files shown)


framework/sandstone.cpp
```diff
@@ -1021,13 +1046,23 @@ static void commit_shmem()
...
+    sApp->shmem->total_cpu_count = num_cpus();
 }
 
 static void attach_shmem(int fd)
 {
     assert(sApp->current_fork_mode() == SandstoneApplication::child_exec_each_test);
 
-    int size = ROUND_UP_TO_PAGE(sizeof(SandstoneApplication::SharedMemory));
+    size_t size;
+    if (struct stat st; fstat(fd, &st) >= 0) {
```
busykai, line 1057: hm, I don't see this as a good coding style.

framework/sandstone.cpp
```diff
@@ -1021,13 +1046,23 @@ static void commit_shmem()
...
+        size = st.st_size;
+        assert(size == ROUND_UP_TO_PAGE(size));
+    } else {
+        fprintf(stderr, "internal error: could not get the size of shared memory (fd = %d): %m\n",
```
busykai, line 1061: Why not `perror()`?

framework/sandstone_p.h
```diff
@@ -443,11 +447,13 @@ struct SandstoneApplication::SharedMemory
...
 
-    LogicalProcessorSet enabled_cpus;
+    int total_cpu_count = 0;
+    alignas(64) struct cpu_info cpu_info[];         // C99 Flexible Array Member
 
-    // in/out per-thread data
-    PerThreadData::Main main_thread_data;
-    PerThreadData::Test per_thread[MAX_THREADS];
+    // in/out per-thread data allocated dynamically
+//  alignas(PAGE_SIZE)
```
busykai, line 454: Please remove.

busykai (commented):
Looks ok, but I do have some comments.

framework/sandstone.cpp
thiagomacieira: Since `st` isn't used beyond the `if`, I prefer it.

framework/sandstone.cpp
thiagomacieira: So we can print `fd = %d`.

framework/sandstone_p.h
thiagomacieira: It was intentional. It's not code I commented out, it's a comment to indicate that this content is aligned.

I'll rephrase.

busykai (approved):

PR #338, opened 2023-07-20
FW/Topology: Cleanup build_topology()

thiagomacieira:
Since 9ed9a0c6c1c87304f80af746e2e996803f86304d (PR #321), the `cpu_info` array is sorted, so we know the highest numbered package is last. Additionally, because we sort in unsigned, we know that any unknown package IDs will be stored last too with an id of -1. That means we can resize the `packages` vector only once... though for most people once is all they were seeing anyway.

### DIFF
--- a/framework/topology.cpp
+++ b/framework/topology.cpp
@@ -824,20 +824,24 @@ static void init_topology_internal(const LogicalProcessorSet &enabled_cpus)
 
 static Topology build_topology()
 {
-    std::vector<Topology::Package> packages;
+    struct cpu_info *info = cpu_info;
+    const struct cpu_info *const end = cpu_info + num_cpus();
 
+    std::vector<Topology::Package> packages;
     bool valid_topology = true;
-    for (int i = 0; i < num_cpus(); ++i) {
-        struct cpu_info *info = &cpu_info[i];
+    if (int max_package_id = end[-1].package_id; max_package_id >= 0)
+        packages.reserve(max_package_id + 1);
+    else
+        valid_topology = false;
+
+    for ( ; info != end && valid_topology; ++info) {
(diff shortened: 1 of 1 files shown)


thiagomacieira:
Need to fix this for when there's a gap in package numbering.

busykai:
> Need to fix this for when there's a gap in package numbering.

I believe there are no gaps in package numbering. The enumeration is done by the OS, unlike core numbering, done by the CPU itself (via CPUID).

framework/topology.cpp
```diff
@@ -848,19 +849,14 @@ static Topology build_topology()
...
 
-        if (thr->cpu != -1) {
-            valid_topology = false;
-            break;
-        }
+        if (thr->cpu != -1)
+            return Topology({});
 
-        thr->cpu = i;
+        thr->cpu = info - cpu_info;
```
busykai, line 855: I really see this as an inconsistency. You moved the cycle to be C++ iterator-based to use C pointer arithmetic?..

thiagomacieira: Yes, though this is going away in a future commit (#348)

thiagomacieira:
> I believe there are no gaps in package numbering. The enumeration is done by the OS, unlike core numbering, done by the CPU itself (via CPUID).

There can be if you use `taskset` and don't select consecutive packages.

busykai:
> > I believe there are no gaps in package numbering. The enumeration is done by the OS, unlike core numbering, done by the CPU itself (via CPUID).
> 
> There can be if you use `taskset` and don't select consecutive packages.

Yep. I should have known better since I hit this myself...

busykai (approved):

PR #343, opened 2023-07-24
FW/Win32: Provide C++ operators new and delete

thiagomacieira:
The ones in libstdc++ do call `malloc` and `aligned_alloc` and other suitable functions we do provide. However, on Windows, that only works if you link statically to libstdc++.a. If you're using libstdc++6.dll, the already-linked DLL will instead go to the runtime (either MSVCRT or UCRT), bypassing our allocating and zeroing functions.

All the functions in this file are the same because we never throw and never return null pointers.

### DIFF
--- a/framework/sysdeps/windows/aligned_alloc.cpp
+++ b/framework/sysdeps/windows/aligned_alloc.cpp
@@ -10,6 +10,9 @@
 #include <string.h>
 
 #include <sandstone_p.h>
+
+#include <new>
+
 #include <windows.h>
 
 #define DEFAULT_ALIGNMENT       2*__alignof(void *)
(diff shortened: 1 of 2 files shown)


pwlazlyn (approved):
LGTM, small suggestion though.

Now that we compile this as a C++ file, shouldn't we make the definitions `extern "C"`?
I know it now works, because the `malloc.h` has it, but maybe we should be explicit about it.

thiagomacieira:
> Now that we compile this as a C++ file, shouldn't we make the definitions `extern "C"`? I know it now works, because the `malloc.h` has it, but maybe we should be explicit about it.

Those **have** to be declared in the standard headers. But you're right, the failure mode here is silent.

Will fix.

thiagomacieira:
> > Now that we compile this as a C++ file, shouldn't we make the definitions `extern "C"`? I know it now works, because the `malloc.h` has it, but maybe we should be explicit about it.
> 
> Those **have** to be declared in the standard headers. But you're right, the failure mode here is silent.
> 
> Will fix.

Done.

PR #347, opened 2023-07-26
FW/Logging: check the message counter before fstat()ing the log file

thiagomacieira:
This allows us to avoid a system call per thread, per test. However fast those system calls are (and they are fast on Linux), they add up. Here's a benchmark of the runtime of 1000 x selftest_skip in microseconds, divided by the number of threads that would have been started, with the cost of 1 of them subtracted:

| # threads  |   Before   |   After    |
| ---------: | ---------: | ---------: |
|         24 |       1.92 |       0.65 |
|         48 |       1.60 |       0.62 |
|         96 |       1.50 |       0.98 |
|        192 |       1.48 |       0.70 |

That would seem to indicate the cost of this system call is between 0.5 and 1.2 µs. The gain on Windows may be considerably bigger.

This can also be seen with selftest_pass under the same conditions, though the cost of starting the threads dwarfs the gain:

| # threads  |   Before   |   After    |
| ---------: | ---------: | ---------: |
|         24 |      37.37 |      36.94 |
|         48 |      34.06 |      33.60 |
|         96 |      32.96 |      32.56 |
|        192 |      39.78 |      38.31 |

Finally, for selftest_logs, to show that this isn't causing a meaningful regression when the test does log:

| # threads  |   Before   |   After    |
| ---------: | ---------: | ---------: |
|         24 |      50.47 |      49.85 |
|         48 |      44.95 |      44.76 |
|         96 |      42.84 |      42.94 |
|        192 |      49.12 |      49.71 |

That's between 99.7 and 101.2%.

Note: 192 threads means both sockets / NUMA domains were tested.

### DIFF
--- a/bats/sanity-check/20-selftests.bats
+++ b/bats/sanity-check/20-selftests.bats
@@ -573,7 +573,7 @@ selftest_pass() {
        skip "Not supported"
     fi
     declare -A yamldump
-    sandstone_selftest -vvv -e selftest_logs_options
+    sandstone_selftest -vvv --max-messages=0 -e selftest_logs_options
     [[ "$status" -eq 0 ]]
     test_yaml_regexp "/exit" pass
     test_yaml_regexp "/tests/0/result" pass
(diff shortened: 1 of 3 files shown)


AthenasJimenez (approved):
LGTM

PR #350, opened 2023-07-26
FW: rewrite wait_for_child() as wait_for_children()

thiagomacieira:
We only start one child for now, but now we can wait for more than one.

### DIFF
--- a/framework/logging.cpp
+++ b/framework/logging.cpp
@@ -116,7 +116,7 @@ enum LogTypes {
 class AbstractLogger
 {
 public:
-    AbstractLogger(const struct test *test, ChildExitStatus state);
+    AbstractLogger(const struct test *test, std::span<const ChildExitStatus> state);
 
     const struct test *test;
     MonotonicTimePoint earliest_fail = MonotonicTimePoint::max();
@@ -136,7 +136,7 @@ protected:
 class TapFormatLogger : public AbstractLogger
 {
 public:
-    TapFormatLogger(const struct test *test, ChildExitStatus state)
+    TapFormatLogger(const struct test *test, std::span<const ChildExitStatus> state)
         : AbstractLogger(test, state)
     {}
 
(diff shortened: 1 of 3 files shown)


framework/sandstone.cpp
```diff
@@ -1391,79 +1381,142 @@ static ChildExitStatus wait_for_child(int ffd, intptr_t child, int *tc, const st
...
+                kill(child, SIGKILL);
+        }
+    };
 #elif defined(_WIN32)
-    (void) ffd;
+    auto kill_children = [] { };
+    auto single_wait = [&](milliseconds timeout) {
+        bool bWaitAll = false;
+        auto handles = reinterpret_cast<const HANDLE *>(children.handles.data());
+        DWORD result = WaitForMultipleObjects(children.handles.size(), handles,
```
thiagomacieira, line 1430: [`WaitForMultipleObjects`](https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-waitformultipleobjects) is limited to `MAXIMUM_WAIT_OBJECTS` objects, which is 64.

That's enough for 2048 cores (16 sockets of 128 or 8 sockets of 256 being the closest, worst-case scenario); after that we'll have a problem.

framework/logging.cpp
```diff
@@ -1645,9 +1645,18 @@ format_duration(MonotonicTimePoint tp, FormatDurationOptions opts = FormatDurati
     return format_duration(tp - sApp->current_test_starttime, opts);
 }
 
-inline AbstractLogger::AbstractLogger(const struct test *test, ChildExitStatus state_)
-    : test(test), childExitStatus(state_), testResult(state_.result)
+static TestResult find_most_serious_result(std::span<const ChildExitStatus> results)
```
busykai, line 1648: The name sounds funny. Maybe "worst" is better?

busykai (approved):
This looks good. Cannot tell for the Windows part though.

thiagomacieira:
> This looks good. Cannot tell for the Windows part though.

"It works" suffices for me. We will need to get back to it when we get more than 64 child processes.

PR #363, opened 2023-08-01
FW/Skip reasons: Add DummySkipCategory to skip reasons

dshivaraintel:
This category can be used when majority of the cores skip a test. When this skip category is used, Sandstone will not log any skip messages. This is equivalent to the case when all cores run the test and the test exits without skips.

One of the use cases for this are the accelerator tests where only a handful of cores run the accelerator tests and others skip. By default all cores will log their defualt/user defined skip messages, the final log looks messy. With the changes in this commit, when this skip category is used, Sandstone will not log the default/user defined skip log message.

Before these changes.

```
test: simple_add
  time-at-start: { elapsed:   1183.377, now: !!timestamp '2023-08-01T04:33:24Z' }
  result: pass
  time-at-end:   { elapsed:   1193.377, now: !!timestamp '2023-08-01T04:33:24Z' }
  test-runtime: 9.056
  threads:
  - thread: 0
    id: { logical:  0, package: 0, core:  0, thread: 0, family: 6, model: 0x55, stepping: 6, microcode: 0x400320a, ppin: null }
    messages:
    - { level: skip, text: 'Os not supported category' }
  - thread: 2
    id: { logical:  2, package: 0, core:  2, thread: 0, family: 6, model: 0x55, stepping: 6, microcode: 0x400320a, ppin: null }
    messages:
    - { level: skip, text: 'Os not supported category' }
```
After these changes.

```
- test: simple_add
  time-at-start: { elapsed:    993.370, now: !!timestamp '2023-08-01T00:05:02Z' }
  result: pass
  time-at-end:   { elapsed:   1010.037, now: !!timestamp '2023-08-01T00:05:02Z' }
  test-runtime: 16.848
- test: simple_add
  time-at-start: { elapsed:   1023.371, now: !!timestamp '2023-08-01T00:05:02Z' }
  result: pass
  time-at-end:   { elapsed:   1043.371, now: !!timestamp '2023-08-01T00:05:02Z' }
  test-runtime: 19.000
- test: simple_add
  time-at-start: { elapsed:   1053.372, now: !!timestamp '2023-08-01T00:05:02Z' }
  result: pass
  time-at-end:   { elapsed:   1073.373, now: !!timestamp '2023-08-01T00:05:02Z' }
  test-runtime: 18.406
```

### DIFF
--- a/framework/logging.cpp
+++ b/framework/logging.cpp
@@ -462,6 +462,8 @@ static const char *char_to_skip_category(int val)
         return "OsNotSupportedSkipCategory";
     case SkipCategory(9):
         return "ThreadIssueSkipCategory";
+    case SkipCategory(10):
+        return "DummySkipCategory";
     }
 
     return "NO CATEGORY PRESENT";
@@ -1249,7 +1251,7 @@ void log_platform_message(const char *fmt, ...)
 #undef log_message_skip
 void log_message_skip(int thread_num, SkipCategory category, const char *fmt, ...)
 {
-    if (current_output_format() == SandstoneApplication::OutputFormat::no_output)
+    if (current_output_format() == SandstoneApplication::OutputFormat::no_output || category == DummySkipCategory)
         return;
 
     va_list va;
--- a/bats/sanity-check/20-selftests.bats
+++ b/bats/sanity-check/20-selftests.bats
@@ -328,6 +328,15 @@ selftest_pass() {
     done
 }
 
+@test "selftest_dummy_skip_run_even_threads" {
+    declare -A yamldump
+    sandstone_selftest -e selftest_dummy_skip_run_even_threads
+    [[ "$status" -eq 0 ]]
+    test_yaml_regexp "/exit" pass
+    test_yaml_regexp "/tests/0/test" selftest_dummy_skip_run_even_threads
+    test_yaml_regexp "/tests/0/result" pass
+}
+
 @test "selftest_log_skip_newline" {
     declare -A yamldump
     sandstone_selftest -e selftest_log_skip_newline
--- a/framework/sandstone.h
+++ b/framework/sandstone.h
@@ -96,7 +96,9 @@ typedef enum SkipCategory {
     RuntimeSkipCategory,
     SelftestSkipCategory,
     OsNotSupportedSkipCategory,
-    ThreadIssueSkipCategory
+    ThreadIssueSkipCategory,
+    // Use this category if you are skipping on majority of the cores and don't want to clog your log with skip messages.
+    DummySkipCategory  
 } SkipCategory;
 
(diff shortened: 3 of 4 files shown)


framework/logging.cpp
```diff
@@ -1249,7 +1251,7 @@ void log_platform_message(const char *fmt, ...)
 #undef log_message_skip
 void log_message_skip(int thread_num, SkipCategory category, const char *fmt, ...)
 {
-    if (current_output_format() == SandstoneApplication::OutputFormat::no_output)
+    if (current_output_format() == SandstoneApplication::OutputFormat::no_output || category == DummySkipCategory)
```
jposwiata, line 1254: Will anything be logged if if all threads return SKIP? I expected handling DummySkipCategory somewhere in "postprocessing" phase

dshivaraintel: As long as you don't use DummySkipCategory, the messages will be logged. Only for DummySkipCategory it won't be logged.

If all threads skip, that's a genuine skip right? you will want to use a different log_skip category for that and NOT use DummySkipCategory.

dshivaraintel:
Actually you can even return `EXIT_SUCCESS`. But they are not semantically correct because, even though the threads are not actually running the test, we are returning `EXIT_SUCCESS`.  The "real" skips can have a skip reasons of their own and they can return `EXIT_SKIP` so only the "real" skips are logged by the framework. 

Let me talk to @jposwiata and see if we even need this and for what.

busykai:
> Actually you can even return `EXIT_SUCCESS`. But they are not semantically correct because, even though the threads are not actually running the test, we are returning `EXIT_SUCCESS`. The "real" skips can have a skip reasons of their own and they can return `EXIT_SKIP` so only the "real" skips are logged by the framework.

Yeah, you can return `EXIT_SUCCESS`, but this is bad since the overall test will be only skipped if *all* threads returned skipped. So unused threads must always return skip otherwise the overall status will be wrong. To me it sounds like special `EXIT_UNUSED` is preferable. Otherwise it's a logical mess to maintain...

jposwiata:
> eah, you can return `EXIT_SUCCESS`

For DSA/IAX it is not possible to build empty topology, so at least one thread will have WQ assigned. The test will be reported as SKIPPED only if threads with assigned WQs have issues (e.g. failed mmap()), but such cases will be handled with appropriate log_skip().
As discussed with Dilip, we've agreed that EXIT_SKIP without a log_skip() is sufficient for "nothing to be done" threads, we don't need any special handling (EXIT_SUCCESS is wrong, as Arzhan has pointed out). We just need to agree, that such pattern is OK.

dshivaraintel:
Since Jarek can use EXIT_SKIP without log_skip for such situations, closing this PR for now. Will discuss with Thiago about adding `EXIT_UNUSED` or `EXIT_NOTHING_TO_BE_DONE_HERE`.

PR #364, opened 2023-08-02
framework: process enabled tests as arguments

mrkz:
Previous to this commit, each test enabled needed to be passed with the
`-e` option (multiple times, as tests names are specified).

This commit adds the capability of specifying tests as arguments

Old command syntax:
```
 $ opendcdiag -e test1 -e test2 -e test3 -e test4
```

After commit syntax:
```
 $ opendcdiag -e test1 test2 test3 test4
```
`test1` will be processed with the `-e` option while `test2`, `test3` and `test4` will be processed as arguments

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -3598,6 +3598,13 @@ int main(int argc, char **argv)
         }
     }
 
+    if (optind < argc) {
+        test_selection_strategy = Ordered;
+        while (optind < argc) {
+            add_tests(test_list, argv[optind++]);
+        }
+    }
+
     if (test_list_randomize) {
         if ((test_selection_strategy == Ordered) || (test_selection_strategy == Alphabetical)) {
             test_selection_strategy = NonRepeating;
--- a/bats/sanity-check/20-selftests.bats
+++ b/bats/sanity-check/20-selftests.bats
@@ -72,7 +72,7 @@ test_yaml_regexp() {
 
 @test "TAP output @positive" {
     local tests=`$SANDSTONE --selftests --list-group-members @positive | sed 's,\r$,,'`
-    run $SANDSTONE --output-format=tap --selftests --timeout=15s --disable=mce_check -e @positive
+    run $SANDSTONE --output-format=tap --selftests --timeout=15s --disable=mce_check @positive
     [[ "$status" -eq 0 ]]
     while read line; do
         printf "# %s\n" "$line"
@@ -131,7 +131,7 @@ test_yaml_regexp() {
 }
 
 @test "YAML silent output" {
-    opts="-Y --quick --selftests --quiet --disable=mce_check -e @positive"
+    opts="-Y --quick --selftests --quiet --disable=mce_check @positive"
     $SANDSTONE $opts > $BATS_TEST_TMPDIR/output.yaml
 
     sed -i -e 's/\r$//' $BATS_TEST_TMPDIR/output.yaml
@@ -291,7 +291,7 @@ selftest_pass() {
 
 @test "selftest_log_skip_init" {
     declare -A yamldump
-    sandstone_selftest -e selftest_log_skip_init
+    sandstone_selftest selftest_log_skip_init
     [[ "$status" -eq 0 ]]
     test_yaml_regexp "/exit" pass
     test_yaml_regexp "/tests/0/test" selftest_log_skip_init
@@ -302,7 +302,7 @@ selftest_pass() {
 
 @test "selftest_log_skip_run_all_threads" {
     declare -A yamldump
-    sandstone_selftest -e selftest_log_skip_run_all_threads
+    sandstone_selftest selftest_log_skip_run_all_threads
     [[ "$status" -eq 0 ]]
     test_yaml_regexp "/exit" pass
     test_yaml_regexp "/tests/0/test" selftest_log_skip_run_all_threads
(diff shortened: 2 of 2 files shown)


mrkz:
note to self: `selftests` failing due to missing `-e` option used within those test runs

framework/sandstone.cpp
```diff
@@ -3226,10 +3226,6 @@ int main(int argc, char **argv)
         case disable_option:
             disable_tests(optarg);
             break;
-        case 'e':
```
thiagomacieira, line 3229: Please don't remove `'e'`. This will break everyone.

The rest is fine.

thiagomacieira (changes requested):
Needs testing in bats/sanity-check - converting some of the tests there to run without `-e` would suffice.

busykai:
I don't think this is a correct change. What about `--disable` option? Why it would behave differently? Also, instead of processing arguments to `-e` this consumes "all the rest of the arguments" as enabled tests which is just wrong. So, no from me unless a good reason for this is provided.

mrkz:
> What about --disable option? Why it would behave differently? Also, instead of processing arguments to -e this consumes "all the rest of the arguments" as enabled tests which is just wrong. 

I feel it makes sense since there's also a `--test-list-file` option, which specifies enabled tests using a text file. This sounds to me like providing a list of enabled tests is more frequent than a list to disable tests (i.e: there's no `--disable-test-list-file`)

> So, no from me unless a good reason for this is provided.

This came out as a talk I had with @thiagomacieira  about processing the rest of the arguments (having processed all options), so that it could be useful by piping with `xargs` (as an example).

> Generally, this is not a right thing to do, IMO.

I'm not strongly opinionated on this one (again, I just landed an idea from a conversation), so feel free to share more feedback on what would make more sense to you or close it if it wouldn't work for you.

busykai:
> > So, no from me unless a good reason for this is provided.
> 
> This came out as a talk I had with @thiagomacieira about processing the rest of the arguments (having processed all options), so that it could be useful by piping with `xargs` (as an example).

We have test lists to manage bulky of enabled lists. Giving preference to enabled tests as free positional arguments is not right. `xargs`, by the way, is perfectly capable of prepending `-e` to each item.

> 
> > Generally, this is not a right thing to do, IMO.
> 
> I'm not strongly opinionated on this one (again, I just landed an idea from a conversation), so feel free to share more feedback on what would make more sense to you or close it if it wouldn't work for you.

What we really need is an exclude test file list. People struggle to maintain all the tests they do not want to run at the command line. It is most frequently requested feature and it's really important for us to have as well. So maybe let's sync.

thiagomacieira:
> We have test lists to manage bulky of enabled lists. Giving preference to enabled tests as free positional arguments is not right. `xargs`, by the way, is perfectly capable of prepending `-e` to each item.

Interesting, I didn't know `-I` could do that.
```
$ (print a; print b; print c) | xargs -I@ echo -e@
-ea
-eb
-ec
```

It's also possible with a shell array (at least with Zsh):
```
$ l=(a b c); echo -e${^l}
-ea -eb -ec
```

But it's much nicer if you don't have to deal with that, which I why I asked for this feature.

> What we really need is an exclude test file list. People struggle to maintain all the tests they do not want to run at the command line. It is most frequently requested feature and it's really important for us to have as well. So maybe let's sync.

I agree, but that's orthogonal. We have the file list parser and we can easily use that to get a list of tests to be disabled.

It might also make sense to extend the syntax of the file to support groups, fatal-skips and, why not, disabling.
```
# comment
-test1
@group1
test2
!test3
```

That means "disable test1" (probably because it would otherwise be enabled by `@group1`), then add the roster of tests in that group, then `test2` and finally `test3` in fatal-skip mode.

Maybe `!` is not the best character to indicate importance, as it's usually used in negations, as in git's exclude and attribute files. Plus, if we use the same syntax as the command-line, we wouldn't want to use `-` for negation either. That leaves still a few options, some of which are not even shell special like `%`.

thiagomacieira (approved):
Good for me but don't merge until @busykai is also satisfied.

busykai:
This is a no-go as implemented. To me it's just broken and introduces more weirdness we will have to inevitably deal with. For example, this works:

```bash
opendcdiag -e zstd_aaa -v zlib9
```

so does this:

```bash
opendcdiag zlib1 -e zstd_aaa -v zlib9
```

If we want to salvage the idea (not the implementation), then another approach is needed. We must not modify the behavior of `-e|--enable` instead, we should allow remainder of the option be test specification, so that either:
```
opendcdiag -v zstd_aaa zlib1 zlib9
```
or
```
opendcdiag -v -e zsta_aaa -e zlib1 -e zlib9
```
work. Better yet, I don't recall if it's possible with the standard `getopt_long`, only allow free test arguments after `--`. Additional perk could be that a test name prefixed with `-` would be a request to exclude, i.e.

```bash
opendcdiag -v -- zstd_aaa zlib1 zlib9
opendcdiag -v -e zstd_aaa -- zlib1 zlib9
opendcdiag -v -- -zstd_aaa # run all but zstd_aaa
```
but NOT
```bash
opendcdiag -v zstd_aaa zlib1 zlib9 # missing double hyphen
opendcdiag -v -e zstd_aaa zlib1 # missing double hyphen
opendcdiag  zlib1 -v -e zstd_aaa zlib9 # we must not allow this
```

CC: @mrkz @thiagomacieira

busykai:
This went stale and the base has/will change significantly. If there's a re-work, it should be a new PR.

PR #366, opened 2023-08-03
FW/child_debug: fix build outside of x86-64

thiagomacieira:
We can't #include <cpuid.h>, as that's an x86-64-specific header.


mrkz (approved):
LGTM

PR #307, opened 2023-07-12
FW/Win32: use spawnvp to find gdbserver

thiagomacieira:
We don't pass the full path to it, so we need the "p" version to do a PATH search for us.

PR #369, opened 2023-08-07
FW/Win32: fix use of gdbserver on Windows

thiagomacieira:
I added the feature to test -fexec on Linux, and clearly didn't actually use it on Windows. Commit a4bfe65d5750a8fb1707051988972be1ce6850b4 (PR #307) fixed finding it and I confirmed it did start... but didn't debug why I was getting "invalid parameter passed to C function".

It turns out that the file descriptorinheritance doesn't work through gdbserver (my guess is that the gdbserver implementation didn't use _spawnv, so the `cbReserved2` and `lpReserved2` fields of `STARTUPINFO` weren't set). That means we must pass the actual path so the child process can open the shmem.

Logging will NOT work for the same reason, but at least one can debug what's happening in the child process. It's possible the handles were inherited so it may be possible to fix this.


AthenasJimenez (approved):
LGTM

PR #370, opened 2023-08-07
FW: add the --max-cores-per-slice and the beginnings of slicing

thiagomacieira:
This adds the --max-cores-per-slice mode that limits the number of cores available in a given slice. This option can also be used to enable slicing even we heuristically wouldn't do it. It also un-deprecates the--np-slicing option to force the slicing feature to be disabled.

This PR also includes the ability to have multiple main threads (one per slice).

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -517,10 +517,13 @@ inline void test_the_test_data<true>::test_tests_finish(const struct test *the_t
 
     // check if the test has failed
     bool has_failed = false;
-    for (int i = -1; i < num_cpus() && !has_failed; ++i) {
-        if (sApp->thread_data(i)->has_failed())
+    auto failchecker = [&has_failed](PerThreadData::Common *data, int) {
+        if (data->has_failed())
             has_failed = true;
-    }
+    };
+    for_each_main_thread(failchecker);
+    if (!has_failed)
+        for_each_test_thread(failchecker);
 
     MonotonicTimePoint now = std::chrono::steady_clock::now();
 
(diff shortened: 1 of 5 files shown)


framework/sandstone.cpp
```diff
@@ -1088,6 +1090,38 @@ static void protect_shmem()
...
+        return;
+    };
+    for (std::vector<CpuRange> &plan : sApp->test_plans.plans)
+        plan.clear();
+
+    if (sApp->current_fork_mode() == SandstoneApplication::no_fork || max_cores_per_slice < 0)
+        return set_to_full_system();
+
+    int max_cpu = num_cpus();
+    if (max_cores_per_slice == 0) {
```
busykai, line 1108: Can we do an alias for this? Say `--no-slicing`? Just like `--no-fracture` should be an alias for `--max-test-loop-count=0`, but it's not.

framework/sandstone_p.h
```diff
@@ -332,9 +332,15 @@ struct SandstoneApplication : public InterruptMonitor, public test_the_test_data
         child_exec_each_test,       // when parent is exec_each_test
     };
 
+    struct TestPlans {
```
busykai, line 335: This is not a good name, I believe. It would be much easier to understand if it was called `TestSlice` and `Slice`. At least it would be consistent in terms of terminology.

busykai (commented):
It's hard to run thought experiments on a change like this, but I don't see anything inherently bad with it. Consider changing the name of `TestPlan`, please.

framework/sandstone.cpp
thiagomacieira: Sure, because "no slicing" is actually the value -1, which is confusing.

framework/sandstone_p.h
thiagomacieira: The plan is how you slice. I'll keep "Plan" in this struct but rename the inner fields.

PR #373, opened 2023-08-09
sandstone_ssl: Add our own random generator

AthenasJimenez:
This depends on PR #381.

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -75,6 +75,7 @@
 
 #if SANDSTONE_SSL_BUILD
 #  include "sandstone_ssl.h"
+#  include "sandstone_ssl_rand.h"
 #endif
 
 #ifdef _WIN32
@@ -3681,8 +3682,10 @@ int main(int argc, char **argv)
     }
 
 #if SANDSTONE_SSL_BUILD
-    if (SANDSTONE_SSL_LINKED || sApp->current_fork_mode() != SandstoneApplication::exec_each_test)
+    if (SANDSTONE_SSL_LINKED || sApp->current_fork_mode() != SandstoneApplication::exec_each_test) {
         sandstone_ssl_init();
+        sandstone_ssl_rand_init();
+    }
 #endif
 
     logging_print_header(argc, argv, test_duration(), test_timeout(test_duration()));
--- a/framework/sandstone_ssl_rand.cpp
+++ b/framework/sandstone_ssl_rand.cpp
@@ -1,0 +1,96 @@
+/*
+ * Copyright 2023 Intel Corporation.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+
+#include "sandstone_ssl.h"
+#include "sandstone_ssl_rand.h"
+
+static OSSL_PROVIDER *s_prov;
+
+/* Random context management */
+static void *sandstone_rand_newctx(void *provctx, void *parent,
+                            const OSSL_DISPATCH *parent_calls)
+{}
+
+static void sandstone_rand_freectx(void *ctx)
+{}
+
+/* Random number generator functions */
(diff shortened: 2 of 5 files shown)


framework/sandstone_ssl_rand.cpp
```diff
@@ -0,0 +1,160 @@
...
+    return 1;
+}
+
+// TODO: Add clean up funcitons
+void sandstone_ssl_rand_init()
+{
+    if (s_OSSL_PROVIDER_add_builtin(NULL, "sandstone-rand", sandstone_ssl_rand_provider_init) != 1
+            || s_RAND_set_DRBG_type(NULL, "sand-rand", NULL, NULL, NULL) != 1
+            || (s_prov = s_OSSL_PROVIDER_try_load(NULL, "sandstone-rand", 1)) == NULL )
+        printf("WARNING: Cannot load sandstone random provider");
```
thiagomacieira, line 159: `fprintf` to `stderr`; please start the warning with "%s" / `program_invocation_name`.

framework/sandstone.cpp
```diff
@@ -3755,7 +3756,10 @@ int main(int argc, char **argv)
 
 #if SANDSTONE_SSL_BUILD
     if (SANDSTONE_SSL_LINKED || sApp->current_fork_mode() != SandstoneApplication::exec_each_test)
+    {
```
thiagomacieira, line 3759: Nitpick: `{` to the previous line

framework/sandstone_ssl_rand.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+    p = s_OSSL_PARAM_locate(params, OSSL_RAND_PARAM_STATE);
+    if (p != NULL && !s_OSSL_PARAM_set_int(p, *(int *)ctx))
+        return 0;
+
+    p = s_OSSL_PARAM_locate(params, OSSL_RAND_PARAM_STRENGTH);
+    if (p != NULL && !s_OSSL_PARAM_set_int(p, 500))
+        return 0;
+
+    p = s_OSSL_PARAM_locate(params, OSSL_RAND_PARAM_MAX_REQUEST);
```
thiagomacieira, line 92: What's the max request? Is that data length or the maximum value we return?

framework/sandstone_ssl_rand.cpp
```diff
@@ -0,0 +1,160 @@
...
+static int sandstone_rand_generate(void *ctx, unsigned char *out, size_t outlen,
+                            unsigned int strength, int prediction_resistance,
+                            const unsigned char *addin, size_t addin_len)
+{
+    memset_random(out, outlen);
+
+    return 1;
+}
+
+// TODO: Do something
```
thiagomacieira, line 53: We don't support reseeding. You can pretend to OpenSSL that we are a true RNG (that is, not a DPRNG).

thiagomacieira (commented):
Holy indirection, Batman!

So we register a provider whose only content is an initialisation function. That function produces a new context whose only content is a dispatch table, adding a query function. The query function supports one query only and that returns another table, which points to yet another function table containing the actual operation functions!

Whew! Good work so far!

framework/sandstone_ssl_rand.cpp
AthenasJimenez: Good question, from documentation: "Specifies the maximum number of bytes that can be generated in a single call to OSSL_FUNC_rand_generate".  Do we have a limit for memset_random?

framework/sandstone_ssl_rand.cpp
AthenasJimenez: It's ok, it is not mandatory to support it. So I removed it.

framework/sandstone_ssl_rand.cpp
thiagomacieira: We don't.

thiagomacieira (approved):

PR #380, opened 2023-08-15
FW/test-tests: fix race condition in checking the main thread(s)

thiagomacieira:
We were using for_each_main_thread() in the child process, which is wrong, because we run one child per slice, so there's only one main thread. This was causing some slices in the test-tests mode to conclude the test had failed, not because the test had, but because one of the other slices had marked the test's test as failed.

```yaml
  threads:
  - thread: main
    messages:
    - { level: info, text: 'I> Test memory use: (6076 - 2996) / 4 = 770 kB' }
  - thread: main 1
    messages:
    - { level: info, text: 'I> Test memory use: (6068 - 2868) / 4 = 800 kB' }
    - { level: error, text: 'E> Test ran shorter than expected: 24.472 ms (desired 1000.000 ms, min 750.000 ms)' }
    - { level: info, text: 'I> Inner loop average duration: 5.038 ms' }
  - thread: 4
    id: { logical: 2, package: 0, core: 2, thread: 0, family: 6, model: 0x8c, stepping: 1, microcode: 0xac, ppin: null }
    loop-count: 4
    messages:
    - { level: debug, text: 'd> Sampled iteration timings: 9.058 ms, 5.476 ms, 4.944 ms, 4.303 ms' }
```

Note no threads 0 through 3 and no timing information, but that slice 1 has an error.

Drive-by rename some variables from "group" to "slice".

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -1025,6 +1025,7 @@ static void init_shmem()
     sApp->shmemfd = fd;
     sApp->shmem = new (base) SandstoneApplication::SharedMemory;
     sApp->shmem->thread_data_offset = thread_data_offset;
+    sApp->shmem->main_process_pid = getpid();
 }
 
 static void commit_shmem()
--- a/framework/sandstone_p.h
+++ b/framework/sandstone_p.h
@@ -419,6 +419,9 @@ struct SandstoneApplication : public InterruptMonitor, public test_the_test_data
         return fork_mode;
     }
 
+    bool is_main_process();
+    [[maybe_unused]] bool is_child_process() { return !is_main_process(); }
+
     PerThreadData::Common *thread_data(int thread);
     PerThreadData::Main *main_thread_data(int group = 0) noexcept;
     PerThreadData::Test *test_thread_data(int thread);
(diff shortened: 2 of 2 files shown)


framework/sandstone.cpp
```diff
@@ -517,14 +517,14 @@ inline void test_the_test_data<true>::test_tests_finish(const struct test *the_t
...
-    };
-    for_each_main_thread(failchecker);
-    if (!has_failed)
-        for_each_test_thread(failchecker);
+    // (we're in the child process, so we only have one main thread)
+    bool has_failed = sApp->main_thread_data()->has_failed();
+    if (!has_failed) {
+        for_each_test_thread([&has_failed](PerThreadData::Common *data, int) {
+            if (data->has_failed())
+                has_failed = true;
```
busykai, line 525: maybe a break here?

thiagomacieira: This is not a loop, it's a lambda. I'd have to modify the `for_each_xxx_thread` runner to deal with return codes and decide to continue or not. Not worth the hassle. We will walk over all the threads, but this is debugging code anyway.

framework/sandstone.cpp
```diff
@@ -517,14 +517,14 @@ inline void test_the_test_data<true>::test_tests_finish(const struct test *the_t
...
+        });
+    }
```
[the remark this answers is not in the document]
thiagomacieira: Not in this variable. It's the memory footprint block.
https://github.com/opendcdiag/opendcdiag/blob/2ee0cc4ad436eff37a0da0549f751fe784b48f1b/framework/sandstone.cpp#L529-L547

I don't remember why it's like that. I'll clean up, which may help in codegen too.

PR #381, opened 2023-08-15
athenas/openssl/sandstone ssl fixes

AthenasJimenez:
Multiple additions and some little fixes fon sandstone_ssl.h.

PR #373 depends on this.

### DIFF
--- a/framework/sandstone_ssl.h
+++ b/framework/sandstone_ssl.h
@@ -23,6 +23,18 @@
 #include <openssl/sha.h>
 //#include <openssl/whrlpool.h>
 
+/* Define macros based on OpenSSL defined macros
+ * - openssl/params.h
+ * TODO: Find a better way to fix this */
+# define s_OSSL_PARAM_DEFN(key, type, addr, sz)    \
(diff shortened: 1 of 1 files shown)


framework/sandstone_ssl.h
```diff
@@ -11,17 +11,33 @@
...
+#include <openssl/provider.h>
+#include <openssl/rand.h>
 #include <openssl/rsa.h>
 #include <openssl/sha.h>
 //#include <openssl/whrlpool.h>
 
+/* Define macros based on OpenSSL defined macros
+ * - openssl/params.h
+ * TODO: Find a better way to fix this */
+# define s_OSSL_PARAM_DEFN(key, type, addr, sz)    \
```
thiagomacieira, line 32: We can use macros directly, without `s_`.

thiagomacieira (approved):

framework/sandstone_ssl.h
AthenasJimenez: I guess we can, as long as we re-defined the macros to call the prefixed functions or we are ending up linking  against OpenSSL on all cases.

PR #386, opened 2023-08-18
Added support for floating point knob values

brad-kelly:
This PR adds support for floating point knob values

### DIFF
--- a/framework/logging.cpp
+++ b/framework/logging.cpp
@@ -1453,6 +1453,10 @@ void logging_mark_knob_used(std::string_view key, TestKnobValue value, KnobOrigi
             else
                 return stdprintf("-0x%" PRIx64, -v);
         }
+        std::string operator()(double v)
+        {
+            return stdprintf("%.17g", double(v));
+        }
         std::string operator()(std::string_view v)
         {
             if (v.data() == nullptr) {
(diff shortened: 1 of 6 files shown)


framework/logging.cpp
```diff
@@ -1453,6 +1453,10 @@ void logging_mark_knob_used(std::string_view key, TestKnobValue value, KnobOrigi
             else
                 return stdprintf("-0x%" PRIx64, -v);
         }
+        std::string operator()(double v)
+        {
+            return stdprintf("%f", double(v));
```
thiagomacieira, line 1458: I recommend "%.17g" here. Also possible, "%.*g" with a first argument of `std::numeric_limits<double>::digits10 + 2`.

Why 17 or that +2? Long story...

framework/test_knobs.h
```diff
@@ -31,6 +31,8 @@ uint64_t get_testspecific_knob_value_uint(const struct test *test, const char *k
                                           uint64_t value_if_not_present);
 int64_t  get_testspecific_knob_value_int(const struct test *test, const char *key,
                                          int64_t value_if_not_present);
+double   get_testspecific_knob_value_double(const struct test *test, const char *key,
```
thiagomacieira, line 34: Can you add the macro in the `RESTRICTED_CMDLINE` section above? (GitHub doesn't let me comment there).

thiagomacieira (changes requested):
Looks mostly good.

Can you also add a selftest to bats/sanity-check/20-selftests.bats? Modify `selftest_logs_options`

framework/logging.cpp
brad-kelly: Used the %.17g"  Just seems easier to read

framework/test_knobs.h
brad-kelly: Done.

thiagomacieira (approved):

PR #394, opened 2023-08-23
FW/Skip: Add new skip category "TestConfigurationSkipCategory"

dshivaraintel:
This skip can be used when the test doesn't have enough information/configuration that it needs.

After the changes:

```
result: skip
skip-category: TestConfigurationSkipCategory
skip-reason: 'Not enough information for the test'
```

@eamartin96, @pwlazlyn and @jposwiata requested for this category as their DSA test needs it.

### DIFF
--- a/framework/logging.cpp
+++ b/framework/logging.cpp
@@ -464,6 +464,8 @@ static const char *char_to_skip_category(int val)
         return "OsNotSupportedSkipCategory";
     case SkipCategory(9):
         return "ThreadIssueSkipCategory";
+    case SkipCategory(10):
+        return "TestConfigurationSkipCategory";
     }
 
     return "NO CATEGORY PRESENT";
--- a/framework/sandstone.h
+++ b/framework/sandstone.h
@@ -91,7 +91,8 @@ typedef enum SkipCategory {
     RuntimeSkipCategory,
     SelftestSkipCategory,
     OsNotSupportedSkipCategory,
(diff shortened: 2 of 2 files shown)


eamartin96 (approved):

framework/logging.cpp
```diff
@@ -464,6 +464,8 @@ static const char *char_to_skip_category(int val)
         return "OsNotSupportedSkipCategory";
     case SkipCategory(9):
         return "ThreadIssueSkipCategory";
+    case SkipCategory(10):
```
jposwiata, line 467: BTW. Could we "reverse" the checking logic? I mean:
```
switch (SkipCategory(val)) {
    case ResourceIssueSkipReason:
         return "ResourceIssueSkipReason";
    ...
    default:
         // code unreachable
         return "NotPresentSkipCategory";
}
```

jposwiata (approved):
Looks like solution suitable for our needs. Is there any documentation to add this category, or this is still "in development" stage?

pwlazlyn (changes requested):
I think `ResourceIssueSkipCategory` is enough to cover our needs.

If the environment was in a configuration that does not allow to execute a test correctly, that's a resource issue - not enough cpus, some runtime feature is not available for the given thread, we cannot allocate memory big enough or with some specific characteristics etc.

dshivaraintel:
Closing this PR as this category is not necessary.

PR #403, opened 2023-08-31
FW: move up the CPUID detection to the parent process

thiagomacieira:
This is where it used to be, in the earliest versions of the framework. I moved it to the child on d0edbae7ad0dde41a3a1b2a41ebfc957f16b409f (2021-03-19) in the old repository:
```
    FW: move the per-test check on CPU features and timing into the child

    This means we'll launch a child for tests that will be skipped, thus
    adding a bit of overhead. But it simplifies logging (especially for
    Windows support).
```
Before this, we were getting away with the `FILE *` logging across parent and child on Linux, but that was a ticking time bomb. Therefore, I simplified by moving all the per-thread logging into the child.

Commit 040bc365e8eeddf27b4a4a892dfb2a7a2a5cb912 (#286) removed the `FILE*` -based logging and moreover 8294f1ae3a1cf3503333bace9954af2e6c4b836f (#324) kept the file descriptors open, so we can now easily bring it back up to the parent. This saves us from launching the child processes when we know we're going to skip -- this is important on Windows where starting processes is relatively expensive.

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -814,6 +814,13 @@ static void init_internal(const struct test *test)
     logging_init(test);
 }
 
+static void init_per_thread_data()
+{
+    auto initer = [](auto *data, int) { data->init(); };
+    for_each_main_thread(initer);
+    for_each_test_thread(initer);
+}
+
 static void initialize_smi_counts()
 {
     std::optional<uint64_t> v = sApp->count_smi_events(cpu_info[0].cpu_number);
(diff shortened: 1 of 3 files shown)


framework/sandstone.cpp
```diff
@@ -2039,7 +2030,18 @@ static void run_one_test_children(ChildrenList &children, int *tc, const struct
...
 {
     ChildrenList children;
-    run_one_test_children(children, tc, test);
+    if (uint64_t missing = (test->minimum_cpu | test->compiler_minimum_cpu) & ~cpu_features) {
+        init_per_thread_data();
+
+        // for brevity, don't report the bits that the framework itself needs
+        missing &= ~_compilerCpuFeatures;
+        log_skip(CpuNotSupportedSkipCategory, "test requires %s\n", cpu_features_to_string(missing).c_str());
+        (void) missing;
```
busykai, line 2039: Why is this needed?..

busykai (approved):

framework/sandstone.cpp
thiagomacieira: What "this" are you referring to? 
a) The moving up into the parent process? See commit message
b) the splitting into two functions? To keep Git Blame more or less intact
c) the `(void) missing` line? To avoid a warning about set-but-unused variable under `NO_LOGGING` mode.

PR #417, opened 2023-09-05
FW/Win32: fix waiting_for_children() when waiting for multiple children

thiagomacieira:
We weren't closing / replacing the handle, so after the first child exited, `WaitForMultipleObjects()` would keep signalling the same child every time. As a consequence, we'd never get the results of the other children, but we'd keep counting children_left down.

`WaitForMultipleObjects` fails with `ERROR_INVALID_HANDLE` if one of the handles is `INVALID_HANDLE_VALUE` (duh!) so we can't use that. We can't repeat the same handle either. The expected solution is to rearrange the array, but instead of that we replace with a dummy, non-signalled handle that will be closed by `~ChildrenList`.

Now this reports properly:
```yaml
  - thread: main
    runtime: 18.816
    messages:
  - thread: main 1
    runtime: 1013.116
    messages:
    - { level: error, text: 'E> Child 384 did not exit, using TerminateProcess()' }
```

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -1518,8 +1518,6 @@ struct ChildrenList
                 CloseHandle(h);
         }
     }
-
-    std::vector<intptr_t> handles;
 #else
     ~ChildrenList()
     {
@@ -1530,8 +1528,8 @@ struct ChildrenList
     }
 
     std::vector<pollfd> pollfds;
-    std::vector<pid_t> handles;
 #endif
+    std::vector<pid_t> handles;
     std::vector<ChildExitStatus> results;
 
     void add(StartedChild child)
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -1648,6 +1648,10 @@ static void wait_for_children(ChildrenList &children, int *tc, const struct test
         children.results[idx] = test_result_from_exit_code(result);
         children.results[idx].endtime = MonotonicTimePoint::clock::now();
         --children_left;
+
+        // replace the handle with a valid, dummy, non-signalled handle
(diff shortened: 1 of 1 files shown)


fenrus75 (approved):

framework/sandstone.cpp
```diff
@@ -1650,6 +1648,10 @@ static void wait_for_children(ChildrenList &children, int *tc, const struct test
         children.results[idx] = test_result_from_exit_code(result);
         children.results[idx].endtime = MonotonicTimePoint::clock::now();
         --children_left;
+
+        // replace the handle with a valid, dummy, non-signalled handle
+        CloseHandle(handles[idx]);
+        children.handles[idx] = intptr_t(CreateEvent(nullptr, true, false, nullptr));
```
pwlazlyn, line 1654: [MSDN](https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-waitformultipleobjects) say that lpHandles may not contain multiple copies of the same handle. Maybe we can just create a single, "global" (static) null handle and reuse it?

pwlazlyn (approved):
Looks good, but maybe we can make it a little bit cheaper without really having much more complex code. It's up to you if you want to change it.

thiagomacieira:
> Looks good, but maybe we can make it a little bit cheaper without really having much more complex code. It's up to you if you want to change it.
> 
> Edit: I actually read it wrong: It says may **_not_**. In that case, nevermind ;]

I really want this to be less expensive, but not right now. We can probably fix this at the same time as we fix the limit on the number of handles. I just don't have the time right now to do it properly.

This is a hotfix.

PR #418, opened 2023-09-14
FW/Win32: statelessly map LogicalProcessor to PROCESSOR_NUMBER

thiagomacieira:
Instead of statefully depending on the previous count of members in each group, just store them in the bits of the `LogicalProcessor` variable. This turns the operations that used to be O(n) to O(1).

More importantly, this means we don't depend on stateful data that needed to be initialised. It was only initialised by `ambient_logical_processor_set()`, which the child process stopped calling (probably in commit bbf1ad64067897803fdd612e992aa7c827bf1baa). That was an intentional change, because the child should never need to know the full system's ambient processor set, only what the parent process tells it to care about.

As a side-effect of that, the child process wouldn't have the state initialised, so `processor_group_for()` would always return zero and `processor_in_group()` would return `n` (cast to `BYTE`). That is only correct for systems with a single processor group, which happens to be all the development machines we tested this on. For systems with more than one group, we would fail to map back from `LogicalProcessor` to `PROCESSOR_NUMBER`. Additionally, we'd cause Undefined Behaviour on line:
```c++
     groupAffinity.Mask = KAFFINITY(1) << processorNumber.Number;
```
for logical processors 64 to 255, 320 to 511, etc.

### DIFF
--- a/bats/sanity-check/20-selftests.bats
+++ b/bats/sanity-check/20-selftests.bats
@@ -308,6 +308,9 @@ tap_negative_check() {
 
     # just verify these exist
     for ((i = 0; i < MAX_PROC; ++i)); do
+        if $is_windows; then
+            test_yaml_numeric "/cpu-info/$i/logical-group" 'value >= 0'
+        fi
         test_yaml_numeric "/cpu-info/$i/logical" 'value >= 0'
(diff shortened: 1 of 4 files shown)


busykai (approved):
LGTM.

PR #419, opened 2023-09-14
FW/Win32: re-fix use of gdbserver

thiagomacieira:
Commit 7cf55b1999c6425670f3b0d7eaf16b0a077023d3 changed it so we'd pass a file name instead of a file descriptor and I am sure I tested this, but something must have changed and now I'm getting "Permission denied" when trying to open the tempfile.

But passing handles does work, as speculated in that commit message, so let's do that.


lukasz-tuz (approved):
LGTM

PR #424, opened 2023-09-26
FW/Logging: include the tool's version in the TAP header too

thiagomacieira:
I've found myself wondering about this in some customer logs. Usually they send failures, which will include the version in the failure header, but it's easier if it is near the top.


lukasz-tuz (approved):
Looks good.

PR #425, opened 2023-09-29
C-functions to measure the time (user/sys/wallclock)

jposwiata:
to allow "profiling" the tests from the code (without additional instrumentation).

### DIFF
--- a/framework/sandstone.h
+++ b/framework/sandstone.h
@@ -29,6 +29,7 @@
 #include "sandstone_data.h"
 #include <sandstone_test_groups.h>
 #include "test_knobs.h"
+#include "sandstone_chrono.h"
 
 #ifdef __cplusplus
 #include <atomic>
@@ -95,7 +96,7 @@ typedef enum SkipCategory {
 } SkipCategory;
 
 /// logs a skip message to the logfile. log_skip accepts the category to which the skip belongs to
-/// and accepts a constant format string followed by 0 or more arguments that provide data for the 
+/// and accepts a constant format string followed by 0 or more arguments that provide data for the
 /// format string.
 #define log_skip(skip_category, ...)          log_message_skip(thread_num, skip_category, __VA_ARGS__)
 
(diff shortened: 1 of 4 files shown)


framework/sandstone.cpp
```diff
@@ -660,6 +660,37 @@ inline void test_the_test_data<true>::test_tests_finish(const struct test *the_t
...
+    if (getrusage(RUSAGE_THREAD, &time) < 0) {
+        return 0;
+    }
+    return (1000000LL * time.ru_stime.tv_sec) + time.ru_stime.tv_usec - from;
+#else
+    return 0;
+#endif
+}
+
+uint64_t sandstone_wall_clock_time(uint64_t from)
```
lukasz-tuz, line 689: I think this should be moved to `sandstone_chrono.h`.

framework/sandstone.cpp
```diff
@@ -660,6 +660,37 @@ inline void test_the_test_data<true>::test_tests_finish(const struct test *the_t
 #undef maybe_log_error
 }
 
+uint64_t sandstone_user_cpu_time(uint64_t from)
```
lukasz-tuz, line 663: `sandstone_test_utils.h` feels like a better home for these two functions (`sandstone_user_cpu_time` and `sandstone_sys_cpu_time`). Also, please add unit tests for these.

lukasz-tuz (dismissed):
Is there any benefit for using these functions over a profiler (perf, valgrind, vtune)?

framework/sandstone.cpp
jposwiata: 'chrono' sounds good.

jposwiata:
> Is there any benefit for using these functions over a profiler (perf, valgrind, vtune)?

wall-clock time allows to measure overall processing (with much higher granularity than "thread time"), e.g. allows to measure time of handling a message with inter-thread communication, measure the time without framework overhead, etc.
Cpu time is fully covered by perf/vtune (with much better resolution), and therefore these functions are intended to get "first impression" only, by execution the app without involving mentioned (heavy and difficult to use) tools.

lukasz-tuz (approved):

thiagomacieira:
```
[ RUN      ] SandstoneTime.CpuTime
../opendcdiag/framework/unit-tests/sandstone_utils_tests.cpp:73: Failure
Expected: (proc_sys + proc_user) <= (proc_wall), actual: 4262 vs 4246
```
Please fix.

Issue #429, opened 2023-10-13
Framework restarts the test after a signal is received


thiagomacieira:
This a regression since b7dbacbdb2c509a26ced509dad93389003d1f974 or thereabouts.

### Test
```sh
./opendcdiag --selftests -e selftest_timedpass_busywait -o - & 
sleep 0.1
kill -INT %+
wait
```

### Expected behaviour
OpenDCDiag exits immediately, reports "exit: interrupted" and re-raises the signal that interrupted it.
```
tests:
- test: selftest_freeze
  details: { quality: beta, description: "Freezes" }
  state: { seed: 'AES:8935d753c2767a50eed11b84ca99b5f376ca28ac3d8985af112ee47b35664a0c', iteration: 0, retry: false }
  time-at-start: { elapsed:      0.000, now: !!timestamp '2023-10-12T23:59:13Z' }
# Caught SIGINT, stopping current test (press Ctrl+C again to exit without waiting)                                                                              
  result: interrupted
  fail: { cpu-mask: null, time-to-fail: null, seed: 'AES:8935d753c2767a50eed11b84ca99b5f376ca28ac3d8985af112ee47b35664a0c'}
  time-at-end:   { elapsed:     97.999, now: !!timestamp '2023-10-12T23:59:13Z' }
  test-runtime: 97.508
exit: interrupted
[2]  + interrupt  ./opendcdiag --selftests -e selftest_freeze -Y -o -
```

### Actual behaviour
The current test is terminated but it is restarted again, for retesting, leading to "exit: fail" and exit code 1 instead.
```
$ ./opendcdiag --selftests -e selftest_timedpass_busywait -o - --retest-on-failure=1 & sleep 0.1; kill -INT %+; fg
...
tests:
- test: selftest_timedpass_busywait
  details: { quality: beta, description: "Runs for the requested time, but busy-waiting" }
  state: { seed: 'AES:5a1081f8f55be6ea4cab76a4f271e845a5ef7e070aa41915b354895b0d8e17ba', iteration: 0, retry: false }
  time-at-start: { elapsed:      0.000, now: !!timestamp '2023-10-12T23:58:01Z' }
[1]  + 3638813 running    ./opendcdiag --selftests -e selftest_timedpass_busywait -o - 
# Caught SIGINT, stopping current test (press Ctrl+C again to exit without waiting)
  result: crash
  result-details: { crashed: true, core-dump: false, code: 2, reason: 'Interrupt' }
[...]

PR #430, opened 2023-10-19
FW/Unix: fix handling of signals in wait_for_children()

thiagomacieira:
The rewrite in commit 37cb3e5404eea2847c5da5fe576be0268282b8ee for wait_for_children() accidentally broke the handling of interruptions by signals on Unix systems, which exists to ensure all the processes get interrupted if the parent process is.

Handling of SIGHUP and SIGTERM was easy, since single_wait() was already returning the correct signal number, but wait_for_all_children() wasn't acting on it. Handling of SIGINT required remembering we had been interrupted once.

Fixes #429.

PR #431, opened 2023-10-20
[signal handling] Proper handling for TERM/INT signals

jposwiata:
Signal handlers causes application to stop as soon as required (first TERM causes immediate app stop, first INT/Ctrl-C causes warning, second stops the app with status 'Interrupted'). After single INT/Ctrl-C app reports exit status Interrupt (with appropriate exit code).

It fixes https://github.com/opendcdiag/opendcdiag/issues/429

### DIFF
--- a/framework/logging.cpp
+++ b/framework/logging.cpp
@@ -672,8 +672,15 @@ int logging_close_global(int exitcode)
     if (!SandstoneConfig::NoLogging) {
         if (exitcode != EXIT_SUCCESS) {
             logging_print_log_file_name();
-            logging_printf(LOG_LEVEL_QUIET,
-                           exitcode == EXIT_FAILURE ? "exit: fail\n" : "exit: invalid\n");
+            if (exitcode == EXIT_FAILURE) {
+                logging_printf(LOG_LEVEL_QUIET, "exit: fail\n");
+            }
+            else if (exitcode > 128) {
+                logging_printf(LOG_LEVEL_QUIET, "exit: interrupted\n");
+            }
+            else {
+                logging_printf(LOG_LEVEL_QUIET, "exit: invalid\n");
+            }
         } else if (sApp->shmem->verbosity >= 0) {
             logging_printf(LOG_LEVEL_QUIET, "exit: pass\n");
         }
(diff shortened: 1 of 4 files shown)


framework/logging.cpp
```diff
@@ -1926,6 +1933,8 @@ void TapFormatLogger::print(int tc)
         extra += format_status_code();
         break;
     case TestResult::OutOfMemory:
+        extra += "Out of memory";
+        break;
```
thiagomacieira, line 1937: While correct, we didn't use to do this and should probably leave it as it used to for TAP.

framework/sandstone.cpp
```diff
@@ -326,34 +326,39 @@ static int test_result_to_exit_code(TestResult result)
...
             r.result = TestResult::OperatingSystemError;
+            r.extra = info.status;
             break;
         }
-    } else if (info.code == CLD_KILLED || info.code == CLD_DUMPED) {
-        r.result = info.code == CLD_KILLED ? TestResult::Killed : TestResult::CoreDumped;
-        if (info.status == SIGKILL)
+    } else if (info.code == CLD_KILLED) {
+        if (info.status == SIGINT) {
+            r.result = TestResult::Interrupted;
```
thiagomacieira, line 350: I'm not sure.

This causes the child having received SIGINT for any reason to declare the testing interrupted, while the current code only considers a test interruption if the *parent* process receives the SIGINT or any of the other interruption signals for that matter. No test in our roster (whether open source or not) currently raises SIGINT or SIGTERM for its workload, but it could happen. We do have tests that use the crashing signals (SIGILL, FPE, SEGV and BUS), after all.

That leaves an external source. SIGHUP and SIGINT sent by the terminal are only sent to the controlling process in the session and that's the parent process.

So I'm leaning towards treating an unexpected SIGINT or SIGHUP as a test crash instead of test interruption.

framework/sandstone.cpp
```diff
@@ -1682,24 +1691,23 @@ static void wait_for_children(ChildrenList &children, int *tc, const struct test
...
+            // handle "immediate" signals
+            if (ret > 0) {
+                logging_print_results(children.results, tc, test);
+                if (ret == SIGINT) {
+                    logging_printf(LOG_LEVEL_QUIET, "exit: interrupted\n");
+                } else if (ret == SIGTERM) {
+                    logging_printf(LOG_LEVEL_QUIET, "exit: terminated\n");
+                } else {
+                    logging_printf(LOG_LEVEL_QUIET, "exit: killed\n");
+                }
```
thiagomacieira, line 1703: This changes our reporting `exit` with two new values we don't currently have. Right now, "interrupted" means we stopped the testing before conclusion, regardless of which signal caused it -- it just happens to match the description for SIGINT.

thiagomacieira (changes requested):

jposwiata:
this change is in conflict with Thiago's change on signal handling (https://github.com/opendcdiag/opendcdiag/pull/430).

PR #439, opened 2023-10-26
Handling signals between tests

jposwiata:
If the application is "in between" test execution mode (waits for credits to execute tests), non-fatal signals are handled with usleep() and forces application loop to stop.
Appropriate log is added, the application returns current test execution status (not 'interrupted').

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -2611,12 +2611,11 @@ static struct test *get_next_test(int tc)
     return test;
 }
 
-static void wait_delay_between_tests()
+static bool wait_delay_between_tests()
 {
     useconds_t useconds = duration_cast<microseconds>(sApp->delay_between_tests).count();
-
     // make the system call even if delay_between_tests == 0
-    usleep(useconds);
+    return usleep(useconds) == 0;
 }
 
 static int exec_mode_run(int argc, char **argv)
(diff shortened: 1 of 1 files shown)


framework/sandstone.cpp
```diff
@@ -3801,10 +3805,17 @@ int main(int argc, char **argv)
...
+            if (sApp->service_background_scan) {
+                if (!background_scan_wait()) {
+                    logging_printf(LOG_LEVEL_VERBOSE(2), "# Background scan: waiting between tests interrupted\n");
+                    break;
+                }
+            } else {
+                if (!wait_delay_between_tests()) {
+                    logging_printf(LOG_LEVEL_VERBOSE(2), "# Test execution interrupted between tests\n");
+                    break;
+                }
```
thiagomacieira, line 3817: It feels that this one should do an `exit: interrupted`, maybe both. It's a matter of chance whether you hit the sleep between tests or whether you hit the test, so it shouldn't change the output.

WDYT?

jposwiata: As it doesn't affect any test state, "original" report seems to be better. What more, this is almost identical with "normal" test done.
If we're expected to report "interrupted".. then all signals should be verified and detailed state ("interrupted", "terminated", "killed" with appropriate signal) should be reported.
Overall I think, that composite status of test execution would be best (at the end, e.g.: "failed/interrupted"). This would be much bigger task.

thiagomacieira: Maybe we should arm/disarm the handlers around `wait_for_children` so if these issues happen while sleeping, the process simply exits. The YAML output will be in a good state, though it will miss the "exit:" field.

thiagomacieira: I withdraw that. I still think we need an `exit: interrupted` because we did not run all tests we were planning to. But I can also see how someone using `-T forever` would want to get the result of tests that were run.

PR #440, opened 2023-10-27
Option to fail a test in cleanup.

jposwiata:
Some tests (e.g. with locks) requires post-run checkin, if anything is inconsistent, test is expected to fail.
Both patterns (logging error(s) and returning non-SUCCESS value) from test_cleanup function are handled properly and report appropriate state in the log. Potential group SKIPs are handled in limited scope.

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -1782,8 +1782,20 @@ static TestResult child_run(/*nonconst*/ struct test *test, int child_number)
         if (sApp->shmem->use_strict_runtime && wallclock_deadline_has_expired(sApp->endtime)){
             // skip cleanup on the last test when using strict runtime
         } else {
-            if (test->test_cleanup)
-                test->test_cleanup(test);
+            if (test->test_cleanup) {
+                ret = test->test_cleanup(test);
+                if (state == TestResult::Passed) {
+                    if (ret == EXIT_SKIP) {
+                        log_skip(RuntimeSkipCategory, "SKIP requested in cleanup");
+                        state = TestResult::Skipped;
+                    } else if (ret < EXIT_SUCCESS) {
+                        log_skip(RuntimeSkipCategory, "Unexpected OS error in cleanup: %s", strerror(-ret));
+                        state = TestResult::Skipped;
+                    } else if (ret > EXIT_SUCCESS) {
+                        state = TestResult::Failed;
+                    }
+                }
+            }
         }
 
         sApp->test_tests_finish(test);
--- a/bats/sanity-check/20-selftests.bats
+++ b/bats/sanity-check/20-selftests.bats
@@ -452,6 +452,50 @@ selftest_pass() {
     test_yaml_regexp "/tests/0/skip-reason" '.*skip.*'
 }
 
+@test "selftest_skip_cleanup" {
+    declare -A yamldump
+    sandstone_selftest -e selftest_skip_cleanup
+    [[ "$status" -eq 0 ]]
+    test_yaml_regexp "/exit" pass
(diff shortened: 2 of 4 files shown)


framework/sandstone.cpp
```diff
@@ -1784,8 +1784,20 @@ static TestResult child_run(/*nonconst*/ struct test *test, int child_number)
...
             // skip cleanup on the last test when using strict runtime
         } else {
-            if (test->test_cleanup)
-                test->test_cleanup(test);
+            if (test->test_cleanup) {
+                ret = test->test_cleanup(test);
+                if (state == TestResult::Passed) {
+                    if (ret == EXIT_SKIP) {
+                        log_skip(RuntimeSkipCategory, "Skip requested in cleanup");
+                        state = TestResult::Skipped;
```
thiagomacieira, line 1792: Does this make sense? If at least one thread ran, we didn't skip...

framework/sandstone.cpp
```diff
@@ -1784,8 +1784,20 @@ static TestResult child_run(/*nonconst*/ struct test *test, int child_number)
...
+                    } else if (ret < EXIT_SUCCESS) {
+                        log_skip(RuntimeSkipCategory, "Unexpected OS error in cleanup: %s", strerror(-ret));
+                        state = TestResult::Skipped;
```
thiagomacieira, line 1795: This is a skip, but it shouldn't say "Unexpected OS error".

thiagomacieira (changes requested):
Good fix. Can you add a bats test too?

framework/sandstone.cpp
jposwiata: It could be usable in producer-consumer type tests or with tests which uses accelerators (with some "cpu processing" threads), as overall test is skipped only when all threads in run() return SKIP. Detecting that the test has "done nothing" might be simplest in cleanup(), custom logic seems to be easiest one.

framework/sandstone.cpp
jposwiata: This is special case, identical with existing per init() implementation.

framework/sandstone.cpp
dilipshivaraju: @jposwiata , if i understood this right, if a test skips on only one 1 cpu in the run function, the test is still considered a `pass`. Now if the same test returns `EXIT_SKIP` in cleanup, does this not change the result of the test to be `result:skip` instead of `result:pass` in the original implementation?

framework/sandstone.cpp
thiagomacieira: Ok, I think this is acceptable though.

framework/sandstone.cpp
thiagomacieira: I think it does, but that may be the intention. The `run` functions may not be able to tell that anything useful was generated, only the `cleanup` would. Hopefully, those `run` threads didn't do much work in the first place, otherwise we just wasted CPU time.

thiagomacieira (approved):

thiagomacieira (dismissed):
Changing from approval to changes: need selftest.

framework/sandstone.cpp
jposwiata: Exactly, this is the intention. SKIP is very tricky case, I think it is (might be) used for statistical reasons (to not increase number of passes).
Speaking about "wasted time".. Identical situation is when memory allocation (with mmap) reaches limits: all threads should return -ENOMEM ("unexpected OS error"), so test will SKIP, but (potentially) a lot of work is done. (BTW, I wonder why malloc() causes test to fail, while mmap() is expected to skip..)

thiagomacieira (approved):

PR #446, opened 2023-11-08
FW/Topology: Add odd/even cpu option to --cpuset

rsagashe03:


framework/topology.cpp
```diff
@@ -722,6 +730,14 @@ void apply_cpuset_param(char *param)
...
             }
             add_to_set(*cpu);
+        } else if ( strcmp( p.data(), "odd") == 0 || strcmp( p.data(), "even") == 0){
+            for (struct cpu_info &cpu : old_cpu_info)
+            {
+                if ((strcmp(p.data(), "even") == 0) && take_cpu(cpu.cpu_number, EVEN))
+                    add_to_set(cpu);
+                if ((strcmp(p.data(), "odd") == 0) && take_cpu(cpu.cpu_number, ODD))
+                    add_to_set(cpu);
+            }
```
thiagomacieira, line 740: `strcmp` is marked in glibc as being a pure function and the compiler is probably smart enough to realise that the comparison results can't change, but why not help it along the way?
```c++
    int desired_remainder = strcmp(p.data(), "odd") == 0 ? 1 : 0;
    for (struct cpu_info &cpu : old_cpu_info) {
        if ((cpu.cpu_number % 2) == desired_remainder)
            add_to_set(cpu);
    }
```

thiagomacieira (changes requested):
Need a bats test. Add next to the other "cpuset=" tests.

busykai:
Whatever this is for? Doesn't this imply Linux CPU numbering (on Windows there's no through numbering)? How is it suppose to interact with other, conflicting options specified on the same `--cpuset`? Is there a real use case for this?

bats/sanity-check/20-selftests.bats
```diff
@@ -1667,10 +1669,23 @@ selftest_cpuset_unsorted() {
     local -a cpuset=($($SANDSTONE --dump-cpu-info | sort -rnk2,3 |
                            awk '/^[0-9]/ { printf "%d,", $1; }'))
     cpuset=${cpuset%,}          # remove last comma
+    local -a cpuinfo=(`$SANDSTONE --dump-cpu-info | awk '/^[0-9]/ { print $1; }'`)
 
     selftest_cpuset_unsorted "$cpuset" "${cpuinfo[@]}"
```
thiagomacieira, line 1674: This should cause the test function to be called like
`selftest_cpuset_unsorted "3,2,1,0" 0 1 2 3`

So it shouldn't work the way you modified the function, with:
```sh
    local cpuinfo=$1
```
That would only see one of the values.

So I'm surprised the tests passed

bats/sanity-check/20-selftests.bats
```diff
@@ -1667,10 +1669,23 @@ selftest_cpuset_unsorted() {
...
 }
 
+@test "cpuset=even" {
+    #Get the even number of logical cpus
+    local -a cpuinfo=(`$SANDSTONE --dump-cpu-info | awk '$1%2==0' | awk  '/^[0-9]/ {printf "%d " ,$1}'`)
```
thiagomacieira, line 1679: You can combine the two conditions in one awk:
```sh
 awk '/^[0-9/ && $1 % 2 == 0 { print $1 }'
```

thiagomacieira (commented):
I need to download and understand what's happening in the script. It shouldn't work, but is working...

bats/sanity-check/20-selftests.bats
rsagashe03: Yes when i tested locally this worked. I even tried this and worked 
**local cpuinfo=("$@")**
but since $1 worked i kept that. 
Another thing i tried was remove shift and use $2 that seemed to work too . but i just left that alone

thiagomacieira: It works because it's incomplete (my fault, the existing test failed to do that). The test is actually missing a check that there were no more CPUs.

Here's the fix:
```patch
diff --git i/bats/sanity-check/20-selftests.bats w/bats/sanity-check/20-selftests.bats
index 81f3b718..dcfb4d2b 100644
--- i/bats/sanity-check/20-selftests.bats
+++ w/bats/sanity-check/20-selftests.bats
@@ -29,6 +29,7 @@ test_yaml_absent() {
     printf "Key is not absent:\\n"
     printf "query:      %s\n" "$query"
     printf "value =     %s\n" "${yamldump[$query]}"
+    return 1
 }
 
 test_yaml_numeric() {
@@ -1650,11 +1651,11 @@ selftest_cpuset_unsorted() {
 
     # Get all the processor numbers
     i=0
-    local cpuinfo=$1
-    for expected_logical in "${cpuinfo[@]}"; do
+    for expected_logical; do
         test_yaml_numeric "/cpu-info/$i/logical" "value == $expected_logical"
         i=$((i + 1))
     done
+    test_yaml_absent "/cpu-info/$i/logical"
 }
 
 @test "cpuset=number (inverse order)" {
```

Please apply, reorder the commits for clean-up and I'll approve.

thiagomacieira (approved):

PR #450, opened 2023-11-13
FW/Logging: Refactor format_and_print_message()

thiagomacieira:
We were "abusing" the value of -1, so stop doing that. The TAP and key-value logging formats didn't need the message severity at all, so split off their implementation from the YAML one, which does, except when the message was a skip.

### DIFF
--- a/framework/logging.cpp
+++ b/framework/logging.cpp
@@ -130,6 +130,7 @@ protected:
     bool should_print_fail_info() const;
 
     // shared between the TAP and key-value loggers; YAML overrides
+    static void format_and_print_message(int fd, std::string_view message, bool from_thread_message);
     int print_one_thread_messages(int fd, PerThreadData::Common *data, struct mmap_region r, int level);
 };
 
@@ -179,6 +180,8 @@ private:
     void print_thread_header(int fd, int cpu, int verbosity);
     void maybe_print_slice_resource_usage(int fd, int slice);
     static int print_test_knobs(int fd, mmap_region r);
+    static void format_and_print_message(int fd, std::string_view level, std::string_view message);
+    static void format_and_print_skip_reason(int fd, std::string_view message);
     int print_one_thread_messages(int fd, mmap_region r, int level);
     void print_result_line();
 
@@ -430,6 +433,7 @@ static const char *quality_string(const struct test *test)
 /* which level of "quiet" should this log print at */
 static uint8_t status_level(char letter)
 {
+    // note: the YAML logger requires that the message levels have a 1:1 mapping
     switch (letter) {
     case 'E':
         return LOG_LEVEL_QUIET;         /* always */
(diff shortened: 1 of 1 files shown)


framework/logging.cpp
```diff
@@ -2344,7 +2351,7 @@ inline int YamlLogger::print_one_thread_messages(int fd, mmap_region r, int leve
                 // Only print if the result line didn't already include this
                 std::string skip_message;
                 format_skip_message(skip_message, message);
-                format_and_print_message(fd, 4, std::string_view{&skip_message[0], skip_message.size()}, true);
+                format_and_print_skip_message(fd, skip_message);
```
dilipshivaraju, line 2354: I think when `format_and_print_skip_message` function is called here, it prints the wrong message when there is a skip in the run function.

Current message format would be:
```
  threads:                                                                                                                                                                                              
  - thread: 0                                                                                                                                                                                           
    id: { logical:  0, package: 0, core:  0, thread: 0, family: 6, model: 0x55, stepping: 6, microcode: 0x400320a, ppin: null }                                                                         
    messages:                                                                                                                                                                                           
    - { level: skip, text: 'CPU' }       
```
With this change, the format would be:
```
 threads:                                 
  - thread: 0                              
    id: { logical:  0, package: 0, core:  0, thread: 0, family: 6, model: 0x55, stepping: 6, microcode: 0x400320a, ppin: null }
    messages:                              
  skip-reason: 'CPU' 
```

framework/logging.cpp
```diff
@@ -2308,6 +2282,39 @@ int YamlLogger::print_test_knobs(int fd, mmap_region r)
...
+        /* single line */
+        writevec(fd, indent_spaces(), "    - { level: ", levels[message_level]);
+        print_content_single_line(fd, ", text: '", message, "' }");
+    }
+}
+
+void YamlLogger::format_and_print_skip_message(int fd, std::string_view message)
+{
+    if (message.find('\n') != std::string_view::npos) {
+        /* multi line */
```
dilipshivaraju, line 2309: Shouldn't this require a trimming of newline here?

```
// trim a trailing newline, if any (just one)
if (message[message.size() - 1] == '\n')
      message.remove_suffix(1);
```

framework/logging.cpp
thiagomacieira: So it does. That means this wasn't properly tested in Bats.

framework/logging.cpp
thiagomacieira: Yes and no. The regular message does it because we had old code that had newline, ported over from `printf` logging. We don't have bad `log_skip`.

framework/logging.cpp
thiagomacieira: I take that back, it did fail here. I don't know why I didn't see it.

PR #454, opened 2023-11-17
FW: condition builtin testlist on CPU arch

mcisowsk:


framework/test_selectors/SelectorFactory.cpp
```diff
@@ -188,17 +189,51 @@ struct test *TestrunSelector::testid_to_test(const char *id, bool silent)
     return nullptr;
 }
 
+namespace {
+static constexpr auto arch_spr = "spr";
```
mcisowsk, line 193: from CodingGuidelines:
Constants and defines should be ALL_CAPS.

framework/scripts/generate_test_list.py
```diff
@@ -86,11 +86,13 @@ def main():
             f.write(',\n'.join(test_list))
             f.write('\n};\n\n')
         # write selector function
-        f.write('optional<const span<struct test * const>> get_test_list(const char *test_list_name) {\n')
+        f.write('optional<const span<struct test * const>> get_test_list(const char *test_list_name, bool fallback_to_default) {\n')
         if default_name is not None:
             f.write(f'    if (!test_list_name) return {default_name}_test_list;\n')
```
jposwiata, line 91: if no default list exists, this will cause SEG_FAULTS. Add appropriate if before compares.

framework/test_selectors/SelectorFactory.cpp
```diff
@@ -188,17 +189,51 @@ struct test *TestrunSelector::testid_to_test(const char *id, bool silent)
...
             fprintf(stderr, "# WARNING: --fatal-skips used with full test suite. This will probably fail.\n"
                             "# You may want to specify a controlled list of tests to run.\n");
-        /* generate test list based on quality levels only */
-        for (struct test &test : test_set) {
-            if (test.quality_level >= min_quality)
-                add_test(test_list, &test);
+        auto builtin_test_list = get_test_list(get_test_list_name_from_cpu_info(sApp->shmem->cpu_info[0]), true); // set fallback to default to true
+        if (builtin_test_list) {
+            for (auto* test : *builtin_test_list) {
+                if (test->quality_level >= min_quality)
```
jposwiata, line 229: seems that "sometimes" nullptr might be passed in the list. Add appropriate check to avoid SEGFAULTS.

jposwiata (changes requested):
Seems that the way of "detection" doesn't suit our needs. Would cpu features be better?

framework/test_selectors/SelectorFactory.cpp
thiagomacieira: I'd prefer we use `ALL_CAPS` only for macros. This is not one.

framework/test_selectors/SelectorFactory.cpp
```diff
@@ -188,17 +189,51 @@ struct test *TestrunSelector::testid_to_test(const char *id, bool silent)
...
+    if (info.model == 0x8F) {
+        return arch_spr;
+    } else if (info.model == 0x6C || info.model == 0x6A) {
+        return arch_icx;
+    } else if (info.model == 0x55 && info.stepping >= 7) { // TODO stepping: is this true?
+        return arch_cpx;
+    } else if (info.model == 0x55 && info.stepping < 7) { // TODO
+        return arch_skx;
+    } else {
+        return nullptr;
```
thiagomacieira, line 214: Please use CPU ISA features to detect the CPU, not the model-stepping.

There's code in `dump_cpu_info()`  in sandstone.cpp to try and detect this, but it's currently broken and only detecting Core2.

thiagomacieira (changes requested):

framework/test_selectors/SelectorFactory.cpp
```diff
@@ -188,17 +190,36 @@ struct test *TestrunSelector::testid_to_test(const char *id, bool silent)
...
 }
 
+namespace {
+const char* get_test_list_name_from_cpu_arch()
+{
+    const auto& detected = detect_cpu_arch();
+    if (strcmp(detected.name, unknown_arch.name) == 0) {
+        return nullptr;
+    } else {
+        return detected.codename;
```
jposwiata, line 200: tolower()? (in simd.conf "short names" are uppercase, I think dictionaries are lowercase..)
Could you attach snippet of generated cpu_features.h?

jposwiata (changes requested):

framework/test_selectors/SelectorFactory.cpp
mcisowsk: `static const struct X86Architecture x86_architectures[] = {
    { cpu_core2, "Core2", "x86_64" },
    { cpu_westmere, "Westmere", "wsm" },
    { cpu_sandybridge, "Sandy Bridge", "snb" },
    { cpu_silvermont, "Silvermont", "slm" },
    { cpu_ivybridge, "Ivy Bridge", "ivb" },
    { cpu_goldmont, "Goldmont", "glm" },
    { cpu_haswell, "Haswell", "hsw" },
    { cpu_broadwell, "Broadwell", "bdw" },
    { cpu_tremont, "Tremont", "tnt" },
    { cpu_skylake, "Skylake", "skl" },
    { cpu_skylake_avx512, "Skylake (Avx512)", "skx" },
    { cpu_cascadelake, "Cascade Lake", "clx" },
    { cpu_cooperlake, "Cooper Lake", "cpx" },`

mcisowsk: I used lc() in perl script

jposwiata: I've missed it :D

framework/cpuid_internal.h
```diff
@@ -216,6 +216,17 @@ static uint64_t detect_cpu()
     return features;
 }
 
+__attribute__((unused))
+static const X86Architecture& detect_cpu_arch()
+{
+    for (const auto &arch : x86_architectures) {
```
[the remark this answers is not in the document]
thiagomacieira: That should be the last one that matches. That's why they are sorted. You just need to iterate backwards, actually.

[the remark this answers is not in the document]
thiagomacieira: #457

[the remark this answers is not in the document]
mcisowsk: Ok I will rebase on it once it is merged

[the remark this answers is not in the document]
jposwiata: If the list is in "reverse" order the comment has no sense.

framework/scripts/generate_test_list.py
```diff
@@ -80,17 +80,37 @@ def main():
...
+            f.write(f'static constexpr struct test * const {name}_test_list_[] = {{')
             test_list = [ f'    &_test_{test}' for test in tests ]
             f.write(',\n'.join(test_list))
             f.write('\n};\n\n')
+            f.write(f'static const struct TestList {name}_test_list = {{ {"cpu_" + name if name != "default" else 0}, "{name}", {name}_test_list_ }};\n\n')
+        
+        # write list detector function
+        f.write('optional<const span<struct test * const>> auto_detect() {\n')
+        for name, tests in test_lists.items():
+            f.write(f'    if (cpu_features == {name}_test_list.features) return {name}_test_list.content;\n')
```
mcisowsk, line 100: to check

mcisowsk: to do it like in detect_cpu_arch(), that is with masking `(cpu_features & arch) == arch`, it would require from user to define the lists in `-Dbuiltin_test_list=default:tests/data/test_list_default.csv,icx:tests/data/test_list_icx.csv...`  in a correct order. This is rather insider knowledge, do we want to require user to know that?

mcisowsk: for example, core2 testlist should go as the last one

framework/scripts/generate_test_list.py
```diff
@@ -75,23 +75,59 @@ def main():
...
+        f.write('};\n\n')
+
+        f.write('namespace {\n')
+        f.write('bool isa_features_sort(const TestList* lhs, const TestList* rhs) {\n')
+        f.write('    return lhs->features > rhs->features;\n')
+        f.write('}\n\n')
+
+        # write list detector function
+        f.write('std::optional<const std::span<struct test * const>> auto_detect() {\n')
+        f.write('    std::sort(test_lists.begin(), test_lists.end(), &isa_features_sort);\n')
```
jposwiata, line 110: use lambda here.

mcisowsk: oh, right. Thanks

jposwiata (dismissed):
LGTM

framework/scripts/generate_test_list.py
```diff
@@ -74,24 +74,60 @@ def main():
...
             test_list = [ f'    &_test_{test}' for test in tests ]
             f.write(',\n'.join(test_list))
             f.write('\n};\n\n')
+            f.write(f'static const struct TestList {name}_test_list = {{ {"cpu_" + name if name != "default" else 0}, "{name}", {name}_test_list_ }};\n\n')
+        
+        f.write('namespace {\n')
+
+        # define sorted set of lists
+        f.write('auto isa_features_cmp = [](const TestList* lhs, const TestList* rhs) {\n')
+        f.write('    return lhs->features > rhs->features;\n')
```
thiagomacieira, line 101: This sorting may produce false positives because the CPU features aren't added always at the end in the CPUID and our bits are therefore also all out of order.  Take for example `cpu_feature_avxvnni`, which is higher than `cpu_feature_avx512f`. That means a list for Sierra Forest would have this bit set and its numeric value would be higher than a list for Ice Lake. Then, when running on a CPU that has both features (e.g., Sapphire Rapids), we may select the SRF one instead of the ICX one.

One way around this is to sort by bit population count instead of numeric value (use [`std::popcount`](https://en.cppreference.com/w/cpp/numeric/popcount)). It may still fail, but is less likely.

Another is to not sort at all (except maybe the "default" fallback last) but instead require the maintainers to pass the lists in the proper search order.

framework/cpuid_internal.h
```diff
@@ -237,11 +248,16 @@ static void check_missing_features(uint64_t features, uint64_t minimum_cpu_featu
...
+static uint64_t detect_cpu_features()
 {
     return 0;
 }
 
+static const X86Architecture& detect_cpu_arch() 
+{
+    return unknown_arch;
+}
+
```
thiagomacieira, line 260: You didn't use any of these modifications or the ones in main.cpp

framework/sandstone.cpp
```diff
@@ -1280,24 +1281,13 @@ static std::string cpu_features_to_string(uint64_t f)
...
+    const auto& detected = detect_cpu_arch();
 
-    // find the best matching CPU
-    const char *detected = "<unknown>";
-    for (const auto &arch : x86_architectures) {
-        if ((arch.features & cpu_features) == arch.features) {
-            detected = arch.name;
-            break;
-        }
-        if (sApp->shmem->verbosity > 1)
```
thiagomacieira, line 1292: You've accidentally removed this in rebasing (from #457)

framework/test_selectors/SelectorFactory.cpp
```diff
@@ -195,10 +197,17 @@ void generate_test_list(std::vector<struct test *> &test_list, std::span<struct
...
+        if (builtin_test_list) {
+            for (auto* test : *builtin_test_list) {
+                if (test->quality_level >= min_quality)
+                    add_test(test_list, test);
+            }
+        } else {
+            for (auto& test : test_set) {
+                if (test.quality_level >= min_quality)
+                    add_test(test_list, &test);
+            }
```
thiagomacieira, line 210: I don't think this is the right place to put this. I would prefer for it to be in sandstone.cpp's `main()`, where we're already calling `get_test_lsit()` (actually, there may be no need for modifications, at all, if you make the function handle a `nullptr` name be the same as "auto").

thiagomacieira (changes requested):

framework/scripts/generate_test_list.py
mcisowsk: no sorting then

framework/test_selectors/SelectorFactory.cpp
```diff
@@ -206,6 +207,26 @@ void generate_test_list(std::vector<struct test *> &test_list, std::span<struct
     }
 }
 
+std::optional<const std::span<struct test * const>> auto_detect_test_list(const std::vector<const TestList*>& test_lists,
+                                                                          const TestList* default_test_list)
+{
+#ifdef SANDSTONE
```
mcisowsk, line 213: If I do not use this flag correctly, then it will be better to create separate file for this specific function auto_detect_test_list(), so that it won't be included in the unittests source files

framework/scripts/generate_test_list.py
```diff
@@ -75,23 +75,40 @@ def main():
...
+
         for name, tests in test_lists.items():
-            f.write(f'static constexpr struct test * const {name}_test_list[] = {{')
+            f.write(f'static constexpr struct test * const {name}_test_list_[] = {{')
             test_list = [ f'    &_test_{test}' for test in tests ]
             f.write(',\n'.join(test_list))
             f.write('\n};\n\n')
+            f.write(f'static const struct TestList {name}_test_list = {{ {"cpu_" + name if name != "default" else 0}, "{name}", {name}_test_list_ }};\n\n')
+
+        f.write('static const std::vector<const TestList*> test_lists = {\n')
```
thiagomacieira, line 92: Same problem as before: this is a static non-POD, so it runs too early. How about a plain C array?

This is unfortunately not a theoretical or academic exercise. Looking at the generated code, I see:
```
Disassembly of section .text.startup:

0000000000000000 <_GLOBAL__sub_I_sandstone_test_lists.cpp>:
   0:   55                      push   %rbp
   1:   48 8d 05 00 00 00 00    lea    0x0(%rip),%rax        # 8 <_GLOBAL__sub_I_sandstone_test_lists.cpp+0x8>
                        4: R_X86_64_PC32        .data.rel.ro.local-0x4
   8:   48 8d 0d 00 00 00 00    lea    0x0(%rip),%rcx        # f <_GLOBAL__sub_I_sandstone_test_lists.cpp+0xf>
                        b: R_X86_64_PC32        .data.rel.ro.local+0x5c
   f:   c4 e1 f9 6e c1          vmovq  %rcx,%xmm0
```
There's an AVX instruction there.

framework/test_selectors/SelectorFactory.cpp
```diff
@@ -206,6 +207,26 @@ void generate_test_list(std::vector<struct test *> &test_list, std::span<struct
     }
 }
 
+std::optional<const std::span<struct test * const>> auto_detect_test_list(const std::vector<const TestList*>& test_lists,
```
thiagomacieira, line 210: Pass a `std::span` here instead of `std::vector`

framework/test_selectors/SelectorFactory.cpp
thiagomacieira: That's ok, the same happens above (line 83)

I actually prefer to have it in this file, instead of generated,  because we do need to apply the sorting. The current generated code for the internal product has:
```c++
    &icx_test_list,
    &cpx_test_list,
    &skx_test_list,
    &spr_test_list,
```

This is not in any special sorting order (ICX later than CPX and SKX, but earlier than SPR). So we must either sort in this function (at a minimum to `assert` that the supplied order is correct).

thiagomacieira (changes requested):
Nearly there.

framework/test_selectors/SelectorFactory.cpp
mcisowsk: But I thought you agreed that there can be no sorting at all, as no method ensures 100% correct sorting. User should define the architecture lists in the correct order.

framework/scripts/generate_test_list.py
```diff
@@ -72,26 +72,47 @@ def main():
...
+            f.write(f'static const struct TestList {name}_test_list = {{ {"cpu_" + name if name != "default" else 0}, "{name}", {name}_test_list_ }};\n\n')
+
+        if has_any_lists:
+            f.write('static const struct TestList* test_lists[] = {\n')
+            for name, _ in test_lists.items():
+                if default_name is not None and name != default_name:
+                    f.write(f'    &{name}_test_list,\n')
+            if default_name is not None:
+                f.write(f'    &{default_name}_test_list\n') # put default list as last one
+            f.write('};\n\n')
```
thiagomacieira, line 101: This is creating an array of pointers:
```c++
static const struct TestList* test_lists[] = {
    &icx_test_list,
    &cpx_test_list,
    &skx_test_list,
    &spr_test_list,
    &default_test_list
};
```

So there's one indirection too many: we could have a simple array of `TestList` items.

framework/test_selectors/SelectorFactory.cpp
thiagomacieira: We can go with that. If all lists are supersets of previous ones, we could search for the one that matches most bits... but the ISA extensions weren't a monotonic sequence. So I agree on enforcing the order on whoever supplies the lists.

thiagomacieira (approved):

framework/scripts/generate_test_list.py
jposwiata: This list is in incorrect order (the order from the option had been agreed, selector factory will use default list in most cases).

jposwiata (changes requested):
The per-gen list is in incorrect order (sorted with increased number of features).

thiagomacieira:
Superseded by #486 .

PR #455, opened 2023-11-18
Remove ResourceIssueSkipCategory and replace its usages.

busykai:
We have decided to split ResourceIssueSkipCategory into, roughly, two: TestResourceIssueSkipCategory and OSResourceIssueSkipCategory. This foments  correct usages of the categories by the test writers as well as helps analyzing the logs. The usages of ResourceIssueSkipCategory were removed as well as the category itself. Since it existed ever so briefly, there's no need to make it a non-breaking change.

### DIFF
--- a/framework/selftest.cpp
+++ b/framework/selftest.cpp
@@ -169,7 +169,7 @@ static int selftest_logs_getcpu_run(struct test *test, int cpu)
     log_skip(OsNotSupportedSkipCategory, "No API to get the CPU number on this OS");
 #endif
     if (cpu_number == -1)
-        log_skip(ResourceIssueSkipCategory, "OS failed: %m");
+        log_skip(OSResourceIssueSkipCategory, "OS failed: %m");
     log_info("%d", cpu_number);
     return EXIT_SUCCESS;
 }
--- a/tests/ifs/ifs.c
+++ b/tests/ifs/ifs.c
@@ -134,7 +134,7 @@ static int scan_common_init(struct test *test)
         int enforce_run = get_testspecific_knob_value_int(test, "enforce_run", -1);
         if (memcmp(status_buf, "fail", strlen("fail")) == 0 && enforce_run != 1 )
         {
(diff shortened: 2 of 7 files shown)


framework/selftest.cpp
```diff
@@ -169,7 +169,7 @@ static int selftest_logs_getcpu_run(struct test *test, int cpu)
     log_skip(OsNotSupportedSkipCategory, "No API to get the CPU number on this OS");
 #endif
```
dilipshivaraju, line 170: Doesn't this require `return EXIT_SKIP` here?

framework/selftest.cpp
```diff
@@ -169,7 +169,7 @@ static int selftest_logs_getcpu_run(struct test *test, int cpu)
...
     if (cpu_number == -1)
-        log_skip(ResourceIssueSkipCategory, "OS failed: %m");
+        log_skip(OSResourceIssueSkipCategory, "OS failed: %m");
```
dilipshivaraju, line 172: I think here too inside the if block. `return EXIT_SKIP`

tests/ifs/ifs.c
```diff
@@ -134,7 +134,7 @@ static int scan_common_init(struct test *test)
         int enforce_run = get_testspecific_knob_value_int(test, "enforce_run", -1);
         if (memcmp(status_buf, "fail", strlen("fail")) == 0 && enforce_run != 1 )
         {
-            log_skip(ResourceIssueSkipCategory, "Previous run failure found! This test will skip until enforced adding flag: "
+            log_skip(TestResourceIssueSkipCategory, "Previous run failure found! This test will skip until enforced adding flag: "
```
dilipshivaraju, line 137: I think @AthenasJimenez wants us to change the message to

`The previous execution encountered an error, and this test will not proceed unless it is explicitly enforced with knob enforce_run=1.`

busykai:
@dilipshivaraju I think we need to update systematically the opendcdiag tests. Your remarks are valid, but are not really in context of this particular change.

dilipshivaraju (approved):

PR #456, opened 2023-11-20
Change triage default behavior

busykai:
Do not run triage by default. Add an option `--triage` to force the previous behavior.

[ChangeLog][Framework] Change the default behavior to not run triage. Add --triage option to force the previous behavior.


thiagomacieira:
Update the bats to pass `--triage` in the triaging test.

thiagomacieira (approved):

PR #460, opened 2023-11-23
FW/Win32/child_debug: don't stop execution on non-system exceptions

thiagomacieira:
This can happen when exceptions are thrown for other reasons, in particular that of C++ exceptions. Both GCC and MSVC use the SEH mechanism to handle their exceptions. We didn't catch this before because the most self-tests are run with --on-crash=core, so the vectored exception handler wasn't installed.

Before:
```yaml
- test: selftest_cxxthrowcatch
  result: crash
  result-details: { crashed: true, core-dump: false, code: 0x20474343, reason: '???' }
  fail: { cpu-mask: 'X', time-to-fail: null, seed: 'AES:1d1726ac3d4d55dd78077b0a6de46d10e2e8d953c2b2aa2287f884f5921b92ef'}
  time-at-end:   { elapsed:     21.343, now: !!timestamp '2023-11-23T23:55:02Z' }
  test-runtime: 19.280
  threads:
  - thread: 0
    id: { logical-group:  0, logical:  0, package: 0, core: 0, thread: 0 }
    messages:
    - { level: error, text: 'E> Received exception 0x20474343 (???), RIP = 0x7ffd7324565c, parameters [1cd428c88a0]' }
```

0x20474343 is "GCC", see [1] and [2].

Now:
```yaml
tests:
- test: selftest_cxxthrowcatch
  result: pass
  time-at-end:   { elapsed:     22.560, now: !!timestamp '2023-11-23T23:55:48Z' }
  test-runtime: 20.128
```

[1] https://gcc.gnu.org/git/?p=gcc.git;a=blob;f=libgcc/unwind-seh.c;hb=HEAD
[2] https://github.com/llvm/llvm-project/blob/main/libunwind/src/Unwind-seh.cpp

### DIFF
--- a/framework/selftest.cpp
+++ b/framework/selftest.cpp
@@ -56,6 +56,8 @@ using uint128_t = __uint128_t;
 
 struct SelftestException : std::exception
 {
+    int cpu;
+    SelftestException(int cpu = -1) : cpu(cpu) {}
     const char *what() const noexcept override
     {
         return "OpenDCDiag C++ selftest exception";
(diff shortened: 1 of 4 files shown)


framework/selftest.cpp
```diff
@@ -939,6 +951,13 @@ static struct test selftests_array[] = {
     .test_run = selftest_logs_random_run,
     .desired_duration = -1,
 },
+{
+    .id = "selftest_cxxthrowcatch",
+.description = "Throws and catches a C++ exception",
```
AthenasJimenez, line 956: missed indentation?

AthenasJimenez (approved):

PR #461, opened 2023-12-04
FW: add cleanup_global() to centralize application exiting

thiagomacieira:
We exited in three different places but only one duplicated the call to `logging_close_global()`. Only the main exit remembered to call `print_application_footer()` and restore any system state we modified (in the current version, the `--vary-frequency` state).

This centralizes everything in a function that Does The Right Thing™, plus removes a magic constant in the code.

Our exit codes are:
* EXIT_SUCCESS (0) - all tests passed (or skipped)
* EXIT_FAILURE (1) - at least one test failed
* EXIT_INVALID (2) - inconclusive result
* EXIT_ABORTED (3) - (documentation only) used by abort() on Windows
* other values (64-78, >=200) - unable to start, logging hasn't started

The shell will report crash (signal) exits by OR'ing the signal number with 128.

### DIFF
--- a/framework/logging.cpp
+++ b/framework/logging.cpp
@@ -675,15 +675,28 @@ int logging_close_global(int exitcode)
 {
     progress_bar_flush();
     if (!SandstoneConfig::NoLogging) {
-        if (exitcode != EXIT_SUCCESS) {
+        if (exitcode != EXIT_SUCCESS)
             logging_print_log_file_name();
-            logging_printf(LOG_LEVEL_QUIET,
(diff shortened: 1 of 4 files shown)


AthenasJimenez (approved):
LGTM, just a quick question, why logging_flush() is no longer needed in this version?

thiagomacieira:
> LGTM, just a quick question, why logging_flush() is no longer needed in this version?

I didn't feel it was justified because there's no new test starting.

PR #466, opened 2023-12-15
FW: add IGNORE_RETVAL() for mprotect()

thiagomacieira:
Either we can actually ignore because they don't make a difference or there's no recourse in case it fails, so we may as well let the started threads crash.


lukasz-tuz (approved):
LGTM

PR #468, opened 2024-01-12
FW/Win32: dynamically load pdh.dll only if --service is enabled

thiagomacieira:
This library is not present on stripped-down builds of Windows, such as Windows PE. So don't link OpenDCDiag to it, but instead load it after we've been told to use --service.

Introduced by 2ac0d930ae7f3472794d4e05fbfdfd014c928e44.


AthenasJimenez (approved):
LGTM. And just to feed my curiosity, why using different error codes (from 0 to 7)?

thiagomacieira:
> LGTM. And just to feed my curiosity, why using different error codes (from 0 to 7)?

It seemed to be using something sequential to indicate what step the problem was happening at.

PR #470, opened 2024-01-23
FW: Remove the old "SANDSTONE_GA" macros

thiagomacieira:
They're unused.


dawid-sabat:
If you wonder, why this PR failed, the commit is not signed-off. 

What is more please check out `meson_options.txt`  We also mention `ga` and `ga-dev`

```
option('flavor', type : 'combo', choices : ['default', 'ga', 'ga-dev'], value : 'default',
    description : 'Select one of the novel build flavors (ga, ga-dev) or the default')
```

mrkz (approved):

PR #475, opened 2024-01-31
Implement Rule of Three in ManagedCodeBuffer

bartlomiejgrzeskowiak:
Added a constructor, copy constructor and copy assignment operator to ManagedCodeBuffer to adhere to the Rule of Three. This ensures proper copying and management of dynamic memory in the class.

### DIFF
--- a/framework/sandstone_test_utils.h
+++ b/framework/sandstone_test_utils.h
@@ -55,6 +55,13 @@ public:
         free_code_buffer();
     }
 
+    // Default constructor
+    ManagedCodeBuffer() {}
+
+    // Do not allow to copy
+    ManagedCodeBuffer(const ManagedCodeBuffer& other) = delete;
+    ManagedCodeBuffer& operator=(const ManagedCodeBuffer& other) = delete;
+
     bool allocate(size_t size) {
         free_code_buffer();
         buffer_ptr = static_cast<uint8_t *>(


framework/sandstone_test_utils.h
```diff
@@ -55,6 +55,31 @@ class ManagedCodeBuffer {
...
+    // Copy constructor
+    ManagedCodeBuffer(const ManagedCodeBuffer& other) : buffer_size(other.buffer_size) {
+        buffer_ptr = static_cast<uint8_t*>(
+            mmap(nullptr, ROUND_UP_TO_PAGE(buffer_size), (PROT_READ | PROT_WRITE), (MAP_ANONYMOUS | MAP_PRIVATE), -1, 0)
+            );
+        std::copy(other.buffer_ptr, other.buffer_ptr + buffer_size, buffer_ptr);
+    }
+
+    // Copy assignment operator
+    ManagedCodeBuffer& operator=(const ManagedCodeBuffer& other) {
```
thiagomacieira, line 70: Maybe the correct thing to do here would be to `= delete` these, so no one attempts to copy them.

bartlomiejgrzeskowiak: Changed to =delete.
Yes. Coverity still complains, but now we can triage this issue.

thiagomacieira (approved):

PR #480, opened 2024-02-12
FW/sandstone_ssl_rand: mark ctx as unused on instantiate/uninstantiate

AthenasJimenez:
This will address compiler warnings.


framework/sandstone_ssl_rand.cpp
```diff
@@ -27,15 +27,15 @@ static void sandstone_rand_freectx(void *ctx)
 }
 
 /* Random number generator functions */
-static int sandstone_rand_instantiate(void *ctx, unsigned int strength, int prediction_resistance,
+static int sandstone_rand_instantiate(ossl_unused void *ctx, unsigned int strength, int prediction_resistance,
                                const unsigned char *pstr, size_t pstr_len,
                                const OSSL_PARAM params[])
 {
     ctx = reinterpret_cast<void *>(EVP_RAND_STATE_READY);
```
thiagomacieira, line 34: The problem is not that they are unused; we don't have the compiler warning for unused parameters and variables enabled (we should, but don't).

The issue is a set-but-unused variable here. This statement is useless and gets removed from the execution.

thiagomacieira (changes requested):

framework/sandstone_ssl_rand.cpp
AthenasJimenez: I removed the ctx set and also leaved the unused as well.

thiagomacieira: You didn't push the update (did you forget the `-a` in `git commit --amend -a`?).

The `ossl_unused` is inconsistent because we're not using any of the other parameters. Either add to them all, or to none.

AthenasJimenez: Yeah, I totally forgot to amend my changes lol.

Regarding to the `ossl_unused`, I'm also fan of consistency so removing all appearances.

thiagomacieira (approved):

PR #482, opened 2024-02-15
FW/child_debug: use std::string

thiagomacieira:
This changes both the Unix and Windows child-debuggers to use `std::string` more instead of writing to file descriptors. This should improve performance a bit because we don't make system calls to store our content as it's being built. It also allows us to remove the `LoggingStream` class (was added in #286's 040bc365e8eeddf27b4a4a892dfb2a7a2a5cb912 to replace the `FILE *`-based API).

Using the API added in #479, this removes `open_memstream()` on Unix systems and more importantly, removes the need for an actual temporary file on Windows.

### DIFF
--- a/framework/logging.cpp
+++ b/framework/logging.cpp
@@ -1139,15 +1139,6 @@ void logging_finish()
 {
 }
 
-LoggingStream logging_user_messages_stream(int thread_num, int level)
-{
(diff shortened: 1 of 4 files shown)


lukasz-tuz (approved):

thiagomacieira:
Rebasing again...

busykai (approved):
Looks good to me. I'll try and spend some time testing it as well...

PR #484, opened 2024-02-15
Add Uncore frequency manager support.

dilipshivaraju:
This PR consists of 2 commits. 

First commit will re-factor already existing core frequency manager support so that it makes easier addition of uncore frequency manager support. This does not change any functionality of core frequency manager.

Second commit will add the actual uncore frequency manager support. This commit will add `--vary-uncore-frequency` option to opendcdiag. If this option is used, every test will run at pre-calculated different uncore frequency per fracture of the test within it's allocated time to run. Like core frequency manager, this option works only on linux and the user should be root.

Testing:

All tests are conducted on 2 socket BDX platform(18 core/socket). opendcdiag option used to measure uncore frequency

`opendcdiag -vv -e eigen_gemm_double14 -e zlib -t30s`

1st Test: `--vary-uncore-frequency` is "NOT" used.

image

Uncore frequency for both sockets remains constant. The dips in the graph is the transition from one fracture to another.

2nd Test: `--vary-uncore-frequency` is used.

image

Uncore frequency for both sockets varies from 2.8Ghz to 1Ghz between fractures.

### DIFF
--- a/framework/frequency_manager.hpp
+++ b/framework/frequency_manager.hpp
@@ -6,7 +6,9 @@
 #ifndef FREQUENCY_MANAGER_HPP
 #define FREQUENCY_MANAGER_HPP
 
-#define BASE_FREQ_PATH         "/sys/devices/system/cpu/cpu"
+#include <unordered_set>
+#define BASE_CORE_FREQ_PATH    "/sys/devices/system/cpu/cpu"
+#define BASE_UNCORE_FREQ_PATH  "/sys/devices/system/cpu/intel_uncore_frequency/package_0"
 #define SCALING_GOVERNOR       "/cpufreq/scaling_governor"
 #define SCALING_SETSPEED       "/cpufreq/scaling_setspeed"
 
@@ -15,6 +17,7 @@ class FrequencyManager
 private:
 
 #ifdef __linux__
+    // core-frequency variables
     int max_core_frequency_supported = 0;
     int min_core_frequency_supported = 0;
     std::vector<std::string> per_cpu_initial_scaling_governor;
@@ -22,7 +25,14 @@ private:
     int current_set_frequency = 0;
     std::vector<int> core_frequency_levels;
     int core_frequency_level_idx = 0;
-    static constexpr int total_frequency_levels = 9;
+    int total_core_frequency_levels = 0;
+
+    // uncore-frequency variables
+    std::vector<std::pair<int, int>> initial_uncore_frequency;  // initial (min, max) un-core pair for each socket
+    std::vector<std::vector<int>> uncore_frequency_levels;  // frequency levels for each socket
+    uint16_t total_sockets = 0;
+    int uncore_frequency_level_idx = 0;
+    int total_uncore_frequency_levels = 0;
 
     std::string read_file(const std::string &file_path)
     {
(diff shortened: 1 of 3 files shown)


framework/frequency_manager.hpp
```diff
@@ -15,15 +16,22 @@ class FrequencyManager
...
+    int max_core_frequency_supported = 0;
+    int min_core_frequency_supported = 0;
     std::vector<std::string> per_cpu_initial_scaling_governor;
     std::vector<std::string> per_cpu_initial_scaling_setspeed;
     int current_set_frequency = 0;
-    std::vector<int> frequency_levels;
-    int frequency_level_idx = 0;
+    std::vector<int> core_frequency_levels;
+    int core_frequency_level_idx = 0;
     static constexpr int total_frequency_levels = 9;
```
lukasz-tuz, line 27: If core and uncore are to share total number of frequency levels, I don't think we need to maintain separate frequency level indices for them. If there's the same number of levels for core and uncore, and frequencies changes in lockstep with every fracture, then `uncore_frequency_level_idx` and `core_frequency_level_idx` would always have the same value.

However, I think `total_frequency_levels` should also be split between core and uncore to give us better control over uncore frequencies. Uncore typically has smaller dynamic range of frequency control, so it probably won't need 9 different levels.

lukasz-tuz (changes requested):
A couple of small changes, otherwise lgtm.

framework/frequency_manager.hpp
```diff
@@ -160,7 +226,25 @@ class FrequencyManager
 #endif
     }
 
-    void restore_initial_state()
+    void change_uncore_frequency()
+    {
+#ifdef __linux__
+        for (size_t socket = 0; socket < total_sockets; socket++) {
+            std::pair<int, int> max_min_frequency;
+            std::string uncore_frequency_path = "/sys/devices/system/cpu/intel_uncore_frequency/package_0";
```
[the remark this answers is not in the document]
dilipshivaraju: Sure will do this.

framework/frequency_manager.hpp
dilipshivaraju: You are right on the first part. But if we decide to implement second suggestion i.e., having different `total_frequency_levels` for core and uncore, wouldn't this require separate `frequency_level_idx` since the frequency_idx has to wrap over once all the levels have been covered?

lukasz-tuz: Yes, it's either-or, can't have both. If you choose to keep the same number of frequency levels, then code can be simplified. I would, however, opt for having different `total_frequency_levels` for core and uncore.

dilipshivaraju: Can you suggest a number that's ideal here? I am not an uncore expert. I will try to make it work on any number.

lukasz-tuz: I'd suggest `min(16, max_core_frequency_supported - min_core_frequency_supported)` for cores, `min(8, max_uncore_frequency_supported - min_uncore_frequency_supported)` (adjusted for units in which the max/min values are represented).

dilipshivaraju: suppose the `max_core_frequency_supported` = 3.6Ghz and `min_core_frequency_supported` is 1.2Ghz, min(16, 3.6-1.2) = min(16, 2.4) = 2.4 ? 

Similarly `max_uncore_frequency_supported` = 2.8Ghz and `min_uncore_frequency_supported` = 1.2Ghz, min(8, 2.8-1.2) = min(8, 1.8) = 1.6? 

Is my math right here? What am i missing?

lukasz-tuz: Sorry, my bad for using frequency multipliers. If you assume base clock is 100MHz (which it is), then min/max frequencies can be expressed as multiplies of 100MHz. For values you provided, this would translate to:
* `min(16, 36 - 12) = min(16, 24) = 16`
* `min(8, 28 - 12) = min(8, 18) = 8`
resulting in 16 frequency levels for cores and 8 for uncore. With number of levels we can calculate actual frequency as `f_min + N * f_step`, where N is in range(0, 16), or (0, 8) for uncore, and `f_step = (f_max - f_min) / N * 100MHz`.

dilipshivaraju: Done. Can you check and approve? I'll squash all the commits to first 2 commits and merge.

lukasz-tuz (approved):
Looks good.

PR #485, opened 2024-02-22
Do not print help by default, instead suggest to use it.

busykai:
The help information can get quite lengthy and does cause the actual error message to scroll up. Instead print short suggestion to use '--help' option when there's an error processing.

(Finding random stuff in my branches that is useful.)

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -1316,6 +1316,10 @@ static void dump_cpu_info()
     }
 }
 
+static void suggest_help(char **argv) {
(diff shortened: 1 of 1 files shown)


framework/sandstone.cpp
```diff
@@ -1316,6 +1316,10 @@ static void dump_cpu_info()
     }
 }
 
+static void suggest_help(char **argv) {
```
thiagomacieira, line 1319: Nitpick: position of the `{` (should be on the next line.

thiagomacieira (approved):

PR #486, opened 2024-02-23
FW: condition builtin testlist on CPU arch

jposwiata:
The generator updated according the discussion from PR #454. The commit includes two behavioural changes:
- default list is always "at the end" (when there's no matching list, either by name or features), location of this entry in the list is not important
- it is not required to name 'default' list (only a filename might be put into the option, without 'default' prefix)
- internal data model slightly changed: each file (by name) is parsed once, might be shared between lists for multiple generations.

If no matching list is defined during compilation (and selected with --use-builtin-test-list), appropriate warning is emitted in runtime.

[ChangeLog] Default list must be explicitly defined in the option. If no matching list is found in the list (by either name or features, checked in the order in the list, 'default' is skipped), the 'default'  is assumed, and appropriate warning is emitted.

### DIFF
--- a/framework/scripts/generate_test_list.py
+++ b/framework/scripts/generate_test_list.py
@@ -29,41 +29,52 @@ def main():
     h_file_basename = os.path.basename(h_file)
     cpp_file = args[2]
     root = args[3]
-    config_str = args[4] # the rest is config tuples
-    # test_list_files = { "name1" => "test_list_file1", "name2" =>
-    # "test_list_file2", ... }
+    config_str = args[4]
     test_list_files = {}
+    default_name = None
+    other_names = False
+    test_lists = {}
+    all_tests = {}
+    files = {}
+
     if (config_str == ''):
         config = {}
     else:
         config = config_str.split(',')
-    default_name = None
     for pair in config:
         parsed = pair.split(':')
-        if len(parsed) != 2:
+        if len(parsed) == 2:
+            name = parsed[0]
+            path = parsed[1]
+        elif len(parsed) == 1:
+            name = "default"
+            path = parsed[0]
+        else:
             print(f'Incorrect configuration string: {pair}')
             exit(2)
-        name = parsed[0]
-        path = parsed[1]
-        if default_name is None:
-            default_name = name # the first on the list is the default one,
-                                # it should be called "default", but this does
-                                # not rely on that
-        test_list_files[name] = root + "/" + path
 
-    # test_lists = { "name1" => [ "test1", "test2", ... ], "name2" => [ "test2",
-    # "test4", ..] }
-    # contains all the .c files to generate mapped to the ordered list of tests
-    test_lists = {}
-    all_tests = {}
+        if name == 'auto':
+            print(f'Reserved generation name used for {path}')
+            exit(2)
+        if name == 'default':
+            default_name = name
+        else:
+            other_names = True
 
-    # read and process all the arguments
-    for name, file in test_list_files.items():
-        with open(file, 'r') as f:
+        test_list_files[name] = path
+
+    # collect file names
+    for name, path in test_list_files.items():
+        if path not in files:
+            files[path] = len(files)
+
+    # read and process all files
+    for path, file_id in files.items():
+        with open(root + "/" + path, 'r') as f:
             lines = f.readlines()
 
-        test_lists[name] = []
-        test_list = test_lists[name]
+        test_lists[path] = []
+        test_list = test_lists[path]
         for l in lines:
             l = l.split('#', 1)[0]
             l = l.strip()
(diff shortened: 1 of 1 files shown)


jposwiata:
If the options is `data/default.txt,spr:data/spr.txt,skx:data/skx.txt,icx:data/icx.txt`, the resulting code is:
`// **This is generated file, do not edit**

#include "sandstone_test_lists.h"

extern struct test _test_all_gpr_using;
extern struct test _test_eigen_gemm_identity_avx512;
extern struct test _test_eigen_sparse;
extern struct test _test_eigen_svd;
extern struct test _test_eigen_svd_cdouble;

struct TestList {
    uint64_t features;
    const char* name;
    const std::span<struct test* const> content;
};

static constexpr struct test * const default_test_list_[] = {
    &_test_eigen_sparse,
    &_test_all_gpr_using
};
static const struct TestList default_test_list = { 0, "default", default_test_list_ };

static constexpr struct test * const spr_test_list_[] = {
    &_test_eigen_gemm_identity_avx512,
    &_test_all_gpr_using
};
static const struct TestList spr_test_list = { cpu_spr, "spr", spr_test_list_ };

static constexpr struct test * const skx_test_list_[] = {
    &_test_eigen_svd
};
static const struct TestList skx_test_list = { cpu_skx, "skx", skx_test_list_ };

static constexpr struct test * const icx_test_list_[] = {
    &_test_eigen_svd_cdouble,
    &_test_all_gpr_using
};
static const struct TestList icx_test_list = { cpu_icx, "icx", icx_test_list_ };

static bool test_list_matches(const TestList* test_list, const char* name)
{
#ifdef SANDSTONE
    if ((name == nullptr) || (strcmp(name, "auto") == 0)) {
        return (test_list->features & cpu_features) == test_list->features;
    }
#endif
    return (name != nullptr) && (strcmp(name, test_list->name) == 0);
}

std::optional<const std::span<struct test * const>> get_test_list(const char *name)
{
    if (test_list_matches(&spr_test_list, name)) return spr_test_list.content;
    if (test_list_matches(&skx_test_list, name)) return skx_test_list.content;
    if (test_list_matches(&icx_test_list, name)) return icx_test_list.content;
    if (name && (strcmp(name, "default") != 0)) {
        logging_printf(LOG_LEVEL_QUIET, "# ERROR: list for %s not found. Using default\n", name);
    }
    return default_test_list.content;
}
`

framework/scripts/generate_test_list.py
```diff
@@ -74,31 +79,63 @@ def main():
...
         for name in test_lists.keys():
-            f.write(f'    if  (!strcmp(test_list_name, "{name}")) return {name}_test_list;\n')
-        f.write('    return nullopt;\n')
+            if default_name is None or name != default_name:
+                f.write(f'    if (test_list_matches(&{name}_test_list, name)) return {name}_test_list.content;\n')
+        if default_name is not None:
+            f.write('    if (name && (strcmp(name, "default") != 0)) {\n')
+            f.write('        logging_printf(LOG_LEVEL_QUIET, "# ERROR: list for %s not found. Using default\\n", name);\n')
+            f.write('    }\n')
+            f.write(f'    return {default_name}_test_list.content;\n')
```
thiagomacieira, line 128: I may need in the future to have this return the `TestList` object so we can include the name of the list in the log output.

thiagomacieira (approved):

framework/scripts/generate_test_list.py
jposwiata: Appropriate function added (with a bit confusing name, but I'd prefer keep modification in the generator only). "C" struct definition moved to generated header (I wonder if it should be hpp with using, as it won't compile with "C").
The structure contains source of the list to improve the experience.

framework/scripts/generate_test_list.py
```diff
@@ -74,32 +79,79 @@ def main():
...
-        f.write('#include <span>\n')
-        f.write('#include "sandstone.h"\n\n')
-        f.write('std::optional<const std::span<struct test * const>> get_test_list(const char *);\n')
+        f.write('#include <span>\n\n')
+
+        # per-gen list definition
+        f.write('typedef struct TestList {\n')
+        f.write('    uint64_t features;\n')
+        f.write('    const char* name;\n')
+        f.write('    const char* filename;\n')
```
thiagomacieira, line 144: What's the point of the file name? It won't exist in the user's system.

jposwiata: I'd assume, that it might be used to identify exact list being used in execution (I don't expect "trivial" mapping where e.g. clx => data/clx.txt, rather expect data/clx_ppv_240226.txt). I don't know details of your "requirement" on metadata assigned to each list.
In the code it is used to determine, whether the list is NULL (the list is not created on any file) or empty. (o/c it might be "fixed" with optional in the structure).

jposwiata: BTW, should we consider optimization, where the test pointers are kept per file (instead of current "per name")?

jposwiata: Replaced "test list per generation" with "shared test lists" to optimize memory usage with duplicated files. See snippets.

thiagomacieira: Ah, "which test list" I had meant "clx", not the file name.

thiagomacieira (approved):

framework/scripts/generate_test_list.py
```diff
@@ -74,32 +85,83 @@ def main():
...
+        f.write("// **This is generated file, do not edit**\n\n")
+        f.write('#include "sandstone.h"\n')
         f.write('#include <optional>\n')
-        f.write('#include <span>\n')
-        f.write('#include "sandstone.h"\n\n')
-        f.write('std::optional<const std::span<struct test * const>> get_test_list(const char *);\n')
+        f.write('#include <span>\n\n')
+
+        # per-gen list definition
+        f.write('typedef struct TestList {\n')
```
busykai, line 153: This does not need to be here. It should go together with the declaration of `select_test_list()` and implementation of `get_test_list()` to `.cpp`. (`get_test_list()` does not need to be inline, as it does not optimize anything).

jposwiata: I've considered moving the header from generated code to the source (what seems to be obvious step, and currently one compile time dependency was removed). Decided to keep both files always generated to reduce number of "visible" changes.
"static inline" is to avoid putting this function in the library interface (regardless optimization level). If it is not used, the body is not even compiled.
select_test_list() has been requested by Thiago ("for future use", ask him for the details).

busykai: I am saying this should go to the generated .cpp, not to the fixed code. It should not be exposed at all.

PR #488, opened 2024-02-28
FW/child_debug: use std::string more

thiagomacieira:
For the dumping the processor context, this removes `open_memstream()` on Unix systems and more importantly, removes the need for an actual temporary file on Windows.

For the gdb communication, we were using a plain file descriptor because that allowed us to use `splice()` straight from the pipe to the backing file, in case the backtrace was big. But that's really unnecessary and was a source of bugs because the non-Linux code was never tested. It's now replaced with direct `std::string` and moved up the location where we call the logging functions to outside of the GDB communication loop.

And doing the same for the communication with `ps`, which allows us to remove the `LoggingStream` class.

### DIFF
--- a/framework/logging.cpp
+++ b/framework/logging.cpp
@@ -1139,15 +1139,6 @@ void logging_finish()
 {
 }
 
-LoggingStream logging_user_messages_stream(int thread_num, int level)
-{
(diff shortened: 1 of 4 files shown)


thiagomacieira:
Already merged elsewhere.

PR #494, opened 2024-03-09
FW: Fix bug in frequency manager

dilipshivaraju:
Return value of fscanf should be checked for EOF instead of relying on boolean logical operator.

Should be working fine:

```
 ./opendcdiag -vv -e eigen_gemm_double14 -e zlib -t30s --vary-frequency
./opendcdiag: Cannot find "userspace" scaling governor from the file: /sys/devices/system/cpu/cpu0/cpufreq/scaling_available_governors
```

### DIFF
--- a/framework/frequency_manager.hpp
+++ b/framework/frequency_manager.hpp
@@ -111,7 +111,7 @@ private:
             exit(EX_IOERR);
         }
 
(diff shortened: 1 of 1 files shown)


dawid-sabat (approved):
Looks good, doesn't hang now.

PR #495, opened 2024-03-11
Refactor/simplify test management

busykai:
This PR replaces entire test_selector infrastructure to pick the next test. It separates the functionality of providing the next test to run into `SandstoneTestSet` class whose only contract is to respond to the question "what is the next test to run?". For this purpose, it provides an iterator which is used by opendcdiag to cycle through the tests. It ensures a cleaner interface to configure and run the tests. It also simplifies greatly the complexity and maintainability of the implementation. The behavior of the iterator is configurable to provide:

- Randomness of the next test selection.
- "Cycle though" behavior: an infinite iterator over the tests.
- An option to ignore tests with unknown names.

This keeps it compatible with the previous implementation.

Additionally,

- Current test count is kept globally and is not reset when the next iteration starts;
- All special handling of `mce_check` test is removed from the main cycle and is encapsulated in the iterator instead.
- The options to use built-in test list, test list file, as well as command-line options are now can be used together and combined. For example, an undesired test from a test list file can be disabled on the command-line using `--disable` option as opposed to editing the test list file needed before.


framework/sandstone.cpp
```diff
@@ -3181,6 +3211,8 @@ int main(int argc, char **argv)
     const char *on_crash_arg = nullptr;
 
     // test selection
+    std::vector<char *> enabled_tests;
```
thiagomacieira, line 3209: `const char *` or `std::string_view`

framework/sandstone.cpp
```diff
@@ -3216,10 +3248,10 @@ int main(int argc, char **argv)
...
         switch (opt) {
         case disable_option:
-            disable_tests(test_set, optarg);
+            disabled_tests.push_back(optarg);
             break;
         case 'e':
-            add_tests(test_set, test_list, optarg);
+            enabled_tests.push_back(optarg);
             test_selection_strategy = Ordered;
             break;
```
thiagomacieira: Thank you, I've been meaning to do this for a while.

framework/sandstone.cpp
```diff
@@ -3589,6 +3621,90 @@ int main(int argc, char **argv)
...
+    SandstoneTestSet::init((sApp->shmem->selftest) ? SandstoneTestSet::SELF_TESTS : SandstoneTestSet::REGULAR_TESTS);
+
+    if (enabled_tests.size() || builtin_test_list_name || test_list_file_path) {
+        /* if anything other than the "all tests" has been specified, start with
+         * an empty list. */
+        test_set = new SandstoneTestSet(SandstoneTestSet::NO_TESTS);
+    } else {
+        /* otherwise, start with all the applicable tests (self tests or
+         * regular. */
+        test_set = new SandstoneTestSet((sApp->shmem->selftest) ? SandstoneTestSet::SELF_TESTS : SandstoneTestSet::REGULAR_TESTS);
```
thiagomacieira, line 3640: Sounds redundant with the `init` call above. Any way to simplify?

framework/sandstone.cpp
```diff
@@ -3246,6 +3278,7 @@ int main(int argc, char **argv)
         case 'l':
         case raw_list_tests:
         case raw_list_groups:
+            test_set = new SandstoneTestSet(SandstoneTestSet::REGULAR_TESTS);
```
thiagomacieira: This means we can't list the selftests. Please don't remove that ability.

framework/sandstone.cpp
```diff
@@ -3589,6 +3621,90 @@ int main(int argc, char **argv)
...
+        if (builtin_test_list_name)
+            logging_printf(LOG_LEVEL_QUIET,
+                           "# WARNING: both --test-list-file and --use-builtin-test-list "
+                           "specified, using test file \"%s\".\n", test_list_file_path);
+        if (test_list.size()) {
+            logging_printf(LOG_LEVEL_QUIET,
+                           "# WARNING: both --test-list-file and --enable specified, using only "
+                           "the test list file \"%s\".\n", test_list_file_path);
+            test_list = {};
+        }
```
thiagomacieira: Shouldn't this be above the enable/disable code above?

framework/sandstone_tests.cpp
```diff
@@ -0,0 +1,102 @@
+#include <fnmatch.h>
+#include "sandstone_tests.h"
+
+static std::map<const char *, struct test *, SandstoneTestSet::cstr_cmp> all_test_map; /* maps test name to test */
+static std::map<const char *, std::vector <struct test *>> all_group_map; /* maps group name to a vector of tests it contains */
```
thiagomacieira: Please don't use `std::map` globals (they have an init function). Move these to inside `SandstoneApplication`.

Use `std::string_view` as the key type instead of `const char *`, so you don't have to have `cstr_cmp`.

framework/sandstone_tests.cpp
```diff
@@ -0,0 +1,102 @@
...
+void SandstoneTestSet::init(enum InitSet init_set)
+{
+    std::span<struct test> known_tests;
+    switch (init_set) {
+        case REGULAR_TESTS:
+            known_tests = regular_tests;
+            break;
+        case SELF_TESTS:
+            known_tests = selftests;
+            break;
```
thiagomacieira, line 18: `#ifndef` for no-self-tests

framework/sandstone_tests.cpp
```diff
@@ -0,0 +1,102 @@
...
+std::vector<struct test *> SandstoneTestSet::lookup(const char *name)
+{
+    std::vector<struct test*> res;
+    if (!name || !strlen(name)) return res;
+    if (!strchr(name, '*')) { /* if it has an asterisk, it's a glob, so expand it. */
+        try {
+            struct test *t = all_test_map.at(name);
+            res.push_back(t);
+        } catch(const std::out_of_range &e) {
+        }
```
thiagomacieira: Don't use the exception-throwing lookup. Use `find` instead.

framework/sandstone_tests.h
```diff
@@ -50,6 +53,87 @@ __attribute__((weak)) extern const struct test_group __stop_test_group;
 inline std::span<struct test> regular_tests = { &__start_tests, &__stop_tests };
 extern const std::span<struct test> selftests;
 
+#if defined(__linux__) && defined(__x86_64__)
+extern struct test mce_test;
+#else
+// no MCE test outside Linux
+static_assert(!InterruptMonitor::InterruptMonitorWorks);
+struct test mce_test = {
```
thiagomacieira, line 37: Will cause duplicate definitions. Move this to the .cpp.

framework/sandstone_tests.cpp
```diff
@@ -20,6 +21,13 @@ void SandstoneTestSet::init(enum InitSet init_set)
     }
     for (struct test &t: known_tests) {
         all_test_map[t.id] = &t;
+        for (auto ptr = t.groups; ptr && *ptr; ++ptr) {
+            if (!all_group_map.contains((*ptr)->id)) {
+                std::vector<struct test *> group_tests;
+                all_group_map[(*ptr)->id] = group_tests;
+            }
+            all_group_map[(*ptr)->id].push_back(&t);
```
thiagomacieira, line 29: Simpler:
```c++
    all_group_map[(*ptr)->id].push_back(&t);
```

thiagomacieira (changes requested):

framework/sandstone.cpp
busykai: `init` creates a catalog of all known tests, constructor is used to creates a instance which contains the tests to be run in this execution of opendcdiag. That's the difference. The reason it's not simply boolean in the constructor is that we need to start with an empty set at times (e.g. when `-e` is specified on the command line).

framework/sandstone.cpp
```diff
@@ -3556,6 +3592,52 @@ int main(int argc, char **argv)
...
+        /* otherwise, start with all the applicable tests (self tests or
+         * regular. */
+        test_set = new SandstoneTestSet(true, sApp->shmem->selftest);
+    }
+
+    test_set->set_ignore_unknown_tests(ignore_unknown_tests);
+
+    /* Add all the tests we were told to enable. */
+    if (enabled_tests.size()) {
+        for (char *name : enabled_tests) {
```
thiagomacieira: `const char *`, otherwise this can't compile.

framework/sandstone.cpp
```diff
@@ -3736,8 +3822,9 @@ int main(int argc, char **argv)
     bool restarting = true;
     int total_tests_run = 0;
     TestResult lastTestResult = TestResult::Skipped;
+    auto it = test_set->begin();
 
-    for (struct test *test = get_next_test(tc); test; test = get_next_test(tc)) {
+    for (struct test *test = get_next_test(it, tc); test; test = get_next_test(it, tc)) {
```
thiagomacieira: And if you use the struct return type I suggested above, this would look like, here:
```c++
    for (auto next = get_next_test(test_set->begin(), tc); next.test; next = get_next_test(next.it, tc))
```

An alternative is to create a Java iterator-like struct:
```c++
    struct TestIteration iteration(test_set);
    while (iteration.hasNext()) {
        struct test *test = iteration.test;
        int tc = iteration.tc;
        ....
```

framework/sandstone_tests.h
```diff
@@ -11,20 +11,85 @@
...
+extern struct test mce_test;
+
 #ifdef __cplusplus
 
 __attribute__((weak)) extern const struct test_group __start_test_group;
 __attribute__((weak)) extern const struct test_group __stop_test_group;
 
 inline std::span<struct test> regular_tests = { &__start_tests, &__stop_tests };
+#ifdef NO_SELF_TESTS
+const std::span<struct test> selftests = {};
```
thiagomacieira, line 57: `constexpr`

framework/sandstone_tests.h
```diff
@@ -11,20 +11,85 @@
...
+    SandstoneTestSet() : SandstoneTestSet(false) {
+    };
+
+    /* searches the test by name in all the available tests. */
+    TestSet lookup(const char *name);
+
+    SandstoneTestSet(struct test_set_cfg cfg) : SandstoneTestSet(cfg, false) {
+    };
+
+    SandstoneTestSet(bool all_tests, bool is_selftest);
```
thiagomacieira: Boolean trap. Please create one or two enums for this, for flags.
```c++
    enum Flag {
        AddAllTests = 0x01,
        IgnoreUnknownTests = 0x02,
...
        UsingSelfTests = 0x80,
    };
    SandstoneTestSet(unsigned flags); // no QFlags
```

framework/sandstone_tests.h
```diff
@@ -11,20 +11,85 @@
...
+    TestSet::iterator begin() { return test_set.begin(); };
+    TestSet::iterator end () { return test_set.end(); };
+
+    std::vector<struct test_cfg_info> disable(const char *name);
+    struct test_cfg_info disable(struct test *t);
+
+    std::vector<struct test_cfg_info> enable(const char *name);
+    struct test_cfg_info enable(struct test *t);
+
+    inline bool is_disabled(const char *name) { return test_map.contains(name) ? (test_map[name].st == TEST_DISABLED) : true; };
```
thiagomacieira, line 139: Double map lookup (twice O(log n)). Instead:
```c++
    auto it = test_map.find(name); 
    if (it != test_map.end())
        return it->st == TEST_DISABLED;
    return true;
```

framework/sandstone_tests.h
```diff
@@ -11,20 +11,85 @@
...
+    inline bool is_enabled(const char *name) { return !is_disabled(name); };
+    
+    static std::vector<struct test *> lookup(const char *name);
+
+private:
+    bool is_selftest;
+    bool ignore_unknown_tests = true;
+    TestSet test_set;
+
+    typedef enum {
```
thiagomacieira: No `typedef` necessary. This isn't C.
```c++
    enum test_rt_status {
        TEST_DISABLED = 0,
        TEST_ENABLED = 1,
    };
```
(yes, please make disabled = 0)

framework/sandstone_tests.h
```diff
@@ -11,20 +11,85 @@
...
+        TEST_DISABLED,
+    } test_rt_status;
+
+    struct test_info {
+        struct test *test;
+        test_rt_status st;
+    };
+
+    /* list of all available tests. */
+    TestSet all_tests;
```
thiagomacieira: Member ordering: first the vectors and maps, then the enums, then any booleans.

meson.build
```diff
@@ -100,6 +100,7 @@ add_project_arguments([
     '-D__USE_MINGW_ANSI_STDIO',
     '-D_POSIX',
     '-D_WIN32_WINNT=0x0A00',
+    '-D_NO_SYS_GUID_OPERATOR_EQ_',
```
thiagomacieira, line 103: What's this for?

framework/sandstone_tests.h
thiagomacieira, line 139: Hint: `inline` is unnecessary here.

framework/sandstone_tests.cpp
```diff
@@ -0,0 +1,232 @@
...
+    if (error) {
+        res.clear();
+    }
+    return res;
+}
+
+SandstoneTestSet::TestSet SandstoneTestSet::add_test_list(const char *fname, std::vector<std::string> &errors)
+{
+    TestSet res;
+    std::ifstream list_file(fname, std::ios_base::in);
```
thiagomacieira, line 235: I don't see the include for `<fstream>` nor for `stringstream` above. Please don't ever include `<iostream>` due to how it generates pre-main code

framework/sandstone_tests.cpp
```diff
@@ -0,0 +1,232 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <fnmatch.h>
+#include "sandstone_p.h"
+#include "sandstone_tests.h"
+
+#if defined(__linux__) && defined(__x86_64__)
+extern struct test mce_test;
+#else
```
thiagomacieira, line 17: `extern` already in the header. Just collapse to `#ifndef`

framework/sandstone_tests.cpp
```diff
@@ -0,0 +1,232 @@
...
+    if (!name || !strlen(name)) return res;
+    if (strchr(name, '*')) { /* if it has an asterisk, it's a glob, so expand it. */
+        for (struct test *t : all_tests) {
+            if (!fnmatch(name, t->id, 0))
+                res.push_back(t);
+        }
+    } else if (name[0] == '@') { /* it's a group name */
+        const char *group_name = name + 1;
+        if (all_group_map.contains(group_name)) {
+            res = all_group_map[group_name];
```
thiagomacieira, line 60: Double map lookup.
```c++
    auto it = all_group_map.find(group_name);
    if (it != all_group_map.end())
        res = *it;
```

framework/sandstone_tests.h
```diff
@@ -11,20 +11,85 @@
...
+    using TestSet = std::vector<struct test *>;
+public:
+
+    SandstoneTestSet() : SandstoneTestSet(false) {
+    };
+
+    /* searches the test by name in all the available tests. */
+    TestSet lookup(const char *name);
+
+    SandstoneTestSet(struct test_set_cfg cfg) : SandstoneTestSet(cfg, false) {
```
thiagomacieira, line 111: `test_set_cfg` isn't declared anywhere.

framework/sandstone_tests.cpp
```diff
@@ -0,0 +1,232 @@
...
+    switch (c) {
+        case '#':
+        case ':':
+            return true;
+        default:
+            return false;
+    }
+}
+
+typedef enum {
```
thiagomacieira, line 159: Ditto about `typedef enum`

framework/sandstone_tests.h
```diff
@@ -11,20 +11,85 @@
...
+    TestSet test_set;
+
+    typedef enum {
+        TEST_ENABLED,
+        TEST_DISABLED,
+    } test_rt_status;
+
+    struct test_info {
+        struct test *test;
+        test_rt_status st;
```
thiagomacieira: Why not just `is_enabled = false` ?

thiagomacieira (changes requested):

framework/sandstone_tests.cpp
```diff
@@ -3,6 +3,11 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+#include <charconv>
+#include <fstream>
+#include <iostream>
```
thiagomacieira, line 8: Please don't include `<iostream>` because of how it generates pre-`main()` code.
https://github.com/gcc-mirror/gcc/blob/releases/gcc-14.1.0/libstdc%2B%2B-v3/include/std/iostream#L78-L83

framework/sandstone_tests.h
```diff
@@ -23,6 +23,18 @@ extern struct test __stop_tests;
 
 extern struct test mce_test;
 
+typedef enum {
+    TEST_ENABLED,
+    TEST_DISABLED,
+    TEST_NOT_FOUND,
+} test_cfg_status;
```
thiagomacieira, line 33:
```c++
enum class test_cfg_status {
    disabled = 0,
    enabled = 1,
    not_found = -1,
};
```
The `class` forces you to use scoping: `test_cfg_status::enabled`. Or move it inside the `struct tst_cfg_info` struct below.

framework/sandstone_tests.h
```diff
@@ -23,6 +23,18 @@ extern struct test __stop_tests;
...
+
+struct test_cfg_info {
+    struct test *test;
+    test_cfg_status st;
```
thiagomacieira, line 46: Spell out `status`

framework/sandstone_tests.h
```diff
@@ -23,6 +23,18 @@ extern struct test __stop_tests;
...
+    std::string attribute;
```
thiagomacieira, line 47: Given you have a non-trivial type here, please initialise the other two members. While you're at it, please reorder:
```c++
struct test_cfg_info
{
    enum test_status {
        disabled = 0,
        enabled = 1-,
        not_found = -1,
    };
    struct test *test = nullptr;
    std::string atribute;
    test_status status = not_found;
};
```

framework/sandstone_tests.h
```diff
@@ -88,6 +93,7 @@ class SandstoneTestSet
     std::map<std::string_view, struct test_cfg_info> test_map;
 
     void load_all_tests();
+
```
thiagomacieira, line 168: Nitpick: unnecessary empty line.

thiagomacieira (changes requested):

framework/sandstone_tests.h
```diff
@@ -52,23 +61,20 @@ class SandstoneTestSet
...
     TestSet lookup(const char *name);
 
     SandstoneTestSet(struct test_set_cfg cfg) : SandstoneTestSet(cfg, false) {
     };
 
-    SandstoneTestSet(bool all_tests, bool is_selftest);
+    SandstoneTestSet(struct test_set_cfg cfg, bool all_tests) : SandstoneTestSet(cfg, true, false) {
+    };
+
+    SandstoneTestSet(struct test_set_cfg cfg, bool all_tests, bool is_selftest);
```
thiagomacieira, line 117: Either merge the `all_tests` and `is_selftest` booleans into the config or move the three new booleans into the flags enum that I suggested earlier.

framework/sandstone_tests.h
```diff
@@ -11,6 +11,7 @@
 
 #ifdef __cplusplus
 #include <span>
+#include <algorithm>
```
thiagomacieira, line 16: Alphabetical order

framework/sandstone_tests.h
```diff
@@ -11,20 +11,85 @@
 
 #ifdef __cplusplus
 #include <span>
+#include <cstring>
+#include <iterator>
+#include <map>
```
thiagomacieira: Nitpick: alphabetical order.

`<string.h>` is already included and you're not using `std::` with any of the functions below, so the `<cstring>` is superfluous.

framework/sandstone_tests.h
```diff
@@ -72,8 +110,17 @@ class SandstoneTestSet
 
     SandstoneTestSet(struct test_set_cfg cfg, bool all_tests, bool is_selftest);
 
-    TestSet::iterator begin() { return test_set.begin(); };
-    TestSet::iterator end () { return test_set.end(); };
+    TestSetIterator begin() {
+        static TestSet::iterator end = test_set.end();
```
thiagomacieira, line 120: Why `static` ?

framework/sandstone_tests.h
```diff
@@ -91,6 +138,15 @@ class SandstoneTestSet
...
     TestSet test_set;
 
+    /* Make a Uniform Random Bit Generator to use our own random as opposed to
+     * standard library's. */
+    struct SandstoneURBG {
+        using result_type = uint32_t;
+        static constexpr result_type min() { return 0; };
+        static constexpr result_type max() { return UINT32_MAX; };
+        result_type operator()() { return random32(); };
+    };
```
thiagomacieira, line 158: Might be useful to have available for tests too, at some point. Good idea.

framework/sandstone_tests.h
```diff
@@ -58,9 +59,46 @@ extern const std::span<struct test> selftests;
...
+        TestSetIterator(TestSet::iterator it, TestSet::iterator *end) : it(it), start(it), end(end) {};
+
+        reference operator*() const { return *it; };
+        pointer operator->() { return &(*it); };
+
+        TestSetIterator& operator++() { next(); return *this; };
+        TestSetIterator operator++(int) { TestSetIterator ret = *this; ++(*this); return ret; };
+
+        friend bool operator== (const TestSetIterator &a, const TestSetIterator &b) { return a.it == b.it; };
+        friend bool operator!= (const TestSetIterator &a, const TestSetIterator &b) { return a.it != b.it; };
```
thiagomacieira, line 89: Add `noexcept` to all of these functions and to `next()` below.

framework/sandstone.cpp
thiagomacieira: I see changes in later commits that may obviate this. Would be nice if you said in the commit message if this is temporary.

thiagomacieira (changes requested):

tests/smi_count/smi_count.cpp
```diff
@@ -40,11 +41,11 @@ static int smi_count_run(struct test *test, int cpu)
     return EXIT_SUCCESS;
 }
 
-
 DECLARE_TEST(smi_count, "Counts SMI events")
+    .test_init = [](struct test *t) { return InterruptMonitor::InterruptMonitorWorks ? EXIT_SUCCESS : EXIT_SKIP; },
     .test_run = smi_count_run,
+    .test_cleanup = [](struct test * t) { initialize_smi_counts(); return EXIT_SUCCESS; },
```
thiagomacieira, line 47: This runs in the child process, so its effects will be discarded. Might as well not do anything.

thiagomacieira (changes requested):

framework/sandstone.cpp
```diff
@@ -3311,6 +3311,9 @@ int main(int argc, char **argv)
             } else {
                 sApp->endtime = sApp->starttime + string_to_millisecs(optarg);
             }
+            sApp->shmem->use_strict_runtime = true;
```
thiagomacieira, line 3316: This is a behaviour change. Why is this here?

framework/sandstone.cpp
```diff
@@ -3618,6 +3621,10 @@ int main(int argc, char **argv)
         }
     }
 
+    /* Add mce_check as the last test to the set. It will be kept last by
+     * SandstoneTestSet in case randomization is requested. */
+    test_set->enable(mce_test.id);
```
thiagomacieira: Given we have the `struct test *` for `mce_test`, why search for it again by id?

Also, this may make more sense to exist inside the `test_set` object anyway, as it is responsible for keeping it last.

framework/sandstone_tests.cpp
```diff
@@ -83,37 +83,47 @@ SandstoneTestSet::SandstoneTestSet(struct test_set_cfg cfg, bool all_tests, bool
...
+struct test_cfg_info SandstoneTestSet::enable(struct test *t) {
+    struct test_cfg_info ti;
+    if (!test_map.contains(t->id)) {
+        ti.st = TEST_ENABLED;
+        ti.test = t;
+        test_map[t->id] = ti;
+    } else {
+        ti = test_map[t->id];
+        ti.st = TEST_ENABLED;
+    }
```
thiagomacieira, line 95: Unnecessary double map lookup. You always end up with the test in enabled state. If you ensure that `test_cfg_info` always initialises its members, you can reduce this to:
```c++
    test_cfg_info &ti = test_map[t->id];
    ti.status = test_cfg_info::enabled;
    ti.test = t;
    return ti;
```

framework/sandstone_tests.cpp
```diff
@@ -83,37 +83,47 @@ SandstoneTestSet::SandstoneTestSet(struct test_set_cfg cfg, bool all_tests, bool
...
 
+struct test_cfg_info SandstoneTestSet::disable(struct test *t) {
+    struct test_cfg_info ti;
+    if (test_map.contains(t->id)) {
+        test_map[t->id].st = TEST_DISABLED;
+    } else {
+        ti.st = TEST_DISABLED;
+        ti.test = t;
+        test_map[t->id] = ti;
+    }
```
thiagomacieira, line 118: Ditto.

framework/sandstone_tests.cpp
```diff
@@ -225,7 +227,26 @@ SandstoneTestSet::TestSet SandstoneTestSet::add_test_list(const char *fname, std
     std::vector<struct test_cfg_info> entries = load_test_list(list_file, this, cfg.ignore_unknown_tests, errors);
     if (!errors.empty()) return res;
     for (auto e : entries) {
-        test_set.push_back(e.test);
+        if (!e.attribute.empty()
+                && !std::ranges::equal(e.attribute, std::string("default"), [] (const char& a, const char& b) { return (std::tolower(a) == std::tolower(b)); })) {
```
thiagomacieira, line 240: Line too long, please split.

Please don't use `std::ranges::equal` with a char-by-char lowercasing. That's rule 1 of lowercasing: it's not a character operation. Instead, just do a `strcasecmp`.

`std::ranges::equal` was added in GCC 11. I think all our internal targets support that, but I'm not sure.

framework/sandstone_tests.cpp
```diff
@@ -225,7 +227,26 @@ SandstoneTestSet::TestSet SandstoneTestSet::add_test_list(const char *fname, std
...
+            /* the only attribute supported by the test list is time in
+             * milliseconds. */
+            uint64_t test_duration_ms;
+            std::string attr = e.attribute;
+            auto res = std::from_chars(attr.data(), attr.data() + attr.size(), test_duration_ms);
```
thiagomacieira: Use `string_to_millisecs` from <sanstone_chrono.h>. That will also allow us to specify units instead of pure milliseconds.

thiagomacieira (changes requested):

meson.build
busykai: Windows defines global == operators which conflict with these:
https://github.com/opendcdiag/opendcdiag/blob/c73dc66bd62ca4ea62c2a1f3fabc8862dd9e7d03/framework/sandstone_tests.h#L88-L89

thiagomacieira: Got it. It deserves a comment, though.

framework/sandstone.cpp
busykai: If strict runtime is not enforced, it's not stopping on a truly infinite supply of the tests, even if allocated time (with `-T`) has expired.

framework/sandstone_tests.cpp
```diff
@@ -251,3 +251,18 @@ SandstoneTestSet::TestSet SandstoneTestSet::add_test_list(const char *fname, std
     return res;
 }
 
+SandstoneTestSet::TestSet SandstoneTestSet::add_builtin_test_list(const char *name, std::vector<std::string> &errors) {
+    TestSet res;
+    BuiltinTestSet builtin = get_builtin_test_set(name);
+    if (!builtin.tests) {
+        std::stringstream msg_strm;
+        msg_strm << "Builtin test list '" << name << "' does not contain any tests.";
```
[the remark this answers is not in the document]
busykai: The default test list if nothing has been configured does not contain any tests. There's no way currently to distinguish between "does not exist" and "has no tests" situations, strictly speaking.

tests/smi_count/smi_count.cpp
busykai: Yes, I was conscious of that, but I wanted to have it there for potentially changing this to no-fork mode (and then it will be important to reset it here).

framework/sandstone_tests.cpp
```diff
@@ -0,0 +1,255 @@
...
+
+std::vector<struct test_cfg_info> SandstoneTestSet::add_test_list(const char *fname, std::vector<std::string> &errors)
+{
+    std::ifstream list_file(fname, std::ios_base::in);
+    std::vector<struct test_cfg_info> entries = load_test_list(list_file, this, cfg.ignore_unknown_tests, errors);
+    if (!errors.empty()) return {};
+    for (auto e : entries) {
+        if (!e.attribute.empty()
+                && strcasecmp(e.attribute.c_str(), "default")) {
+            e.test->desired_duration = string_to_millisecs(e.attribute).count();
```
thiagomacieira, line 230: I've just debugged a buggy test that ran into an infinite loop when its `desired_duration` field was changed from -1 to a positive value. Can you fix this here so we never reset a -1?

thiagomacieira (commented):
All the squashes look good!

framework/sandstone_tests.cpp
busykai: Done. See the last commit.

framework/sandstone.cpp
thiagomacieira: We'll need to document this.

framework/sandstone.cpp
```diff
@@ -3638,18 +3673,74 @@ int main(int argc, char **argv)
...
+            // FIXME: handle errors
+            ;
+            exit(EX_USAGE);
+        }
+    }
+
+
+    /* Add mce_check as the last test to the set. It will be kept last by
+     * SandstoneTestSet in case randomization is requested. */
+    test_set->enable(&mce_test);
```
thiagomacieira, line 3723: This here could use an `if (InterruptMonitor::InterruptMonitorWorks)`

thiagomacieira (approved):

PR #497, opened 2024-03-27
simd.conf: Enable APX-F bit.

busykai:


thiagomacieira (changes requested):
We're out of bits in the CPU features bit field. Adding this one will push lzcnt over the edge:
```c++
// in CPUID Leaf 80000001h, ECX:
#define cpu_feature_lzcnt                           (UINT64_C(1) << 63)
```

We need to either trim the feature list or go to 128-bit integers. If the latter, the Perl script needs to be updated to emit 128-bit constants.

thiagomacieira (approved):
Never mind, my build has one feature more that pushed `lzcnt` to 63:
```c++
#define cpu_feature_apx_f                           (UINT64_C(1) << 62)

// in CPUID Leaf 80000001h, ECX:
#define cpu_feature_lzcnt                           (UINT64_C(1) << 63)
```

PR #498, opened 2024-03-28
FW/Logging: obey --max-logdata for the memcmp_or_fail mem dump

thiagomacieira:
It's often useful to get the full buffers for debugging, because either you need to interpret the data that came before to make sense of what you're seeing or because you want to see if more errors have shown up.

This removes the hardcoded 64 from the source file. The default value of --max-logdata is 128, which this commit does not change behaviour of the defaults.

[ChangeLog][Framework] The --max-logdata option now also controls the memory dumps when memory miscompares are detected. In particular, --max-logdata=0 causes the full blocks that failed to compare to be printed.


AthenasJimenez (approved):
LGTM

AthenasJimenez (approved):

PR #500, opened 2024-04-09
Tests/Eigen: make the dimensions changeable at runtime

thiagomacieira:
All of these tests use `Dynamic` as their structural size (i.e., not a fixed at compile-time). So do allow changing them with a command-line option: we've had an interest in doing that to measure how the test behaves.

### DIFF
--- a/tests/eigen_svd/sandstone_eigen_common.h
+++ b/tests/eigen_svd/sandstone_eigen_common.h
@@ -32,23 +32,28 @@ template <typename SVD, int Dim> struct EigenSVDTest
     }
 
     template <typename FP> static inline std::enable_if_t<boost::is_complex<FP>::value>
-    compare_or_fail(const FP *actual, const FP *expected, const char *name)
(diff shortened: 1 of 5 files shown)


tests/eigen_svd/sandstone_eigen_common.h
```diff
@@ -32,23 +32,28 @@ template <typename SVD, int Dim> struct EigenSVDTest
...
+                       2 * rows * cols, name);
     }
 
     template <typename FP> static inline std::enable_if_t<!boost::is_complex<FP>::value>
-    compare_or_fail(const FP *actual, const FP *expected, const char *name)
+    compare_or_fail(const FP *actual, const FP *expected, typename Mat::Index rows,
+                    typename Mat::Index cols, const char *name)
     {
-        memcmp_or_fail(actual, expected, Dim * Dim, name);
+        memcmp_or_fail(actual, expected, rows * cols, name);
```
busykai, line 47: This should be correlated to the FP size. Otherwise, it's shorter region to compare.

busykai (changes requested):

PR #502, opened 2024-04-10
FW: Fix deprecation warning.

busykai:
optind points to the next option to be processed, so pass the argument to getopt to get the current index.

Before:
```
./opendcdiag: option '(null)' is ignored and will be removed in a future version.
```
After:
```
./opendcdiag: option 'schedule-by' is ignored and will be removed in a future version.
```


lukasz-tuz (approved):
Looks good!

PR #503, opened 2024-04-15
ifs test: skip reason when non linux

alexjch:
Provide a skip reason when the tests: ifs, ifs_array_bist are run in an OS that is not linux.


dawid-sabat (approved):
LGTM

@AthenasJimenez ?

AthenasJimenez:
@alexjch, I think I'm not following, why is needed to add a skip message?

PR #505, opened 2024-04-30
FW/Frequency Manager: Handle "userspace" not present scenario

dilipshivaraju:
Older version exits the framework with error message if `userspace` is not present in sysfs file `/sys/devices/system/cpu/cpu%/cpufreq/scaling_available_governors`.

In this version, the framework will enable `userspace` if not present and set the system back to it's original state at the end.


lukasz-tuz (approved):
Looks good.

PR #508, opened 2024-05-08
kvm: windows skip generic functions

alexjch:
Current generic functions used in kvm are available only in linux, this change provides the same functions declaration for windows builds so a skip message can be printed.

Signed-off-by: Alex V Jaramillo <alex.v.jaramillo@intel.com>


dawid-sabat (approved):
LGTM
@thac0 could you take a look on this? I don't see anything disturbing.

PR #510, opened 2024-05-17
FW/Win32: add --on-hang=gdb

thiagomacieira:
I tested and it works. I'd like to expand it to have more functionality like the Unix version, but that's a job for a later time.


AthenasJimenez (approved):
LGTM

PR #512, opened 2024-05-23
FW/Win32: add some more process-summary information

thiagomacieira:
Available in -vvv output or in case of hang.

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -77,6 +77,7 @@
 #  include <ntstatus.h>
 #  include <shlwapi.h>
 #  include <windows.h>
+#  include <psapi.h>
 #  include <pdh.h>
 
 #  ifdef ftruncate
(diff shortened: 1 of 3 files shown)


framework/sandstone.cpp
```diff
@@ -1663,6 +1664,24 @@ static void wait_for_children(ChildrenList &children, int *tc, const struct test
 
         auto childResult = test_result_from_exit_code(result);
         childResult.endtime = MonotonicTimePoint::clock::now();
+        if (FILETIME dummy, stime, utime; GetProcessTimes(hExited, &dummy, &dummy, &stime, &utime)) {
+            auto cvt = [](FILETIME f) {
+                uint64_t us = f.dwHighDateTime;
+                us = (us << 32) | f.dwLowDateTime;
+                us /= 10;   // FILETIME stores tenths of microseconds
```
AthenasJimenez, line 1671: Shouldn't this need to be `us *= 10`?

AthenasJimenez (changes requested):

framework/sandstone.cpp
thiagomacieira: Yes.

AthenasJimenez (approved):

PR #513, opened 2024-05-23
FW/Win32: print PIDs not HANDLEs when blaming frozen children

thiagomacieira:
The HANDLEs are local to the current process. The PID can be used elsewhere, like Task Manager.


AthenasJimenez (approved):
Makes sense.

PR #515, opened 2024-06-10
openssl_sha: Print skip in windows

alexjch:
Before the change this test will not be build in windows. This change makes possible for the test to skip when executing in windows OS and prints a not supported in OS skip message.


dawid-sabat (approved):
LGTM

AthenasJimenez (approved):
LGTM

PR #516, opened 2024-06-13
FW/Unix: kill the children using their PGID, not PID

thiagomacieira:
In `signals_init_child()`, we call `setsid()`, so each direct child we start is in its own process group and session, whose IDs match the child PID. We do that so the main process gets Ctrl+C from the terminal, and all it did was relay that to each child.

This commit makes the Unix version instead relay to each child's process group. That way, if a child forks and has children of its own, all those children will get the signal too. The unit tests I added or modified confirm that the grandchildren die in both timeout and the signal cases (Ctrl+C/SIGINT, SIGTERM, SIGHUP).

### DIFF
--- a/bats/sanity-check/20-selftests.bats
+++ b/bats/sanity-check/20-selftests.bats
@@ -1301,6 +1301,37 @@ function selftest_logerror_common() {
     done
 }
 
+@test "selftest_freeze_fork" {
+    if $is_windows; then
+        skip "Unix-only test"
+    fi
+    declare -A yamldump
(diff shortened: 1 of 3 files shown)


jposwiata (approved):
If the test uses fork, there's no need to handle it separately (with pthread_cleanup() or any other "smart" implementations).

thiagomacieira:
> If the test uses fork, there's no need to handle it separately (with pthread_cleanup() or any other "smart" implementations).

There's still some danger, because if the middle process exits, the grandchild gets a new session ID and is reparented to PID 1. Unix process groups aren't a tight seal; we'd need cgroups for that.

But so long as the middle process isn't exiting before its time, this should be enough.

PR #517, opened 2024-06-25
Required file for OSSF Scorecard

rdower:


SECURITY.md
```diff
@@ -0,0 +1,5 @@
+# Security Policy
+Intel is committed to rapidly addressing security vulnerabilities affecting our customers and providing clear guidance on the solution, impact, severity and mitigation.
```
fenrus75, line 2: This is not an Intel only project so this sentence is inappropriate.

thiagomacieira: True, but can we say that, in lieu of an OpenDCDiag procedure, Intel is providing security support for this project?

busykai: `The maintainers of this project are committed to...` and further.

PR #525, opened 2024-07-10
[signal handling] reportFail() in main process

jposwiata:
It is allowed to call reportFail() from <<main>> process to fail the test.
Small fixes to allow better signal handling in multithraded scenarios (when signal is blocked in some threads, esp. in the one that just has failed on critical condition).


framework/sysdeps/unix/child_debug.cpp
```diff
@@ -210,14 +210,11 @@ static void child_crash_handler(int, siginfo_t *si, void *ucontext)
...
-        signal(si->si_signo, SIG_DFL);
-        if (si->si_signo != SIGABRT) {
-            // if we return, we'll execute the same instruction and crash again
-            // but the core dump should point to the exact locus now
-            return;
-        }
-        raise(si->si_signo);
+        // restore the signal handler so the process can be killed
+        signal(sig, SIG_DFL);
+        // allow the app to kill all threads (not only current one)
```
thiagomacieira, line 215: This change I don't agree with. Please keep the ability to return on crash.

thiagomacieira (dismissed):

framework/sysdeps/unix/child_debug.cpp
jposwiata: Some libraries sends signal to current process to notify about "critical" conditions (FPE in this particular case).
Is "special handling" for ABORT and FPE better here?
If they send FPE, probably other signals might be used as critical notifications (and "return" causes ignoring it and might run unexpected/unreachable code). It is not possible to see exact list and scenarios of possible signals.

thiagomacieira: No, it's not. I need to understand what exactly is required by those libraries, because your explanation doesn't make sense, at least not in context of the change you're making.

This function handler handles five different signals (SIGILL, SIGABRT, SIGFPE, SIGBUS, SIGSEGV) and, after doing what it needs to do, it resets the handler to the system default and re-delivers the same signal back to the application. Your change **still** does that, it only changes the means by which we deliver the four crashing signals.

So you need to explain why the means by which we deliver those four makes a difference. Or, more likely, why you think it does, because it shouldn't (doesn't!) make a difference. There's a misunderstanding somewhere.

jposwiata: Frankly speaking, I don't know the library requirements (and have no option to check "why"). These are only my pure observations. Stack trace is not very useful either ("unknown addresses").
If the signal is sent on "real" rootcause (e.g. illegal instruction), PC is not updated (it still points to the instruction which has failed, and cpu tries to execute this instruction once again after return from the handler). If the signal is from `kill()`/`raise()` the code freely continue (without clear "terminate" disposition, `kill()`/`raise()` has already succeeded and won't be redone).
In the context I mentioned, the `if (sig == SIGABRT)` condition is completely unclear for me. Do we expect identical handling for UD2 and `kill SIGABRT`? Why only for ABORT? In my particular case SIGFPE is sent "from somewhere" (different thread, it is not failure in single instruction, this is probably an "effect" of data validity checking). It seems that SIGFPE is "yet another" case where both failed instruction and/or signal should be handled in the same way as ABORT. From that point I extended all 5 signals to be handled in the same way (as not sure if the library doesn't send any other signals for any purpose..). The only disadvantage is a bit longer stack trace in the core file.
This change is intended to call default handler to report the **whole process** signaled. This will be caught by "test runner child" and handled properly with `IFSIGNALLED()` (without introducing "special" cases for `WIFEXITED()`)

thiagomacieira: Posting conclusions from direct discussion: it looks like a test (or library used by a test) is deliberately sending `SIGFPE` to the application, by way of `raise(SIGFPE);`. In that scenario, returning from the handler does not cause it to be re-raised. That's a use-case we need to handle. 

A solution to this is to inspect the `si_code` field of `siginfo` and see if it contains `SI_KERNEL`. If so, then the crashing signals came from the kernel and we can return. If it contains anything else or if the pointer is null, then we re-`raise()`.

thiagomacieira: Please add a self-test that `raise`s one of the crash signals and we do terminate the application with non-success state.

framework/sandstone.cpp
```diff
@@ -388,6 +388,13 @@ static void __attribute__((noreturn)) report_fail_common()
     /* does call the cleanup handalers */
     pthread_setcancelstate(PTHREAD_CANCEL_ENABLE, nullptr);
     pthread_setcanceltype(PTHREAD_CANCEL_ASYNCHRONOUS, nullptr);
+
+    // <<main>> thread is requested to report failure. We need to exit()
+    // whole process, and all running threads will be unconditionally killed
+    if (getpid() == gettid()) {
+        _exit(EXIT_FAILURE);
+    }
```
thiagomacieira, line 396: Please leave this and `selftest_reportfail_cleanup` test out. That's a different issue: one is a bug that needs fixing and the other is new feature for test construction.

framework/sysdeps/unix/child_debug.cpp
```diff
@@ -210,14 +210,18 @@ static void child_crash_handler(int, siginfo_t *si, void *ucontext)
...
         }
 
-        // restore the signal handler so we can be killed
-        signal(si->si_signo, SIG_DFL);
-        if (si->si_signo != SIGABRT) {
+        // restore the signal handler so the process can be killed
+        signal(sig, SIG_DFL);
+
+        // non-ABORT FROMKERNEL only
+        if ((sig != SIGABRT) && si && (si->si_code > 0)) {
```
thiagomacieira, line 217: Do we need to check the signal number? Wouldn't it be enough to get a `>= SI_KERNEL` here? The kernel will not send SIGABRT.

thiagomacieira (changes requested):

PR #527, opened 2024-07-12
CI: update the -T test

thiagomacieira:
This was added by commit 9601936870325fe2d857973bfd3f750135db0af0 and passed on my machine and in the CI, but I don't know how because it doesn't now. Maybe f60599d3b77d79dfd08a13b11c839b2dee98f352 has an effect.

In any case, we can't tell much about the time-at-end: `should_start_next_iteration()` only starts the next iteration if the remaining time until the deadline is more than the average iteration time. In CIs, we can't tell much about the average time and the deviation from the average for this next iteration can be huge.


dawid-sabat (approved):
Makes sense...

busykai (approved):

PR #529, opened 2024-07-17
FW: Add command line options `enable-by-feature` and `disable-by-feature`

dawid-sabat:
These new commandline options allows user enable/disable some tests by specifying a string with CPU features.

For example:

Let's have a subset of tests that require `avx512f` to run and another subset of tests that require `avx2` to run. We may have other kinds of tests which we don't want to run to shorten the runtime of OpenDCDiag.

We'd like to run one experiment that will run only `avx512f` and another one which will use `avx2` tests and last but not least both subsets during the same run.

The command line would look like this:

```
opendcdiag --enable-by-feature avx512f
opendcdiag --enable-by-feature avx2
opendcdiag --enable-by-feature avx512f,avx2
```

If we want to exclude some tests by feature we just simply need to use `--disable-by-feature` option.

If a feature that isn't supported by device uder test, the tests will skip. And the user will see a log skip message about not meeting test required features. The `enable-by-feature` and `disable-by-feature` options can also be combined with other commandline options. Especially with `--enable`, `--disable`, `--beta`, `-T` and `-t` and others.

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -112,11 +112,12 @@ enum {
     thirty_sec_option,
     two_min_option,
     five_min_option,
-
     cpuset_option,
     disable_option,
+    disable_by_feature_option,
     dump_cpu_info_option,
     fatal_skips_option,
+    enable_by_feature_option,
     gdb_server_option,
     ignore_os_errors_option,
     ignore_unknown_tests_option,
@@ -1326,6 +1327,45 @@ static std::string cpu_features_to_string(uint64_t f)
             comma = ",";
         }
     }
+    return result;
+}
+
+static uint64_t cpu_features_to_cpuid(std::string f_str)
+{
+    uint64_t result = 0;
+    uint32_t substrings_num = 0;
+    uint32_t substrings_found_num = 0;
+
+    // Split the input string by commas
+    std::vector<std::string_view> substrings;
+    size_t start = 0;
+    size_t end = f_str.find(',');
+
+    while (end != std::string_view::npos) {
+        substrings.push_back(f_str.substr(start, end - start));
+        substrings_num++;
+        start = end + 1;
+        end = f_str.find(',', start);
+    }
+    substrings.push_back(f_str.substr(start)); // Add the last substring
+    substrings_num++;
+
+    for (const auto& sub : substrings) {
+        for (size_t i = 0; i < std::size(x86_locators); i++) {
+            std::string_view checked_feature = std::string_view(features_string + features_indices[i] + 1);
+
+            if (checked_feature == sub) {
+                substrings_found_num++;
+                result |= UINT64_C(1) << i;
+                break;
+            }
+        }
+    }
+
+    if(substrings_found_num != substrings_num) {
+        return 0x00;
+    }
+
     return result;
 }
 
@@ -3148,8 +3188,10 @@ int main(int argc, char **argv)
         { "beta", no_argument, &sApp->requested_quality, 0 },
         { "cpuset", required_argument, nullptr, cpuset_option },
         { "disable", required_argument, nullptr, disable_option },
+        { "disable-by-feature", required_argument, nullptr, disable_by_feature_option },
         { "dump-cpu-info", no_argument, nullptr, dump_cpu_info_option },
         { "enable", required_argument, nullptr, 'e' },
+        { "enable-by-feature", required_argument, nullptr, enable_by_feature_option },
         { "fatal-errors", no_argument, nullptr, 'F'},
         { "fatal-skips", no_argument, nullptr, fatal_skips_option },
         { "fork-mode", required_argument, nullptr, 'f' },
@@ -3247,6 +3289,10 @@ int main(int argc, char **argv)
         .ignore_unknown_tests = false,
         .randomize = false,
         .cycle_through = false,
+        .enable_by_feature_req = false,
+        .enabled_features_mask = UINT64_MAX, // we set all bits during initialization, so we permit all features
+        .disable_by_feature_req = false,
+        .disabled_features_mask = 0x00, //  we set no bits during initialization, so we disable no features
     };
     const char *builtin_test_list_name = nullptr;
     int starting_test_number = 1;  // One based count for user interface, not zero based
@@ -3373,9 +3419,37 @@ int main(int argc, char **argv)
         case cpuset_option:
             apply_cpuset_param(optarg);
             break;
+        case disable_by_feature_option:
+            if(optarg == nullptr || *optarg == '\0') {
+                fprintf(stderr, "%s: --disable-by-feature requires a non-empty argument\n", argv[0]);
+                return EX_USAGE;
+            }
+
+            test_set_config.disable_by_feature_req = true;
+            test_set_config.disabled_features_mask = cpu_features_to_cpuid(optarg);
+
+            if(test_set_config.disabled_features_mask == 0x00) {
+                fprintf(stderr, "%s: --disable-by-feature: unknown CPU feature(s) requested: %s\n", argv[0], optarg);
+                return EX_USAGE;
+            }
+            break;
         case dump_cpu_info_option:
             dump_cpu_info();
             return EXIT_SUCCESS;
+        case enable_by_feature_option:
+            if(optarg == nullptr || *optarg == '\0') {
+                fprintf(stderr, "%s: --enable-by-feature requires a non-empty argument\n", argv[0]);
+                return EX_USAGE;
+            }
+
+            test_set_config.enable_by_feature_req = true;
+            test_set_config.enabled_features_mask = cpu_features_to_cpuid(optarg);
+
+            if(test_set_config.enabled_features_mask == 0x00) {
+                fprintf(stderr, "%s: --enable-by-feature: unknown CPU feature(s) requested: %s\n", argv[0], optarg);
+                return EX_USAGE;
+            }
+            break;
         case fatal_skips_option:
             sApp->fatal_skips = true;
             break;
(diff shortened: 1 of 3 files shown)


thiagomacieira (commented):
Looks interesting, but I don't see how this can actually work. I don't see where the tests are getting actually enabled.

Can you add a Bats test in bats/sanity-check/20-selftest.bats? You can use the `selftest_skip_minimum_cpu` selftest, because that one declares it requires every single CPU feature. If you enable-by-feature with any feature, it should get enabled (and skipped); if you disable-by-feature with any feature, it should get disabled and not run.

framework/sandstone_tests.h
```diff
@@ -27,14 +27,21 @@ extern struct test __stop_tests;
...
-                                   source of the tests. */
+    bool ignore_unknown_tests;          /* whether an error should be reported if
+                                           there's no test matching the specified name,
+                                           either on command-line or in a test list. */
+    bool randomize;                     /* iterate through the tests in random order. */
+    bool cycle_through;                 /* start over when an iteration through the test
+                                           set has finished. */
+    bool is_selftest;                   /* whether to use regular or selftests as the
+                                           source of the tests. */
+    bool enable_by_feature_req;         /* whether to enable tests by the requested
```
busykai, line 38: You do not need separate boolean. Instead the default for `enabled_features_mask` should be `1 << 64 - 1` (for 64 bit value). Likewise for disabled.

framework/sandstone_tests.h
```diff
@@ -95,6 +102,9 @@ class SandstoneTestSet
     std::vector<struct test_cfg_info> enable(const char *name);
     struct test_cfg_info enable(struct test *t);
 
+    std::vector<struct test_cfg_info> enable_by_feature(uint64_t enabled_features_mask);
+    std::vector<struct test_cfg_info> disable_by_feature(uint64_t disabled_features_mask);
+
```
busykai, line 107: These calls (API) are not needed. `SandstoneTestSet` is instantiated once after arguments are processed. Just pass the values from the options in `test_set_cfg` and change `enable()`/`disable()` to take it into account.

busykai (changes requested):
I think this is not the right design. We do not need to have an extra call for that, instead should include in `struct test_set_cfg`. The value of the mask (or lack of it) must be known at the moment `SandstoneTestSet` is instantiated. Then all the `enable(struct test *)`/`disable(struct test *)` should use the mask as the criterion of adding or not the test to the active test set.

dawid-sabat:
> I think this is not the right design. We do not need to have an extra call for that, instead should include in `struct test_set_cfg`. The value of the mask (or lack of it) must be known at the moment `SandstoneTestSet` is instantiated. Then all the `enable(struct test *)`/`disable(struct test *)` should use the mask as the criterion of adding or not the test to the active test set.

Well, that was my first idea to modify enable/disable methods but it looks like this had any effect only on `--enable` command line option and I think on builtin test lists.

For example when enabling all tests:

```
SandstoneTestSet::SandstoneTestSet(struct test_set_cfg cfg, unsigned int flags) : cfg(cfg), flags(flags) {
    load_all_tests(); /* initialize the catalog */
    if (!(flags & enable_all_tests)) return;
    std::span<struct test> source = !cfg.is_selftest ? regular_tests : selftests;
    for (struct test &test : source) {
        struct test_cfg_info ti;
        ti.status = test_cfg_info::enabled;
        ti.test = &test;
        test_map[test.id] = ti;
        test_set.push_back(&test);
    }
};

void SandstoneTestSet::load_all_tests()
{
    std::span<struct test> known_tests = cfg.is_selftest ? selftests : regular_tests;

    for (struct test &t : known_tests) {
        all_tests.push_back(&t);
        for (auto ptr = t.groups; ptr && *ptr; ++ptr) {
            all_group_map[(*ptr)->id].push_back(&t);
        }
    }
    /* add "special" mce_check as well */
    all_tests.push_back(&mce_test);
}
```

We call `load_all_tests` which doesn't use `enable` method and then return. Same with `add_test_list`

The bottom line is that right now `enable` isn't the only interface we add tests to the runtime list. That's the reason behind current design. If we want to modify the way the runtime test list is created that everything goes through `enable/disable` then we can put the logic in them.

I'll spend more time on this and try to do required adjustments to narrow down the interface. I'll contact @busykai if I'm out of ideas.

dawid-sabat:
This is going to be reworked.

framework/sandstone_tests.h
dawid-sabat: Fixed

framework/sandstone_tests.h
dawid-sabat: Working on it

dawid-sabat:
I reworked this PR. Most of all I ensured that method `add` is responsible for adding a test to the `test_set`.

framework/sandstone.cpp
```diff
@@ -1336,6 +1338,54 @@ static std::string cpu_features_to_string(uint64_t f)
...
 }
 
+static uint64_t cpu_features_to_cpuid(std::string_view f_str)
+{
+    auto find_feature = [](std::string_view what) -> std::optional<uint64_t> {
+        for (size_t i = 0; i < std::size(x86_locators); i++) {
+            std::string_view checked_feature = std::string_view(features_string + features_indices[i] + 1);
+
+            if (checked_feature == what) {
+                return UINT64_C(1) << i;
```
thiagomacieira, line 1348: We're likely going to 128 bits soon, so please be future-proof here. Add before the lambda:
```c++
using Int = decltype(_compilerCpuFeatures);
```
Here use `Int(1) << i`

framework/sandstone_tests.h
```diff
@@ -28,14 +28,16 @@ extern struct test __stop_tests;
...
+    bool ignore_unknown_tests;          /* whether an error should be reported if
+                                           there's no test matching the specified name,
+                                           either on command-line or in a test list. */
+    bool randomize;                     /* iterate through the tests in random order. */
+    bool cycle_through;                 /* start over when an iteration through the test
+                                           set has finished. */
+    bool is_selftest;                   /* whether to use regular or selftests as the
+                                           source of the tests. */
+    uint64_t enabled_features_mask;     /* features requested by the user. */
+    uint64_t disabled_features_mask;    /* features requested by the user. */
```
thiagomacieira, line 40: Ditto on `decltype`

framework/sandstone_tests.cpp
```diff
@@ -71,34 +71,54 @@ std::vector<struct test *> SandstoneTestSet::lookup(const char *name)
...
-};
-
-struct test_cfg_info SandstoneTestSet::add(test_cfg_info t)
+int SandstoneTestSet::add(test_cfg_info t)
 {
     struct test_cfg_info &ti = test_set.emplace_back(t);
+
+    auto it = std::remove_if(test_set.begin(), test_set.end(), [&](const test_cfg_info &ti) {
+        return ti.test == t.test;
+    });
```
thiagomacieira, line 80: This moves the matching tests to the end. Why are we doing that? Wouldn't this make
```
  -e A -e B -e A
```
execute "B, A, A" ?

Did you mean to insert this loop inside the `if` below? Maybe move the `remove_if` + `erase` operations into a lambda.

And why are we `add`ing disabled testconfigs?

framework/sandstone_tests.cpp
```diff
@@ -71,34 +71,54 @@ std::vector<struct test *> SandstoneTestSet::lookup(const char *name)
...
+    // do not add disabled tests
+    if (t.status == test_cfg_info::disabled) {
+        test_set.erase(it, test_set.end());
+        return 0;
+    }
+
+    // enable_by_feature() is "disable unless feature": if the test does not require
+    // at least one of the features in the mask, we disable it
+    // disable_by_feature() disables all tests that requires listed features.
+    uint64_t t_f_req = (t.test->minimum_cpu | t.test->compiler_minimum_cpu);
```
thiagomacieira, line 91: Use `auto` instead of `uint64_t` here, for the same future-proofing.

framework/sandstone.cpp
```diff
@@ -3177,6 +3229,8 @@ int main(int argc, char **argv)
         .ignore_unknown_tests = false,
         .randomize = false,
         .cycle_through = false,
+        .enabled_features_mask = UINT64_MAX, // we set all bits during initialization, so we permit all features
```
thiagomacieira, line 3232: `~decltype(_compilerCpuFeatures)(0)`

thiagomacieira (changes requested):
Need at least one bats test for the enabling and one for disabling. You can use `--selftests -e selftest_skip_minimum_cpu` because that test needs all feature bits, so if you pass *any* feature, it will match: it's the only test that should get enabled if you pass `--enable-by-feature` and it's the only test that gets disabled if you disable by feature. For the latter case, you need to do something like `-e selftest_skip -e selftest_skip_minimum_cpu` and verify that the total test list count was 1.

framework/sandstone.cpp
thiagomacieira: Or modify framework/scripts/x86simd_generate.pl  to emit a `typedef` with the type we've used (remember the output is also used by C code).

dawid-sabat: I like more modifying `x86simid_generate.pl`

framework/sandstone_tests.h
dawid-sabat: Done by changes in `x86simid_generate.pl`

framework/sandstone_tests.cpp
dawid-sabat: I removed confusing part of the code. I meant not to add tests that are already disabled. 

I took the lambda from `remove` method. The idea here is to remove for the `test_set` all tests that match `disable_by_feature` or not match `enable_by_feature`

framework/sandstone.cpp
dawid-sabat: Done

dawid-sabat:
Bats done.

dawid-sabat:
I doesn't seem we need this. Closing this PR to clean up.

PR #531, opened 2024-07-18
Remove trailing spaces.

dawid-sabat:
This commit is a simple celan up of the sources by removing spotted trailing spaces.


busykai (approved):
@dawid-sabat could you add this check to the CI sanity check? (In a separate PR.)

PR #532, opened 2024-07-19
Fix enable/disable logic.

dawid-sabat:
We need to check if the test is enabled or disabled before any modification of the test list.

### DIFF
--- a/framework/sandstone_tests.cpp
+++ b/framework/sandstone_tests.cpp
@@ -71,19 +71,6 @@ std::vector<struct test *> SandstoneTestSet::lookup(const char *name)
     return res;
 }
 
-SandstoneTestSet::SandstoneTestSet(struct test_set_cfg cfg, unsigned int flags) : cfg(cfg), flags(flags) {
-    load_all_tests(); /* initialize the catalog */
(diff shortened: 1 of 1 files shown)


dawid-sabat:
This PR is related to review from #529. If we want to move all [enable|disable]_by_feature logic to enable and disable methods, we need to make sure first that we use their results while creating the test set.

dawid-sabat:
That's great news! I'll read the code and close this PR as not needed anymore.

dawid-sabat:
I examined the recent change and sadly this is not the case:
Thiago's commit https://github.com/opendcdiag/opendcdiag/commit/432f0e8465a059fc31769e1ceac32c9d7a8c7af1 changes some names but it doesn't change how the interface works.

Right now `add/remove` methods apply only to `--enable` and `--disable`

If the framework builds the runtime test set from built-in test list, or from provided test list file, or from available test content it doesn't use add/remove methods.

dawid-sabat:
Oh, I totally missed this change: https://github.com/opendcdiag/opendcdiag/commit/abfc013a0f08ffe860d6eb335d25d21d95f453fb

Closing this for now, until I'll do my homework.

PR #535, opened 2024-07-24
forkfd: Detect when waitid() returns with a reason not SIGCHLD

thiagomacieira:
This can happen on Linux if we are ptracing the process, because ptrace(2) uses the wait() mechanism to get new events from the traced process. I don't know if anyone is using forkfd() to debug from the parent (it should be possible), but I expect that they use the wait functions directly instead of forkfd_wait() for the debugging work.

The problem is when something else traces our child processes. When a buggy debugger detaches from our process, waitid returns without setting si_signo (probably because SIGCHLD was never delivered). strace shows:
```
 waitid(P_PIDFD, 19, {}, WNOHANG|WEXITED, {...}) = 0
```
And the child process never changes state.

### DIFF
--- a/framework/forkfd/forkfd_linux.c
+++ b/framework/forkfd/forkfd_linux.c
@@ -220,11 +220,17 @@ int system_forkfd_wait(int ffd, struct forkfd_info *info, int ffdoptions, struct
             options |= WNOHANG;
     }
 
-    si.si_status = si.si_code = 0;
+    si.si_signo = si.si_status = si.si_code = 0;
     ret = sys_waitid(P_PIDFD, ffd, &si, options, rusage);
     if (info) {
         info->code = si.si_code;
         info->status = si.si_status;
+    }
+    if (ret == 0 && si.si_signo != SIGCHLD) {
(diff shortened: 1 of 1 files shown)


AthenasJimenez (dismissed):
LGTM

framework/forkfd/forkfd_linux.c
```diff
@@ -220,8 +220,12 @@ int system_forkfd_wait(int ffd, struct forkfd_info *info, int ffdoptions, struct
             options |= WNOHANG;
     }
 
-    si.si_status = si.si_code = 0;
+    si.si_signo = si.si_status = si.si_code = 0;
     ret = sys_waitid(P_PIDFD, ffd, &si, options, rusage);
+    if (si.si_signo != SIGCHLD) {
```
thiagomacieira, line 225: I think I need to check for `ret == 0` here.

thiagomacieira: Fixed.

AthenasJimenez (approved):

PR #536, opened 2024-07-27
FW: ensure we handle tests that log_xxx() but EXIT_SUCCESS in their init()/run()/cleanup()

thiagomacieira:
For more complex tests, it's possible that some nested function sets the state but fails to notify the outer functions that they did. If the callback function returned with `EXIT_SUCCESS`, we may end up not noticing the state had changed. We already handled some cases, but not all of them:

| Function \ Stage | init | run | cleanup |
| --- | --- | --- | --- |
| `log_skip()` | ✘ | ✔* | ✘ |
| `log_error()` | ✔ | ✔ | ✘ |
| `report_fail_msg()` / `memcmp_or_fail()` | N/A | ✔ | N/A |

(*) Probably didn't work if running in no-output logging mode.

This PR fixes the issue by storing whether the test has skipped, which is done by using a negative `fail_time`. It also changes the non-fail case to store zero there to simplify the cleaning up of the shared memory area between tests.

Drive-by fix running the cleanup in the case of strict runtime. Some tests perform testing during clean up, so it's not optional.

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -927,10 +927,10 @@ static uintptr_t thread_runner(int thread_number)
         // let SIGQUIT handler know we're done
         ThreadState new_state = thread_failed;
         if (!this_thread->has_failed()) {
-            if (ret == EXIT_SUCCESS)
-                new_state = thread_succeeded;
-            else if (ret < EXIT_SUCCESS)
+            if (this_thread->has_skipped() || ret < EXIT_SUCCESS)
                 new_state = thread_skipped;
+            else if (ret == EXIT_SUCCESS)
+                new_state = thread_succeeded;
         }
         this_thread->thread_state.store(new_state, std::memory_order_relaxed);
 
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -1672,16 +1672,29 @@ static TestResult child_run(/*nonconst*/ struct test *test, int child_number)
             pthread_create(&init_thread, nullptr, init_thread_runner, test);
             pthread_join(init_thread, &retptr);
             ret = intptr_t(retptr);
+            if (ret > 0) {
+                state = TestResult::Failed;
+            } else if (ret < 0) {
+                state = TestResult::Skipped;
+            } else {
+                // check if the thread has failed with log_error() or skipped
+                // with log_skip()
+                PerThreadData::Main *thr = sApp->main_thread_data();
+                if (thr->has_skipped())
+                    state = TestResult::Skipped;
+                else if (thr->has_failed())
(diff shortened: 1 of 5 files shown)


thiagomacieira:
ping @dawid-sabat @busykai

framework/selftest.cpp
```diff
@@ -234,13 +234,13 @@ static int selftest_logerror_init(struct test *test)
...
     log_skip(SelftestSkipCategory, "SUCCESS after skipmsg from cleanup");
-    return EXIT_SKIP;
+    return EXIT_SUCCESS;
 }
 
 static int selftest_skipmsg_skip_cleanup(struct test *test)
 {
     log_skip(SelftestSkipCategory, "SKIP after skipmsg from cleanup");
-    return EXIT_SKIP;
+    return EXIT_SUCCESS;
```
busykai, line 243: Doesn't it beat the purpose of the test (at least the message in `log_skip()`?

thiagomacieira: `log_skip` should mark it as having skipped, regardless of what you return, especially because you may forget or not all code paths may get it. That's the purpose of this change.

busykai: I am sorry, I was making the comment on the run. What I mean is the test declares to test `EXIT_SKIP` after `log_skip()` (in the func signature, in the `log_skip()` message too). With this change it changes the nature of the test completely.

thiagomacieira: Sorry, not following.

If the test fails after having skipped, then it's a fail. But if it says it skipped, it can't change its mind and be a success.

busykai: My question is simpler:

`selftest_skipmsg_success_cleanup` returns `EXIT_SUCCESS`.

`selftest_skipmsg_skip_cleanup` returns `EXIT_SUCESS`.

Why? To be fair, they were both returning `EXIT_SKIP` before. Why different functions and different skip messages if they are doing the same thing?

framework/sandstone.cpp
```diff
@@ -1667,39 +1667,47 @@ static TestResult child_run(/*nonconst*/ struct test *test, int child_number)
...
+        } else if (state == TestResult::Skipped) {
+            if (ret != 0 && ret != EXIT_SKIP)
                 log_skip(RuntimeSkipCategory, "Unexpected OS error: %s", strerror(-ret));
             state = TestResult::Skipped;
             break;
         }
 
         run_threads(test);
 
-        if (sApp->shmem->use_strict_runtime && wallclock_deadline_has_expired(sApp->endtime)){
```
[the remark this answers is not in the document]
thiagomacieira: Yes: we **must always** run the clean up function.

PR #541, opened 2024-07-30
FW/KVM: reset the SS descriptor before KVM entry

thiagomacieira:
Sometimes it is invalid right after creation.


dawid-sabat (approved):

thiagomacieira:
For the record, this is the issue: https://lore.kernel.org/kvm/Zq0EOf--E30QNyhy@google.com/T/#t

PR #542, opened 2024-08-07
FW/KVM: Work around the KVM/VMX cache bug (attempt #2)

thiagomacieira:
We previously tried to simply ensure that sregs.ss was set and not passed through from whatever the kernel gave us upon KVM_CREATE_VCPU. That has not sufficed: we are still getting KVM_EXIT_ENTRY_FAILs with an invalid SS descriptor state. So instead of attempting to work around the cache before KVM_RUN, let's just detect the symptom and force the vCPU to be recreated.

See https://lore.kernel.org/kvm/20240725175232.337266-1-mlevitsk@redhat.com/

### DIFF
--- a/framework/sysdeps/linux/kvm.c
+++ b/framework/sysdeps/linux/kvm.c
@@ -20,6 +20,9 @@
 #include <asm/kvm.h>
 #include <asm/processor-flags.h>
 #include <linux/kvm.h>
+#ifdef __x86_64__
+#  include <asm/vmx.h>
+#endif
 
 #include "sandstone_kvm.h"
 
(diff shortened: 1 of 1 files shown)


eamartin96 (approved):
LGTM

thiagomacieira:
The workaround in #541 may not have worked because there are multiple races. In https://lore.kernel.org/kvm/ZrF55uIvX2rcHtSW@chao-email/:

> vmx_segment_cache_clear() is called in a few other sites. I think at least the
> call in __vmx_set_segment() should be fixed, because QEMU may read SS.AR right
> after a write to it. if the write was preempted after the cache was cleared but
> before the new value being written into VMCS, QEMU would find that SS.AR held a
> stale value.

PR #550, opened 2024-09-25
Basic build configuration for fuzzing with AFL++

lukasz-tuz:
Commit adds simple infrastructure for fuzzing the framework's functions with AFL++. New build options are introduced:

- `fuzzing`: enables the fuzzing target build (default false)
- `afl_inc`: absolute path to directory containing AFL++ headers
- `coverage`: enables building with coverage compiler options

For convinience, `build-fuzzing.sh` script wraps meson/ninja commands for building the fuzzing target:

```bash
./scripts/build-fuzzing.sh <builddir> -Dafl_inc=/absolute/path/to/afl/include
```

`afl_inc` argument is optional, meson will attempt to locate the header in typical locations if `-Dafl_inc` is not passed.

Resulting binary contains a small set of fuzzing tests grouped in the `fuzzing` test group. Fuzzing can be started using standard AFL++ command line options, e.g.:

```bash
afl-fuzz -i <inputs> -o <outputs> -- <builddir>/opendcdiag --beta -e fuzzer_memcmp_or_fail @@
```

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,0 @@
-{
-    "C_Cpp.default.compileCommands": "/mnt/home/ltuz/dev/fork-opendcdiag/builddir/compile_commands.json",
-    "C_Cpp.default.configurationProvider": "mesonbuild.mesonbuild"
-}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,4 @@
+{
+    "C_Cpp.default.compileCommands": "/mnt/home/ltuz/dev/fork-opendcdiag/builddir/compile_commands.json",
+    "C_Cpp.default.configurationProvider": "mesonbuild.mesonbuild"
+}
--- a/meson.build
+++ b/meson.build
@@ -4,12 +4,12 @@
 project(
     'OpenDCDiag',
     ['c', 'cpp'],
-    default_options: [
+    default_options : [
         'b_ndebug=if-release',
         'c_std=gnu17',
         'cpp_std=gnu++20',
     ],
-    meson_version: '>=0.57',
+    meson_version : '>=0.57',
 )
 
 cc = meson.get_compiler('c')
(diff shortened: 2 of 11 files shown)


meson.build
```diff
@@ -4,12 +4,12 @@
 project(
     'OpenDCDiag',
     ['c', 'cpp'],
-    default_options : [
+    default_options: [
```
thiagomacieira, line 7: Clean ups are correct but noise here. Can you submit a separate patch for them?

meson.build
```diff
@@ -264,33 +282,53 @@ executable(
...
+    native: true,
 )
 
-summary({
+if get_option('fuzzing') == true
+    executable(
+        'fuzzing',
+        install: false,
+        dependencies: target_deps,
+        link_with: [
```
thiagomacieira, line 308: Why a separate executable, instead of the regular one?

tests/fuzzer/fuzzer.c
```diff
@@ -0,0 +1,70 @@
+/**
```
thiagomacieira, line 1: Should this test live inside framework, like the selftests and MCE check?

thiagomacieira (changes requested):

.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
+    "C_Cpp.default.compileCommands": "/mnt/home/ltuz/dev/fork-opendcdiag/builddir/compile_commands.json",
+    "C_Cpp.default.configurationProvider": "mesonbuild.mesonbuild"
+}
```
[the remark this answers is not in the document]
lukasz-tuz: Yes, definitely `.vscode` needs to be added to `.gitignore`.

meson.build
lukasz-tuz: Yes, you're right, I need to refrain from auto-linting these files.

meson.build
lukasz-tuz: Seemed like a good idea at the time. Using the regular one makes it simpler though, so I'll remove the separate build target.

tests/fuzzer/fuzzer.c
lukasz-tuz: You're right, it is better suited inside framework.

thiagomacieira (approved):
Thanks. This looks good.

But I wonder if we should add it to the selftests instead of the regular test roster instead.

lukasz-tuz:
> But I wonder if we should add it to the selftests instead of the regular test roster instead.

I'll look into it and submit a separate PR.

PR #551, opened 2024-09-25
IFS: don't load the next image if we are retesting a failure

thiagomacieira:
Because if we load the next image, then we aren't retesting the same thing. We'd be testing the next image.

Logging with only `log_debug` to avoid creating unnecessary noise in the log output. We already know it's a retest.


AthenasJimenez (approved):
LGTM

AthenasJimenez (approved):

PR #554, opened 2024-10-03
FW: refactor cmdline options parsing; use boost::program_options; add unittests

mcisowsk:
This PR intends to organize cmdline parsing, by usage of boost::program_options. The library does type checking for us, we can also easily check for contradictory arguments. The intention was to get rid of cmdline options order dependency and unexpected behaviourr. Moreover, putting all the cmdline-parsing-related code in sandstone_opts.hpp/cpp files allowed for relatively easy integration into unit-tests.


busykai:
Any good reason to switch to `boost::program_options`? In other words, what does it bring compared to what we have that enticed you to switch to it?

mcisowsk:
first of all - args order of appearance in cmdline finally does not matter. I've heard that some people are confused about that and the 'correct' order of cmdline args is a tribal knowledge. Secondly, this is widely used framework for options parsing - does type checking, etc. Another goal was to encapsulate all options-parsing-related code into a single function, so that it could be easily tested with unittests. So far we do not have such tests for cmdline.

lukasz-tuz (commented):
+1 from me, I'll wait for others' comments before approving, though.

busykai:
> first of all - args order of appearance in cmdline finally does not matter. I've heard that some people are confused about that and the 'correct' order of cmdline args is a tribal knowledge. Now, the order of parsing is fixed, as we do not iterate over the options provided. We can set our own constraints about what should be parsed before something else, like for example 'selftests' before 'list-group-members', 'cpuset' before 'dump-cpu-info', quality options before test listing options.

This does not answer my question why boost is needed for that. Because it's not. This is a breaking change. While I conceptually agree that the order of the arguments should not have mattered, many customers may rely on this (unconsciously so) in their scripts.

> Secondly, this is widely used framework for options parsing - does type checking, validates presence of the argument's value, provides API to create --help output automatically, based on the list of defined options.

`getopt(3)` is also a widely used framework for option parsing. Something is "widely used" is a non-argument for introducing a change. In this form the PR would break the build (in anything other than the build containers) since it does not require boost, actually. Additionally, it is now required that boost is installed in order to run `opendcdiag` since it's going to be a dependency on another .so. I do not see this as a reasonable requirement to add just because something is a "widely used".

> Another goal was to encapsulate all options-parsing-related code into a single function (and separate file), so that it could be easily tested with unittests. So far we do not have such tests for cmdline.

I absolutely wholeheartedly agree with this goal. Why we don't we start there based on the existing code and there without breaking anything and then add up on that.

busykai (changes requested):
I do not believe adding boost as 1. a coding framework and 2. a runtime dependency adds any value to the project. Please reconsider the approach. I suggest that this PR focuses on moving option parsing into a separate compilation unit and adds tests that such separation will allow.

busykai:
Closing this one. This is not a desirable change, superseded by #558.

PR #555, opened 2024-10-04
CI: Install full boost package

mcisowsk:
Especially when we advise to do so in README

libboost-dev contains only headers, non-header-only packages are in libboost-all-dev


lukasz-tuz (approved):
+1

PR #556, opened 2024-10-09
Add message when running beta tests without `--beta` option

eamartin96:
This change removes the iteration condition for tests with quality lower that requested and add it before calling `run_one_test`. Also, a warning is logged for BETA tests.

Example:
For a test with `TEST_QUALITY_PROD`
```
$ ./builddir/opendcdiag -e smi_count
command-line: 'opendcdiag -e smi_count'
version: opendcdiag-7df166cfa4e4-dirty
exit: pass
```

For a test with `TEST_QUALITY_BETA`
```
$ ./builddir/opendcdiag -e simple_add
command-line: 'opendcdiag -e simple_add'
version: opendcdiag-7df166cfa4e4-dirty
Test simple_add is in BETA quality. Try again with --beta option.
exit: pass
```

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -2160,7 +2160,11 @@ static TestResult run_one_test_once(const struct test *test)
         (void) missing;
 
         children.results.emplace_back(ChildExitStatus{ TestResult::Skipped });
-    } else {
+    } else if (test->quality_level >= 0 && test->quality_level < sApp->requested_quality) {
+        init_per_thread_data();
+        log_skip(TestResourceIssueSkipCategory, "Test %s is in BETA quality, try again using --beta option", test->id);
+        children.results.emplace_back(ChildExitStatus{ TestResult::Skipped });
+    } else if (test->quality_level >= sApp->requested_quality) {
         run_one_test_children(children, test);
     }
 
@@ -2556,7 +2560,7 @@ static SandstoneTestSet::EnabledTestList::iterator get_first_test()
 {
     logging_print_iteration_start();
     auto it = test_set->begin();
-    while (it != test_set->end() && it->test->quality_level < sApp->requested_quality)
+    while (it != test_set->end() && (it->test->quality_level < 0 && sApp->requested_quality >= 0))
         ++it;
     return it;
 }
@@ -2569,7 +2573,7 @@ get_next_test(SandstoneTestSet::EnabledTestList::iterator next_test)
         return test_set->end();
 
     ++next_test;
-    while (next_test != test_set->end() && next_test->test->quality_level < sApp->requested_quality)
+    while (next_test != test_set->end() && (next_test->test->quality_level < 0 && sApp->requested_quality >= 0))
         ++next_test;
     if (next_test == test_set->end()) {
         if (should_start_next_iteration()) {


framework/sandstone.cpp
```diff
@@ -3762,7 +3758,10 @@ int main(int argc, char **argv)
             }
         }
 
-        lastTestResult = run_one_test(*it, per_cpu_failures);
+        if (it->test->quality_level >= sApp->requested_quality)
+            lastTestResult = run_one_test(*it, per_cpu_failures);
+        else
+            logging_printf(LOG_LEVEL_QUIET, "Test %s is in BETA quality. Try again with --beta option.\n", it->test->id);
```
thiagomacieira, line 3764: Printing it like this will break the YAML output. Instead, move the checking into `run_one_test` and do a `log_skip()` like we do for the CPU feature limit.

thiagomacieira (changes requested):

framework/sandstone.cpp
thiagomacieira: The CI caught this issue.

thiagomacieira:
A couple of other suggestions:
* don't print anything for tests with quality below 0. The CI should never show anything about test "selftest_pass_low_quality"
* we're going to need to have a test for this. As `--selftests` sets the quality down to zero, you'll need to run something like `--quality 100 -e selftest_pass` and that should cause it to print "skip"

framework/sandstone.cpp
```diff
@@ -3762,7 +3758,10 @@ int main(int argc, char **argv)
             }
         }
 
-        lastTestResult = run_one_test(*it, per_cpu_failures);
+        if (it->test->quality_level >= sApp->requested_quality)
```
busykai, line 3761: This check does not belong in `main()`. It belongs to where the check was originally performed, inside `get_(first
|next)_test()` functions. You can opt to reorganize the code so that it checks and outputs there. Maybe it is even worth unifying the two functions.

framework/sandstone.cpp
busykai: It also should not be `LOG_LEVEL_QUIET`. At least on one `-v`.

busykai:
I would actually suggest doing it in a different way. We should make this part of standard log file (not output) when a test was requested but not run, in the same way it's done when a feature is missing, it should be formally skipped (with the corresponding output) as opposed to generating non-standard output or warnings.

thiagomacieira:
> I would actually suggest doing it in a different way. We should make this part of standard log file (not output) when a test was requested but not run, in the same way it's done when a feature is missing, it should be formally skipped (with the corresponding output) as opposed to generating non-standard output or warnings.

That's what I suggested to edgar in a teams chat. He has implemented it but apparently hasn't pushed yet. There was something wrong that the test was marked as "fail" instead of "skip" that he was debugging when we last talked.

framework/sandstone.cpp
```diff
@@ -2160,8 +2160,15 @@ static TestResult run_one_test_once(const struct test *test)
...
-    } else {
+    } else if (test->quality_level >= 0 && test->quality_level < sApp->requested_quality) {
+        init_per_thread_data();
+        log_skip(TestResourceIssueSkipCategory, "Test %s is in BETA quality, try again using --beta option", test->id);
+        children.results.emplace_back(ChildExitStatus{ TestResult::Skipped });
+    } else if (test->quality_level >= sApp->requested_quality) {
         run_one_test_children(children, test);
+    } else {
+        init_per_thread_data();
+        children.results.emplace_back(ChildExitStatus{ TestResult::Skipped });
```
thiagomacieira, line 2171: Do we still reach this at any time? The `get_first_test` and `next_text` functions below will skip the test if the quality is negative and the requested quality non-negative.

thiagomacieira (approved):

PR #557, opened 2024-10-10
Splitting and renaming CPU-related code

lukasz-tuz:
This PR proposes further separation of CPU-specific code from the rest of the framework. Specifically:
- devicedeps/ directory restructured to put device type first, OS interfaces and other code second
- some of the variables are renamed to become more generic or more device-specific, where it makes sense
- added build-time flag to indicate selected type of device (currently only CPU)


busykai:
Why this is arranged as a single commit? Could you split it into multiple (and reviewable) commits?

docs/device_support.md
```diff
@@ -0,0 +1,126 @@
...
+
+- `cpu`, default; build OpenDCDiag for CPU devices
+
+Both framework and tests require certain information about hardware they run on.
+Commonly needed data include:
+
+- number of devices; for scheduling and memory allocation
+- topology of the devices: which logical devices (execution units) belong to which
+  physical device (e.g., which cores/logical CPUs belong to a CPU package)
+- cache/memory sizes and structure
```
thiagomacieira, line 23: Do we need this for non-CPU devices?

framework/logging.cpp
```diff
@@ -1788,7 +1791,7 @@ void KeyValuePairLogger::print(int tc)
         logging_printf(LOG_LEVEL_VERBOSE(1), "%s_random_generator_state = %s\n", test->id,
```
thiagomacieira, line 1791: I'd completely disable all but the YAML logger for a non-CPU device. That way, you don't have to worry about porting them.

thiagomacieira (approved):
Minor nitpicks.

PR #558, opened 2024-10-11
FW: move cmdline parsing to separate file

mcisowsk:


framework/sandstone_opts.hpp
```diff
@@ -0,0 +1,36 @@
...
+ * Copyright 2024 Intel Corporation.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef SANDSTONE_OPTS
+#define SANDSTONE_OPTS
+
+#include "sandstone_tests.h"
+
+struct ParsedOpts {
```
busykai, line 11: Maybe `CommandLineOpts` would be a better name for this struct?

framework/sandstone.cpp
```diff
@@ -3208,418 +2671,12 @@ int main(int argc, char **argv)
...
-        sApp->delay_between_tests = 50ms;
-        sApp->thermal_throttle_temp = INT_MIN;
-        fatal_errors = true;
-        builtin_test_list_name = "auto";
-
-        static_assert(!SandstoneConfig::RestrictedCommandLine || SandstoneConfig::HasBuiltinTestList,
-                "Restricted command-line build must have a built-in test list");
+    ParsedOpts opts;
+    auto ret = parse_cmdline(argc, argv, sApp, opts);
+    if (ret == EXIT_SUCCESS || ret == EX_USAGE) {
```
busykai, line 2676: I think that now we have a chance to improve on separation of parsing vs. acting. What if we introduced "commands" concept that would indicate an action that is to be taken (e.g. "list tests") as opposed to doing it in the place where we're parsing them. That is, after `parse_cmdline()` returns, `ParsedOpts` would contain specific single command that the subsequent code needs to execute and then `switch (opts.command) { case HELP: ...; break; case LIST_TESTS: ...; break; etc.}` with the default command being `RUN` which would actually proceed with the execution.

The same applies to changes to the configuration that need to be taken (e.g. applying different topology/cpuset). Ideally, none of that should be happening inside `sandstone_opts.cpp`, but here, in `main()`...

busykai (commented):
This is good change that gets rid of a lot of command-line option clatter we have in `sandstone.cpp`, but it could be further improved upon to keep the separation between parsing and performing an action cleaner.

framework/sandstone.cpp
mcisowsk: this particular PR is intended to only separate the code. I have a followup which refactors the commandline parsing completely, and also does this 'parsing vs acting' separation. So if you don't mind, I would not perform any major changes to the parsing itself in this PR.

framework/sandstone_opts.hpp
mcisowsk: `CommandLineOpts` sounds a bit raw, while they are fully parsed already.

framework/sandstone.cpp
busykai: I do mind. The reason I am mentioning this is that you change API and passing around sApp to accomplish the separation. What I am asking for is `enum Command` and a field in the options struct, then you can keep the action code in `sandstone.cpp`.

framework/sandstone_opts.cpp
```diff
@@ -0,0 +1,846 @@
...
+        { "is-debug-build", no_argument, nullptr, is_debug_option },
+        { "test-tests", no_argument, nullptr, test_tests_option },
+#endif
+        { nullptr, 0, nullptr, 0 }
+    };
+
+    int opt;
+    int coptind = -1;
+
+    while (!SandstoneConfig::RestrictedCommandLine &&
```
thiagomacieira, line 416: This works, but we lose the Dead Code Elimination that this produced in sandstone.cpp: the compiler knows `SandstoneConfig::RestrictedCommandLine` is a constant, so it knew whether the parsing was one was or another, and that some of the variables could only assume certain specific values.

thiagomacieira (commented):
Please write a motivation in the commit message.

framework/sandstone_opts.cpp
busykai: Why don't we put it in an `if` instead. And the `if (SandstoneConfig::RestrictedCommandLine)` below will become the `else`. That will work.

framework/sandstone_opts.cpp
```diff
@@ -0,0 +1,846 @@
...
+    }
+    return EXIT_SUCCESS;
+}
+
+void warn_deprecated_opt(const char *opt)
+{
+    fprintf(stderr, "%s: option '%s' is ignored and will be removed in a future version.\n",
+            program_invocation_name, opt);
+}
+}
```
busykai, line 323: Please put a comment here `} /* anonymous namespace */`. (I know it's a copy-paste, but while you're at it.)

busykai (commented):
This looks good to me. We need to address the restricted command-line scenario.

framework/sandstone_opts.cpp
```diff
@@ -0,0 +1,845 @@
...
+        if (SandstoneConfig::NoLogging) {
+            app->shmem->output_format = SandstoneApplication::OutputFormat::no_output;
+        } else  {
+            app->shmem->verbosity = 1;
+        }
+
+        app->delay_between_tests = 50ms;
+        app->thermal_throttle_temp = INT_MIN;
+        opts.fatal_errors = true;
+        opts.builtin_test_list_name = "auto";
```
thiagomacieira, line 833: Question: should we move the test file processing here, from sandstone.cpp?

That would remove the need to have `builtin_test_list_name`, `enable_tests` and such in `opts`.

framework/sandstone.cpp
```diff
@@ -3209,418 +2790,38 @@ int main(int argc, char **argv)
...
+        test_set = new SandstoneTestSet(opts.test_set_config, SandstoneTestSet::enable_all_tests);
+        list_group_members(opts.list_group_name);
+        return EXIT_SUCCESS;
+    case Action::version:
+        logging_print_version();
+        return EXIT_SUCCESS;
+    case Action::exit:
+        return EXIT_SUCCESS;
+    case Action::run:
+    default:
```
thiagomacieira, line 2821: Don't add `default` cases to exhaustive `switch` on  an enum.

framework/sandstone.cpp
```diff
@@ -3630,59 +2831,59 @@ int main(int argc, char **argv)
     if (sApp->total_retest_count < -1 || sApp->retest_count == 0)
         sApp->total_retest_count = 10 * sApp->retest_count; // by default, 100
 
-    if (unsigned(thread_count) < unsigned(sApp->thread_count))
-        restrict_topology({ 0, thread_count });
-    slice_plan_init(max_cores_per_slice);
+    if (unsigned(opts.thread_count) < unsigned(sApp->thread_count))
+        restrict_topology({ 0, opts.thread_count });
+    slice_plan_init(opts.max_cores_per_slice);
```
thiagomacieira, line 2836: By moving the options into `opts`, now the compiler can't perform dead code elimination based on the values of `thread_count`, `max_cores_per_slice`, etc.

For the cases where we call out-of-line functions (like `debug_init_global` below), there's little code to be eliminated. For everything that gets used here in `main()`, please implement like this:
* rename the variable in `SandstoneOpts` to be `m_varname` or `varname_`
* add a function with the front-end name that is like:
```c++
    int thread_count() const { if (SandstoneOptions::RestrictedCommandLine) return -1; return thread_count_; }
```

framework/sandstone.cpp
```diff
@@ -3630,59 +2831,59 @@ int main(int argc, char **argv)
...
+            if (!tis.size() && !opts.test_set_config.ignore_unknown_tests) {
                 fprintf(stderr, "%s: Cannot find matching tests for '%s'\n", program_invocation_name, name);
                 exit(EX_USAGE);
             }
         }
     }
 
     /* Add the test list file */
-    if (test_list_file_path) {
+    if (opts.test_list_file_path) {
```
thiagomacieira, line 2872: Should all of this move to sandstone_opts.cpp?

thiagomacieira (changes requested):
Looks good, but possibly incomplete. If you can move the test processing, do it too on a separate commit in the same PR.

framework/sandstone_opts.cpp
mcisowsk: this is more of a processing of the parsed opts, so I don't think this is a good idea. We could isolate it into a separate function though. `main` would then become even more concise. But this is not a scope of this PR.

framework/sandstone.cpp
mcisowsk: Question: why not `if constexpr`?

framework/sandstone_opts.cpp
```diff
@@ -0,0 +1,848 @@
...
+            else if (v > max || (erange && parsed == std::numeric_limits<long long>::max()))
+                v = Integer(max);
+        }
+        return v;
+    }
+};
+
+auto parse_testrun_range(const char *arg)
+{
+    int starting_test_number = 1; // One based count for user interface, not zero based
```
mcisowsk, line 294: note that I resigned from passing `starting_test_number` and `ending_test_number` by reference - they are only used in the scope of this function, so that seemed redundant

framework/sandstone_opts.cpp
```diff
@@ -0,0 +1,911 @@
...
+                break;
+            case retest_on_failure_option:
+                app->retest_count = ParseIntArgument<>{
+                        .name = "--retest-on-failure",
+                        .max = SandstoneApplication::MaxRetestCount,
+                        .range_mode = OutOfRangeMode::Saturate
+                }();
+                break;
+            case reschedule_option:
+                if (app->thread_count < 1) {
```
busykai, line 662: This is a bug, actually, it needs to be `< 2`. I know it wasn't introduced by this change, but please fix it.

busykai (dismissed):
I think this is very good change. I want to see it in, as well as the follow-up #563.

framework/sandstone_opts.cpp
thiagomacieira: Yeah, this is dead code. Thanks for copying it, but can you delete in a later commit saying it is dead code? You'll remove the command-line option, so that requires a `[ChangeLog][Framework]` note.

thiagomacieira (dismissed):

framework/sandstone_opts.cpp
mcisowsk: ok, in a follow-up

thiagomacieira (approved):

PR #562, opened 2024-10-15
FW: reorganize current_max_loop_count assignment

mcisowsk:
We want to give `.fracture_loop_count` value  priority over cmdline,
so that it won't execute more loops than that.

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -2339,19 +2339,20 @@ run_one_test(const test_cfg_info &test_cfg, SandstoneApplication::PerCpuFailures
     sApp->current_test_duration = test_duration(test_cfg);
     first_iteration_target = MonotonicTimePoint::clock::now() + 10ms;
 
-    if (sApp->max_test_loop_count) {
-        sApp->shmem->current_max_loop_count = sApp->max_test_loop_count;
+    if (test->fracture_loop_count) {
+        // handled first, not overridable by --max-test-loop-count
+        sApp->shmem->current_max_loop_count = test->fracture_loop_count;
     } else if (test->desired_duration == -1) {
         sApp->shmem->current_max_loop_count = -1;
     } else if (sApp->test_tests_enabled()) {
         // don't fracture in the test-the-test mode
         sApp->shmem->current_max_loop_count = -1;
-    } else if (test->fracture_loop_count == 0) {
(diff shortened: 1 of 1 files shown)


busykai:
`fracture_loop_count` had the lowest priority, now it has the highest. why is this change?

framework/sandstone.cpp
```diff
@@ -2339,13 +2339,15 @@ run_one_test(const test_cfg_info &test_cfg, SandstoneApplication::PerCpuFailures
     sApp->current_test_duration = test_duration(test_cfg);
     first_iteration_target = MonotonicTimePoint::clock::now() + 10ms;
 
-    if (sApp->max_test_loop_count) {
-        sApp->shmem->current_max_loop_count = sApp->max_test_loop_count;
+    if (test->fracture_loop_count) {
+        sApp->shmem->current_max_loop_count = test->fracture_loop_count;
```
busykai, line 2343: The fallback `else` at the end of this construct does not make sense anymore. However, the question is why we're honoring `fracture_loop_count` over `max_test_loop_count` with this change.

framework/sandstone.cpp
```diff
@@ -2339,13 +2339,15 @@ run_one_test(const test_cfg_info &test_cfg, SandstoneApplication::PerCpuFailures
...
     } else if (test->desired_duration == -1) {
         sApp->shmem->current_max_loop_count = -1;
     } else if (sApp->test_tests_enabled()) {
         // don't fracture in the test-the-test mode
         sApp->shmem->current_max_loop_count = -1;
+    } else if (sApp->max_test_loop_count) {
+        sApp->shmem->current_max_loop_count = sApp->max_test_loop_count;
     } else if (test->fracture_loop_count == 0) {
```
jposwiata, line 2351: this should be "else" part

framework/sandstone.cpp
busykai: ok. @mcisowsk please include the reasoning in the commit message (and, optionally, in the PR description).

busykai (changes requested):
Reorganize the if/else if logic according to the change. Include the reason behind the change in the commit message.

jposwiata (approved):

jposwiata (approved):

busykai (approved):

PR #563, opened 2024-10-22
FW: implement CmdlineParser

mcisowsk:


framework/sandstone_opts.cpp
```diff
@@ -0,0 +1,1210 @@
...
+void warn_deprecated_opt(const char *opt)
+{
+    fprintf(stderr, "%s: option '%s' is ignored and will be removed in a future version.\n",
+            program_invocation_name, opt);
+}
+
+struct CmdlineParser {
+    struct ArgsPresent {
+        // endpoints
+        bool list = false;
```
busykai, line 419: It is `false` by default, it's inconsistent with how `std::optional<>` are defined. Either initialize them both explicitly or use implicit (which is what is needed here).

Edit: see the comment above, this entire structure (`ArgsPresent`) is not needed.

framework/sandstone_opts.cpp
```diff
@@ -0,0 +1,1210 @@
...
+        std::optional<std::string> max_messages_per_thread;
+        std::optional<std::string> max_test_count;
+        std::optional<std::string> max_test_loop_count;
+        std::optional<const char*> builtin_test_list_name;
+        std::optional<std::string> thermal_throttle_temp;
+        std::optional<std::string> total_retest_count;
+        std::optional<const char*> test_list_file_path;
+
+        std::optional<std::vector<const char*>> enabled_tests;
+        std::optional<std::vector<const char*>> disabled_tests;
```
busykai, line 489: These should be defined in scope of the parser itself, not here.

framework/sandstone_opts.cpp
```diff
@@ -0,0 +1,1210 @@
...
+}
+
+void warn_deprecated_opt(const char *opt)
+{
+    fprintf(stderr, "%s: option '%s' is ignored and will be removed in a future version.\n",
+            program_invocation_name, opt);
+}
+
+struct CmdlineParser {
+    struct ArgsPresent {
```
busykai, line 417: This is very heavy and convoluted (yet another place to edit if adding an option!). Just use `std::map<int, std::variant<...>>` to keep track of options parsed.

framework/sandstone_opts.cpp
```diff
@@ -0,0 +1,1210 @@
...
+        app->delay_between_tests = 50ms;
+        app->thermal_throttle_temp = INT_MIN;
+
+        static_assert(!SandstoneConfig::RestrictedCommandLine || SandstoneConfig::HasBuiltinTestList,
+                "Restricted command-line build must have a built-in test list");
+        return EXIT_SUCCESS;
+    }
+
+    ArgsPresent args_present;
+    int verbosity = -1;
```
busykai, line 1191: Why `verbosity` is discriminated?

framework/sandstone_opts.cpp
```diff
@@ -0,0 +1,1210 @@
...
+};
+} /* anonymous namespace */
+
+int ParsedCmdLineOpts::parse(int argc, char** argv, SandstoneApplication* app, ParsedCmdLineOpts& opts) {
+    // isolate CmdlineParser existance so that we create it only for the sake of cmdline parsing
+    CmdlineParser parser;
```
busykai, line 1197: I believe the entire `CmdlineParser` is not unnecessary. It's not a class. It's only state is the bookkeeping of the arguments parsed, its scope is this file only, and it's only created once in the app lifecycle. Which, to me, is a hint that it does not need to be a class. Instead use file-global `std::map` to keep track of parsed arguments and refer it in file-scope `parse/collect/validate` functions.

framework/sandstone_opts.hpp
```diff
@@ -0,0 +1,70 @@
...
+enum class Action {
+    list_tests,
+    list_group,
+    dump_cpu_info,
+    version,
+    exit,
+    run,
+};
+
+struct ParsedCmdLineOpts {
```
busykai, line 20: Please name this `ProgramOptions`.

busykai (changes requested):
It's a good clean up of `sandstone.cpp` compilation unit, but `sandstone_opts.[hc]pp` require some polishing/simplification.

framework/sandstone.cpp
```diff
@@ -3213,418 +2794,42 @@ int main(int argc, char **argv)
...
-            suggest_help(argv);
-            return EX_USAGE;
+    ParsedCmdLineOpts opts;
+    try {
+        auto ret = parse_cmdline(argc, argv, sApp, opts);
+        if (ret != EXIT_SUCCESS) {
+            return ret;
         }
+    } catch (std::invalid_argument& e) {
+        fprintf(stderr, "Error while parsing opts: %s\n", e.what());
```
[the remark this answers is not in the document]
mcisowsk: do we want to print something, so like:
```
    if (!parse_cmdline(argc, argv, sApp, opts)) {
        fprintf(stderr, "Error while parsing opts\n");
        return EX_USAGE;
    }
```
?

framework/sandstone_opts.cpp
```diff
@@ -0,0 +1,1210 @@
...
+            case triage_option:
+            case weighted_testrun_option:
+                warn_deprecated_opt(long_options[coptind].name);
+                break;
+
+            case 0:
+                /* long option setting a value */
+                continue;
+            default:
+                // suggest_help(argv);
```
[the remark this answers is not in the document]
mcisowsk: probably by mistake

framework/sandstone_opts.cpp
```diff
@@ -0,0 +1,1210 @@
...
+        if (args_present.is_asan || args_present.is_debug) {
+            opts.action = Action::exit;
+            return EXIT_SUCCESS;
+        }
+        if (args_present.list) {
+            opts.list_tests_include_descriptions = true;
+            opts.list_tests_include_tests = true;
+            opts.list_tests_include_groups = true;
+            opts.action = Action::list_tests;
+            // return EXIT_SUCCESS; // do we exit immediately or proceed further?
```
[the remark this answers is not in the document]
mcisowsk: if we return, then we have to put parsing of all args that may be needed by those actions, before them

framework/sandstone_opts.cpp
mcisowsk: the idea was to count occurrences of `-v` in the command. That's why it's kept as the Parser's object state

framework/sandstone_opts.hpp
```diff
@@ -0,0 +1,70 @@
...
+        on_hang_arg = nullptr;
+        on_crash_arg = nullptr;
+        builtin_test_list_name = "auto";
+    }
+};
+
+inline int parse_cmdline(int argc, char** argv, SandstoneApplication* app, ParsedCmdLineOpts& opts) {
+    auto ret = opts.parse(argc, argv, app, opts);
+    if constexpr (SandstoneConfig::RestrictedCommandLine) {
+        opts.apply_restrictions();
```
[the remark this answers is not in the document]
mcisowsk: it's hard to choose between Thiago's suggestion to go with `apply_restrictions()` and yours `restricted_cmdline_defaults()`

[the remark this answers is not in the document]
mcisowsk: the idea of code elimination is to put it _after_ parse

framework/sandstone_opts.cpp
mcisowsk: you're right, especially when we already have the enum of all options, thanks. Good idea

mcisowsk: I wonder, however, is it worth it to keep entire map instead of this structure. Also, acquiring map elements is 'uglier'. Using `std::get` is not the prettiest call:
```
            case 'O':
                if (!args_present.log_test_knobs) {                                          // before
                    args_present.log_test_knobs.emplace();
                }
                args_present.log_test_knobs->emplace_back(optarg);

                if (!args_present_.count('O')) {                                               // after
                    args_present_.emplace('O', std::vector<const char*>());
                }
                std::get<std::vector<const char*>>(args_present_.at('O')).emplace_back(optarg);
                break;
```

framework/sandstone_opts.cpp
busykai: Keep it as int and increment instead.

framework/sandstone_opts.cpp
busykai: well, it's C++ for you. i did not expect any grace and beauty out of the box. my point is that this duplicates all the names, unnecessarily so (we have enum + literal char). separate the functionality. you actually need to keep track of how many times you've seen the option which is `map<int, int>` and a place to store the value you've seen (where the value is needed). right now the struct aims to do both. in pseudo C++:

```c++
// file scope
namespace {
static std:map<int, int> opts_seen;
static struct {
    std::string list_group_members_name;
    const char* fork_mode;
    int verbosity = -1;
    ...
    std::vector<const char *> log_test_knobs;
    ...
} opt_vals;
}
```

So you'd do:
```
 case 'O':
    if (!opts_seen.count('O')) {
        opts_seen['O'] = 1;
    }
    opt_vals.log_test_knobs.emplace_back(optarg);
    break;
case 'v':
    opt_vals.verbosity++;
    break;
...
```

mcisowsk: yeah ok, I will work it out

framework/sandstone_opts.cpp
mcisowsk: isn't file global variable worse idea than this self contained struct, whose object will be destroyed after we return from parse()? When exactly would be file global map destroyed?

busykai: The way you're declared -- yes. Normally, file-scope variables are not allocated where declared (declared as pointers). That's fine, keep it like you have it, but please call it `ProgramOptionsParser`. Or, even better, declare it as anonymous inside `parse()` function (`class { ... } parser;`). My point is that it's not a datatype if you use it once.

framework/sandstone.cpp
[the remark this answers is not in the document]
busykai: No, this one should just return. Any error reporting should be done where the code has the context to report it properly. Right now it reports, for example:

```
$ ./opendcdiag -k
./opendcdiag: invalid option -- 'k'
Try './builddir/opendcdiag --help' for more information.
```

No need to print `Error while parsing opts` after it -- it adds no value.

framework/sandstone_opts.cpp
mcisowsk: anonymous class would be nice, if it wasn't that big

framework/sandstone_opts.cpp
```diff
@@ -933,7 +824,427 @@ int parse_cmdline(int argc, char** argv, SandstoneApplication* app, ParsedCmdLin
                 return EX_USAGE;
             }
         }
-    } else {
+        return EXIT_SUCCESS;
+    }
+
+    // validate dynamic conditions, like other conflicting arguments' presence
+    int validate_args() const
```
busykai, line 831: We had a discussion with @thiagomacieira about this. We cannot introduce breaking changes (e.g. customer scripts based on existing behavior). At most, we can warn about certain conditions. As such, argument validation cannot return `EX_USAGE` for any checks that were newly introduced. I think warning is OK, but not termination. Please make sure that the warning message is not too harsh, i.e. instead of "a and b are mutually exclusive" use "a and b were specified together, using <value>". Alternatively, we can just eliminate `validate_args()` altogether from this changeset and introduce them later.

busykai (changes requested):
This breaks current behavior (silently take the last specified) by prohibiting mutually exclusive options which is may or may not be the right thing to do, but it's too late to introduce this at this point in time. To keep the current behavior and produce meaningful warnings, some more work needs to be done to track the last one specified as well as process and expose them at the stage of parsing instead of when consuming. That is, there should be "duration" option which keeps the value of the last one specified in addition to `one_sec_option`, `five_sec_option`, and so on. Same for "alias" options, like `--no-slicing` which should also set the value of `max_core_per_slice`.

framework/sandstone_opts.cpp
```diff
@@ -445,470 +536,270 @@ class RandomDeviceSchedule : public DeviceSchedule
...
+                    opts_map.emplace(disable_option, std::vector<const char*>());
+                }
+                std::get<std::vector<const char*>>(opts_map.at(disable_option)).emplace_back(optarg);
                 break;
             case 'e':
-                opts.enabled_tests.push_back(optarg);
+                if (!opts_map.count('e')) {
+                    opts_map.emplace('e', std::vector<const char*>());
+                }
+                std::get<std::vector<const char*>>(opts_map.at('e')).emplace_back(optarg);
```
busykai, line 565: I believe the proper C++ way of doing this is to

```C++
<std::get<std::vector<const char *>>(opts_map['e']).emplace_back(optarg);
```

(here and in other places).

framework/sandstone_opts.cpp
```diff
@@ -933,7 +824,418 @@ int parse_cmdline(int argc, char** argv, SandstoneApplication* app, ParsedCmdLin
...
+            auto endtime = std::get<const char*>(opts_map.at('T'));
+            if (strcmp(endtime, "forever") == 0) {
+                app->endtime = MonotonicTimePoint::max();
+            } else {
+                app->endtime = app->starttime + string_to_millisecs(endtime);
+            }
+            opts.test_set_config.cycle_through = true; /* Time controls when the execution stops as
+                                                        opposed to the number of tests. */
+        }
+        if (opts_map.count(one_sec_option)) {
```
busykai, line 991: This is what I was referring to in my previous comments. You need to take the last specified option, not the longest duration. (Same logic applies to `--alpha` and `--beta`.) For that you'd need to have a `duration` option and set it when duration-related option (`--1sec`, `--5sec` etc. is parsed). And then here:

```C++
app->endtime = app->starttime + std::get<Duration>(opts_map.at[duration_option]);
```

busykai (changes requested):

framework/sandstone_opts.cpp
mcisowsk: for `std::variant` this trick does not work unfortunately. I need to emplace the concrete variant first. 

```sh
terminate called after throwing an instance of 'std::bad_variant_access'
  what():  std::get: wrong index for variant
Aborted
```

Alternatively, I can propose such monstrosity, if you insist on doing it with one function:
```c++
std::get<std::vector<const char*>>(opts_map.try_emplace('e', std::vector<const char*>()).first->second).emplace_back(optarg);
```

framework/sandstone_opts.cpp
mcisowsk: I see your point, but do we want to also mimic the behaviour in which:
1. first is `--1sec`, so we set 
```c++
            opts.test_set_config.randomize = true;
            opts.test_set_config.cycle_through = true;
            app->shmem->use_strict_runtime = true;
            app->endtime = app->starttime + 1s;
```

2. then goes `-T` which sets:
```c++
            app->endtime = app->starttime + string_to_millisecs(endtime);
            opts.test_set_config.cycle_through = true;
```
but *also*, it does *not unset* `opts.test_set_config.randomize` and `app->shmem->use_strict_runtime` which were set before...

To reproduce this mess I would need to basically destroy my implementation on purpose.

busykai: > but also, it does not unset opts.test_set_config.randomize and app->shmem->use_strict_runtime which were set before...

Understood, yes. We can neglect imitating the side-effects of setting specific option. However, we'd need to track what option was used to set the duration (I believe it's the only case where using an "alias" has implications of setting other options?). In other words, we'd be fine if `--1sec` followed by `-T 5m` would result in *not setting* randomization and strict runtime which is a slight deviation from previous behavior. Makes sense?

framework/sandstone_opts.cpp
busykai: Nah, that's fine. That's right -- it couldn't work with `std::variant`!

framework/sandstone_opts.cpp
```diff
@@ -445,470 +536,306 @@ class RandomDeviceSchedule : public DeviceSchedule
...
+                opts_map.emplace(version_option, true);
+                break;
+
+            // XXX For legacy reasons, we allow those conflicting opts and take only the last specified one
             case one_sec_option:
-                opts.test_set_config.randomize = true;
-                opts.test_set_config.cycle_through = true;
-                app->shmem->use_strict_runtime = true;
-                app->endtime = app->starttime + 1s;
+                opts_map.emplace(one_sec_option, true);
```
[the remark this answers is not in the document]
busykai: I do not like it since it's more than one place you'd have to edit if you change anything.

I'd much rather had hidden "duration_option" (a new one) which this code would set and then deal with its value when trying to figure. That is, extend [enum](https://github.com/opendcdiag/opendcdiag/pull/563/files#diff-1a2f5bda4995d232e63b394b8ea66ef8c3adf020c9677bcfb2e15a22125056efR23) with:

```C++
duration_option, /* synthetic */
```

This code then does:

```C++
opts_map.emplace(duration_option, one_sec_option);
```

For `'T'` (or anything with value):

```C++
opts_map.emplace(duration_option, 'T');
opts_map.emplace('T', optarg);
```

And then later:

```C++
if (opts_map.count(duration_option)) {
    switch(opts_map[duration_option] {
        case 'T':
....
        case one_sec_option:
....
}
```
What do you think?

framework/sandstone_opts.cpp
```diff
@@ -722,127 +728,121 @@ struct ProgramOptionsParser {
...
                 break;
 
             case test_delay_option:
+                opts_map.insert_or_assign(_delay_option, test_delay_option);
                 opts_map.insert_or_assign(test_delay_option, string_to_millisecs(optarg));
-                override_options({quick_run_option});
                 break;
 
             case quick_run_option:
+                opts_map.insert_or_assign(_delay_option, quick_run_option);
```
mcisowsk, line 812: this one is a bit tricky, as it interferes with both max loop count and delay

framework/sandstone_opts.cpp
```diff
@@ -933,7 +864,436 @@ int parse_cmdline(int argc, char** argv, SandstoneApplication* app, ParsedCmdLin
...
+
+        app->force_test_time = opts_map.count(force_test_time_option);
+        app->fatal_skips = opts_map.count(fatal_skips_option);
+        app->ignore_mce_errors = opts_map.count(ignore_mce_errors_option);
+        app->ignore_os_errors = opts_map.count(ignore_os_errors_option);
+        app->vary_frequency_mode = opts_map.count(vary_frequency);
+        app->vary_uncore_frequency_mode = opts_map.count(vary_uncore_frequency);
+
+        app->shmem->use_strict_runtime = opts_map.count(strict_runtime_option);
+        app->shmem->ud_on_failure = opts_map.count(ud_on_failure_option);
```
thiagomacieira, line 1071: These are fine, provided they're not producing warnings on the cast from `size_t` to bool. If they do, just add `!= 0`.

framework/sandstone_opts.cpp
```diff
@@ -933,7 +864,436 @@ int parse_cmdline(int argc, char** argv, SandstoneApplication* app, ParsedCmdLin
...
+        }
+        if (opts_map.count('o')) {
+            app->file_log_path = std::get<const char*>(opts_map.at('o'));
+        }
+        if (opts_map.count(syslog_runtime_option)) {
+            app->syslog_ident = std::get<const char*>(opts_map.at(syslog_runtime_option));
+        }
+        if (opts_map.count(use_builtin_test_list_option)) {
+            opts.builtin_test_list_name = std::get<const char*>(opts_map.at(use_builtin_test_list_option));
+        }
```
thiagomacieira, line 1099:
```c++
    auto string_option_for = [&](int option) -> const char * {
        auto it = opts_map.find(option);
        return it == opts_map.end() ? nullptr : std::get<const char *>(*it);
    };
    opts.seed = string_option_for('s');
    opts.on_crash_arg = string_option_for(on_crash_option);
    opts.on_hang_arg = string_option_for(on_hang_arg);
...
    if (const char *mode = string_option_for('f')) {
        std::string_view value = mode;
        if (value == "no" || value == "no-fork") {
```

thiagomacieira (changes requested):
The overall design looks good and I think we should merge it. I have comments on the implementation (inefficiencies, duplication, plus the unnecessary warnings).

framework/sandstone_opts.cpp
```diff
@@ -445,470 +545,301 @@ class RandomDeviceSchedule : public DeviceSchedule
...
+                opts_map.insert_or_assign(_quality_option, alpha_option);
+                opts_map.emplace(alpha_option, true);
+                break;
+            case beta_option:
+                opts_map.insert_or_assign(_quality_option, beta_option);
+                opts_map.emplace(beta_option, true);
+                break;
+            case quality_option:
+                opts_map.insert_or_assign(_quality_option, quality_option);
+                opts_map.insert_or_assign(quality_option, optarg);
```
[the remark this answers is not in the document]
mcisowsk: agree, although I would like to keep the `quality` case close to the rest of quality options, for more clarity

thiagomacieira (changes requested):
One behaviour change to fix, then minor things.

framework/sandstone_opts.cpp
```diff
@@ -446,443 +546,359 @@ class RandomDeviceSchedule : public DeviceSchedule
...
-
-            case test_tests_option:
-                app->enable_test_tests();
-                if (app->test_tests_enabled()) {
-                    // disable other options that don't make sense in this mode
-                    app->retest_count = 0;
+            case 'Y':
+                opts_map.insert_or_assign(_format_option, 'Y');
+                if (auto it = opts_map.find('Y'); it != opts_map.end()) {
+                    if (optarg) // do not allow '-Y' to reset previously set indent
```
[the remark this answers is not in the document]
mcisowsk: yes, and `-Y4 -Y2` will still work as two-space indentation. What I meant here was `-Y4 -Y` case, where we expect four-space indent. Without this `if` it would reset back to 2

framework/sandstone_opts.cpp
```diff
@@ -446,443 +546,359 @@ class RandomDeviceSchedule : public DeviceSchedule
...
+
+            case 't':
+            case test_delay_option:
+            case timeout_option:
+                opts_map.insert_or_assign(opt, string_to_millisecs(optarg));
+                break;
+
+            case cpuset_option:
+                if (!opts_map.try_emplace(cpuset_option, optarg).second) {
+                    fprintf(stderr, "cpuset defined more than once\n");
```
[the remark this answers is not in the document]
mcisowsk: for this PR, should I change this behaviour to allow overwriting?
```
mcisowska@X299-6:~/opendcdiag$ build/opendcdiag -vvv --disable=* --cpuset=t0 --cpuset=c0
cpuset defined more than once
```

framework/sandstone_opts.cpp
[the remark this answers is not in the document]
thiagomacieira: Got it!

thiagomacieira (approved):

PR #565, opened 2024-10-28
Framework cpu code separation

lukasz-tuz:
This commit isolates CPU specific code in the framework using #ifdef directives and build-time flag (`SANDSTONE_DEVICE_CPU`). Where feasible, methods/variables were renamed to device agnostic versions (e.g., `num_cpus()` -> `num_devices()`. First elements of CPU code were moved under `devicedeps/cpu`; follow-up commits will continue this move and gradually remove `#ifdef`s.

> [!Note]
> PR depends on #564 

Signed-off-by: Lukasz Tuz <lukasz.tuz@intel.com>


lukasz-tuz:
Implemented by #699 and others.

PR #567, opened 2024-11-15
Make device_specific_init overloadable

lukasz-tuz:
`device_specific_init()` function was incorectly defined in devices.h header (as empty function). This was preventing from overloading the function, as intended.

Commit fixes this by restoring original approach from `cpu_specific_init()`; empty implementation is provided with 'weak' attribute set, allowing for overloading the function as needed.

### DIFF
--- a/framework/devicedeps/cpu/devices.h
+++ b/framework/devicedeps/cpu/devices.h
@@ -13,11 +13,4 @@ struct DeviceRange
     int device_count;
 };
 
-/*
- * Called from sandstone_main(). The default implementation performs no
- * checks, they just return. Feel free to implement a strong version elsewhere
- * if you prefer the framework to check for system or CPU criteria.
- */
-static __attribute__((unused, noinline)) void device_specific_init() {}
-
 #endif // INC_DEVICES_H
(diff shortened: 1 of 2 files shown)


framework/devicedeps/cpu/cpu_device.cpp
```diff
@@ -13,6 +13,17 @@
 
 uint64_t cpu_features = 0;
 
+/// Windows cross compilation doesn't handle __attribute__((weak)) well.
+#ifdef _WIN32
+__attribute__((noclone, noinline)) void device_specific_init()
```
thiagomacieira, line 18: The problem now is I don't see how this can work. Won't we have two symbols `device_specific_init` when another .cpp source provides this?

lukasz-tuz: Yes, you are right, it would create two definitions of `device_specific_init`. Since mingw doesn't seem to support weak symbols, I restored original solution where `device_specific_init` is defined as weak, but in the same translation unit where it is used (`sandstone.cpp`).

thiagomacieira: How was this working before, then? Ping me tomorrow early so you can show me the errors you're getting.

lukasz-tuz:
Closing in favor of the "CPU-separation" pull requests.

PR #568, opened 2024-11-21
Meson: Add multiple compilation flags

AthenasJimenez:
List of added flags:

    -D_FORTIFY_SOURCE=2
    -z relro (release only)
    -z noexecstack (release only)
    -Wformat
    -Wformat-security
    -pie

[ChangeLog] Add Read-only Relocation, Position Independent Execution, Format String and Inexecutable Stack flags to Meson file.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -94,6 +94,15 @@ nasm_system_flags = [
 default_c_flags = []
 default_cpp_flags = []
 
+add_project_arguments([
+    '-Wl,-z,relro',
+    '-Wl,-z,noexecstack',
+    '-pie',
+    '-fstack-protector-strong',
+    '-fcf-protection=full',
+], language: [ 'c', 'cpp' ])
+
+
 if target_machine.system() == 'windows'
 add_project_arguments([
     '-D__STDC_FORMAT_MACROS',
@@ -138,16 +147,26 @@ else
 if get_option('buildtype') == 'debug'
     debug_nasm_flags = '-gdwarf'
 endif
+if cc.get_id() == 'gcc'
+    march_flags += [
+        '-mcet-switch',
+    ]
+endif
+march_flags += [
+    '-fstack-clash-protection',
+]
 endif
 
 default_c_flags += [
     '-DSANDSTONE',
+    '-D_FORTIFY_SOURCE=2',
     '-fno-associative-math',
     '-fno-asynchronous-unwind-tables',
 ]
 
 default_cpp_flags += [
     '-DSANDSTONE',
+    '-D_FORTIFY_SOURCE=2',
     '-fno-associative-math',
     '-fvisibility-inlines-hidden',
 ]
@@ -166,6 +185,7 @@ endif
 default_c_warn = [
     '-Wall',
     '-Wextra',
+    '-Werror=format-security',
     '-Werror=incompatible-pointer-types',
     '-Werror=implicit-function-declaration',
     '-Werror=int-conversion',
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -163,9 +167,21 @@ if get_option('b_sanitize') != 'none'
...
 
+if get_option('buildtype') == 'release'
+    default_cpp_flags += [
+        '-z relro',
+        '-z noexecstack',
+    ]
+    default_c_flags += [
+        '-z relro',
+        '-z noexecstack',
+    ]
```
thiagomacieira, line 178: Heh, things you learn: the compiler accepts `-z`. I knew this only as a linker option and that you'd need to spell them `-Wl,-z,relro -Wl,-z,noexecstack`. But since they are linker options (and as is `-pie`), you may need to use the `add_project_link_arguments()` Meson function.

Also, this shouldn't be for release builds only, so go ahead and enable for everything.

thiagomacieira (changes requested):

meson.build
```diff
@@ -94,6 +95,16 @@ nasm_system_flags = [
 default_c_flags = []
 default_cpp_flags = []
 
+add_project_arguments([
+    '-Wl,-z,relro',
+    '-Wl,-z,noexecstack',
+    '-pie',
+    '-fstack-protector-strong',
+    '-fstack-clash-protection',
```
thiagomacieira, line 103: All additions are good. These two are the only ones I want to validate first that we don't compromise on our detection capabilities in our tests.

thiagomacieira (commented):
Approved in principle. Testing needed. @dawid-sabat, can you arrange a run of OpenDCDiag on the machines that have issues with the open source tests?

meson.build
```diff
@@ -49,6 +49,7 @@ if host_machine.cpu_family() == 'x86_64'
         '-mpclmul',
         '-mno-rdrnd',
         '-falign-loops=32',
+        '-mcet-switch',
```
thiagomacieira, line 52: Build failed with Clang: unknown option. You probably need something like:
```
    if cc.get_id() = 'gcc'
        march_flags += [
            '-mcet-switch',
        ]
    endif
```

thiagomacieira (changes requested):

dawid-sabat:
I can schedule testing.

thiagomacieira (approved):
Please wait for data before staging, but syntactically this looks good.

dawid-sabat (approved):
+1

meson.build
```diff
@@ -94,6 +94,15 @@ nasm_system_flags = [
 default_c_flags = []
 default_cpp_flags = []
 
+add_project_arguments([
+    '-Wl,-z,relro',
+    '-Wl,-z,noexecstack',
+    '-pie',
```
eli-schwartz, line 100: pie can also be handled via meson's `pie: true` per-target kwarg override or the global `-D b_pie=true` setting.

meson.build
```diff
@@ -138,16 +147,26 @@ else
...
+    ]
+endif
+march_flags += [
+    '-fstack-clash-protection',
+]
 endif
 
 default_c_flags += [
     '-DSANDSTONE',
+    '-D_FORTIFY_SOURCE=2',
```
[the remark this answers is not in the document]
thiagomacieira: I don't think that's necessary. We don't need to fortify debug builds, so doing it as #581 did is better.

As for removing the `-U`, well, anyone who chooses to build in debug mode and gets those warnings knows what to do. It's their environment that set it.

meson.build
thiagomacieira: Interesting, I didn't know about `pie: true`. It might not have been supported with Meson 0.56, which we still needed due to a CentOS 7 dependency. We don't any more, so we can modernise.

eli-schwartz: Docs say added in 0.49 :)

PR #572, opened 2024-12-10
FW: Fix hang-on-crash for second or further slice

thiagomacieira:
Amends commit bdc345642e79555cfbfc62a34028d75ddf4b6ad8 ("FW/child_debug: use futexes to make the child wait for gdb") and more importantly commit b92a2efd7eab9cbb9ad4d1a4289304ec26be1b31 ("FW/SharedMemory: add support for multiple main threads").

This caused OpenDCDiag to wake up the wrong child (always the first) when GDB backtrace support was active, which is the default. Because of that, the actual crashed child would be stuck forever in `futex_wait()` in `child_crash_handler()`:
https://github.com/opendcdiag/opendcdiag/blob/b8bdda557a388a06d66eb6a9d4455263a56e472b/framework/sysdeps/unix/child_debug.cpp#L207-L211

Additionally, we printed the backtrace in the to the first ("main") slice, which can be misleading because one might conclude it was that child that crashed.

On Windows we don't support backtraces, so there the child doesn't wait and no changes are necessary.

### DIFF
--- a/framework/sysdeps/unix/child_debug.cpp
+++ b/framework/sysdeps/unix/child_debug.cpp
@@ -1023,7 +1023,16 @@ void debug_init_child()
 #endif
 }
 
-void debug_crashed_child()
+static int find_slice_by_pid(pid_t child, std::span<const pid_t> children)
+{
+    for (size_t i = 0; i < children.size(); ++i) {
+        if (children[i] == child)
+            return int(i);
+    }
+    return -1;
+}
+
+void debug_crashed_child(std::span<const pid_t> children)
 {
     if (!SandstoneConfig::ChildDebugCrashes || sApp->shmem->server_debug_socket == -1)
         return;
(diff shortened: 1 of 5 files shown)


dawid-sabat (approved):
Looks reasonable.

framework/sysdeps/unix/child_debug.cpp
```diff
@@ -1036,16 +1045,25 @@ void debug_crashed_child()
 
         if (ctx.contents == CrashContext::NoContents)
             break;
+
+        // find out which slice this context is for
+        int slice = find_slice_by_pid(ctx.fixed.pid, children);
+        if (slice < 0) {
+            log_message(-1, "Received a context dump from PID %ld that doesn't seem to match any "
```
busykai, line 1052: We need to have a log for internal errors. I've seen quite a few places where we cannot actually continue, but it's not logged as our own problem.

busykai (approved):

framework/sysdeps/unix/child_debug.cpp
thiagomacieira: This here is expected to be an impossible condition (I could have used an `assert`). The socket is created in 
https://github.com/opendcdiag/opendcdiag/blob/dbb1bae6e9a41973af13e8728a4aed17303602d7/framework/sysdeps/unix/child_debug.cpp#L374-L386
And as a socketpair that is not accessible to any other processes, at least not accidentally. So the only thing that should write a PID is the OpenDCDiag framework itself.

The other way of doing this would be to use [`SCM_CREDENTIALS`](https://man7.org/linux/man-pages/man7/unix.7.html), but since users often run the tool as root, a corrupt child process would still be able to send corrupt data.

PR #576, opened 2025-01-30
FW: Implement cpu reschedule

AthenasJimenez:


framework/sandstone.cpp
```diff
@@ -1038,6 +1038,18 @@ int num_packages() {
     return Topology::topology().packages.size();
 }
 
+void reschedule() {
+    // Let's choose a random CPU to hop to
+    int cpu_idx = random32() % num_cpus();
```
busykai, line 1043: I think we ought to make this a queue or implement different strategies (as a test config option/knob). The reason for the queue is: it helps avoiding oversubscription of a HW for a prolonged period of time which will help the test execution time and may reveal different problems. I think we should have both, actually.

framework/sandstone.cpp
```diff
@@ -1038,6 +1039,43 @@ int num_packages() {
     return Topology::topology().packages.size();
 }
 
+int num_enabled_cpus() {
+    return sApp->enabled_cpus_count;
+}
+
+void reschedule() {
+    // Choose a random CPU to hop to
+    int cpu_idx = random32() % num_enabled_cpus();
```
busykai, line 1048: We need to make this a command-line option. For example, `--cpu-reschedule=(none|random|queue)`. This way a test would have only one function to call (`reschedule()`) and the actual rescheduling behavior could be controlled from the command-line.

framework/sandstone.cpp
```diff
@@ -1038,6 +1039,43 @@ int num_packages() {
...
+
+    if (pin_to_logical_processor(LogicalProcessor(next_cpu)))
+        log_debug("Rescheduled thread %d (%ld) to CPU %d",thread_num, pthread_self(), next_cpu);
+    else
+        log_debug("Failed to reschedule %d (%ld) to CPU %d", thread_num, pthread_self(), next_cpu);
+
+    return;
+}
+
+void reschedule_queue(std::vector<int> *reschedule_cpu_q, std::atomic<int> *cpu_idx) {
```
busykai, line 1059: This cannot and need not to be C++. Also, it should be just 1 call -- reschedule() as I outlined in the other comment.

busykai (changes requested):
We need to simplify the API to 1 call and command-line option.

framework/sandstone.cpp
busykai: `queue` should probably be the default setting (as it eliminates the chance of a CPU executing more than 1 opendcdiag thread at the same time).

framework/sandstone.cpp
```diff
@@ -1039,6 +1041,41 @@ int num_packages() {
...
+    // Get next random cpu from queue
+    int next_cpu = sApp->cpu_reschedule->get_next_cpu();
+    reschedule_next_cpu(next_cpu);
+    return;
+}
+
+void reschedule_random(std::function<void(int)> reschedule_next_cpu) {
+    // Choose a random CPU to hop to
+    int cpu_idx = random32() % num_enabled_cpus();
+    int next_cpu = sApp->cpu_reschedule->get_next_cpu_by_index(cpu_idx);
```
busykai, line 1054: It will be much cleaner design if you had a `CpuSchedule` specialized in each algorithm. Instead, the logic for one case is here and for the other elsewhere. If you did have specialized classes, you wouldn't need two different `reschedule_` calls. the `reschedule()` would look something like:

```C++

class CpuSchedule {
     ...
protected:
     struct cpu_info *cpu_info;
     CpuSchedule(struct cpu_info *cpu_info) {
     }
     ...
public:
     virtual int get_next_cpu() = 0;
}

class RandomCpuSchedule : CpuSchedule {
     public:
     RandomCpuSchedule(struct cpu_info *cpu_info) : CpuSchedule(cpu_info) {
     }
     int get_next_cpu() {
        // return random cpu
     }
}

class QueueCpuSchedule : CpuSchedule {
     public:
     QueueCpuSchedule(struct cpu_info *cpu_info) : CpuSchedule(cpu_info) {
     }
     int get_next_cpu() {
        // return cpu from the queue
     }
}

void reschedule() {
    if (sApp->reschedule_mode == RescheduleMode::Reschedule_none) return;
    int next_cpu = sApp->cpu_reschedule->get_next_cpu();
    if (!pin_to_logical_processor(LogicalProcessor(next_cpu))) {
        log_debug("Failed to reschedule %d (%ld) to CPU %d", thread_num, pthread_self(), next_cpu);
    }
}
```

`sApp` would hold a pointer to `CpuSchedule` type.

framework/sandstone_p.h
```diff
@@ -320,6 +330,40 @@ class LoggingStream
     friend LoggingStream logging_user_messages_stream(int thread_num, int level);
 };
 
+struct CpuReschedule {
```
busykai, line 333: Maybe just `Schedule` is a better name, as in `CpuSchedule`.

framework/sandstone_p.h
```diff
@@ -440,6 +488,8 @@ struct SandstoneApplication : public InterruptMonitor, public test_the_test_data
 
     SandstoneBackgroundScan background_scan;
 
+    CpuReschedule *cpu_reschedule;
```
busykai, line 491: the same schedule comment for the variable name. we have a "schedule" for using CPUs, `reschedule()` uses uses it for pick a new one.

busykai (changes requested):

thiagomacieira (changes requested):
Thank you, this is very clean. In general, I like this, and I like the queue mode that is quite simple. I like that you added another mode, showing how the modes can be extended, though I'm not convinced the random mode is a good idea.

Without a barrier, however, the queue mode can be a problem too: queuing. We may create a traffic bottleneck this way. Suppose one CPU was slower in reaching `reschedule()`, for any reason. This means the thread that got scheduled to it will compete for CPU time with the one that is already there, making both of them slower. Meanwhile, the other N-2 threads continue running and eventually one of them will get scheduled for this CPU again, causing now 3 threads to compete for CPU time. Hopefully the first one is done and gets rescheduled, but what if it doesn't yet? Remember the queue gets shuffled every N calls to `reschedule()`, which without a barrier doesn't mean "N threads" - it could be 1 thread calling it N times.

This is an unlikely, but possible scenario. It's worse if you consider that system admins will have other tasks running on the system. We know from experience that non-proficient cloud admins leave crappy stuff running (monitoring, telemetry) and we've seen 150% CPU use on idle on servers. If OpenDCDiag is being run on all CPU cores, then the OS will have to make at least one thread compete for CPU time with those background tasks.

My conclusion is that this is _very good but incomplete_. I want this to go in (once the implementation details are fixed) so we can build upon it. But it should not be enabled by default, so I think the default mode should be "none".

framework/sandstone_p.h
```diff
@@ -421,6 +427,9 @@ struct SandstoneApplication : public InterruptMonitor, public test_the_test_data
 
     int thread_count;
     int enabled_cpus_count;
+
+    int reschedule_mode = Reschedule_queue;
```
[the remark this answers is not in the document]
thiagomacieira: If we use OOP like @busykai suggested, we won't need to store the reschedule mode in `sApp`. Then just make it a local variable in the `main()` function, initialised to the default schedule mode. That will help with dead-code elimination for the `RestrictedCmdline` mode.

thiagomacieira (changes requested):

busykai:
> Without a barrier, however, the queue mode can be a problem too: queuing. We may create a traffic bottleneck this way. Suppose one CPU was slower in reaching `reschedule()`, for any reason. This means the thread that got scheduled to it will compete for CPU time with the one that is already there, making both of them slower. Meanwhile, the other N-2 threads continue running and eventually one of them will get scheduled for this CPU again, causing now 3 threads to compete for CPU time. Hopefully the first one is done and gets rescheduled, but what if it doesn't yet? Remember the queue gets shuffled every N calls to `reschedule()`, which without a barrier doesn't mean "N threads" - it could be 1 thread calling it N times.

Oversubscribing is totally fine. We're not trying to solve performance problem with this. We do need stop-the-world synchronization mechanism for this. We do need to protect the queue modifications.

On a side note, I suggested the queue as a mechanism to ensure that we don't accidentally oversubscribe CPUs (even though that's highly unlikely either with uniformly distributed `random()`).

framework/sandstone.cpp
```diff
@@ -3125,6 +3126,7 @@ int main(int argc, char **argv)
         { "quick", no_argument, nullptr, quick_run_option },
         { "quiet", no_argument, nullptr, 'q' },
         { "retest-on-failure", required_argument, nullptr, retest_on_failure_option },
+        { "cpu-reschedule", required_argument, nullptr, reschedule_option },
```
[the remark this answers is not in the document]
busykai: yeah, actually, I like it. `--reschedule` is better!

busykai:
After thinking and talking this through a bit more: let's implement a FIFO later which would be the most fair implementation to avoid thread imbalance and oversubscription ("when I'm done, my current CPU goes to the end of the queue"). That does not mean that we should get rid of either current queue implementation or random scheduling (even though they potentially risk the test timing out). The default should be `None` at the moment.

framework/sandstone.cpp
```diff
@@ -1040,6 +1040,33 @@ int num_packages() {
...
+            log_debug("Failed to reschedule %d (%ld) to CPU %d", thread_num, pthread_self(), next_cpu);
+    };
+
+    switch(sApp->reschedule_mode) {
+        case RescheduleMode::Reschedule_queue:
+            reschedule_queue(reschedule_next_cpu);
+            break;
+        case RescheduleMode::Reschedule_random:
+            // TODO: Implement reschedule random
+            break;
```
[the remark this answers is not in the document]
AthenasJimenez: The switch no longer exists, so I'm omitting this.

framework/sandstone.cpp
```diff
@@ -1040,6 +1040,33 @@ int num_packages() {
     return Topology::topology().packages.size();
 }
 
+void reschedule_queue(std::function<void(int)> reschedule_next_cpu) {
```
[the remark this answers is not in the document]
AthenasJimenez: Does this also applies for Class' methods?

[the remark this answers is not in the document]
thiagomacieira: Yes, please. These are fine:
```c++
   int foo() const { return a_; }
   void setFoo(int v)
   { a_ = v; }
   void something()
   {
          ....
   }
```

framework/sandstone_p.h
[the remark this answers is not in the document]
AthenasJimenez: After thinking about it, I think we don't need it at all at the end. So I removed it.

framework/sandstone.cpp
AthenasJimenez: Implemented

framework/topology.h
```diff
@@ -193,4 +193,8 @@ void update_topology(std::span<const struct cpu_info> new_cpu_info,
                      std::span<const Topology::Package> sockets = {});
 void restrict_topology(CpuRange range);
 
+/// Returns the number of logical CPUs that are enabled in the system but not
+/// necessarily available for the test.
+int num_enabled_cpus() __attribute__((pure));
```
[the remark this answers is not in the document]
AthenasJimenez: I was trying to have something similar to `num_cpus()` but with all available cpus on the system. Given the changes I made, this is no longer needed so I removed it.

framework/sandstone.cpp
```diff
@@ -3753,7 +3762,7 @@ int main(int argc, char **argv)
         }
     }
 
-    if (sApp->reschedule_mode == RescheduleMode::Reschedule_queue) {
+    if (sApp->reschedule_mode == RescheduleMode::Reschedule_queue || sApp->reschedule_mode == RescheduleMode::Reschedule_random) {
```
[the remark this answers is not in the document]
AthenasJimenez: hahaha removed as well

framework/sandstone.cpp
```diff
@@ -3726,6 +3753,12 @@ int main(int argc, char **argv)
         }
     }
 
+    if (sApp->reschedule_mode == RescheduleMode::Reschedule_queue) {
+        sApp->cpu_reschedule = new CpuReschedule(sApp->enabled_cpus_count, sApp->shmem->cpu_info);
```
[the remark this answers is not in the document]
AthenasJimenez: It's removed now, but just for the sake of learning, do you mean `total_cpu_count`?

AthenasJimenez:
@busykai @thiagomacieira  Thank you for your feedback!

I did addressed most if not all changes requested. At the end, I removed, re-write and re-org most of the code but I think it looks much better now. Let me know what you think.

framework/sandstone_p.h
```diff
@@ -136,6 +136,12 @@ enum ThreadState : int {
     thread_skipped = -2,
 };
 
+enum RescheduleMode : int8_t {
```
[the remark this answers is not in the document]
AthenasJimenez: Removed as well, it wasn't really needed.

framework/sandstone.cpp
```diff
@@ -1039,6 +1041,16 @@ int num_packages() {
     return Topology::topology().packages.size();
 }
 
+void reschedule()
+{
+    if (sApp->device_schedule == nullptr) return;
+    int next_cpu = sApp->device_schedule->get_next_cpu();
+    if (!pin_to_logical_processor(LogicalProcessor(next_cpu))) {
```
thiagomacieira, line 1048: Either make `get_next_cpu` return a `LogicalProcessor` (and rename to "get_next_processor") or this needs to use `LogicalProcessor(cpu_info[next_cpu].cpu_number)`

I think you need to make it the latter. See below.

framework/sandstone_p.h
```diff
@@ -320,6 +324,67 @@ class LoggingStream
...
+    }
+
+    int get_next_cpu() override
+    {
+        // Returns next CPU from the queue
+        std::lock_guard lock(q_mutex);
+        int idx = q_idx.fetch_add(1, std::memory_order_relaxed);
+        if (q_idx.load() >= queue.size())
+            shuffle_queue();
+        return queue[idx];
```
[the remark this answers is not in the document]
thiagomacieira: BTW, since you're locking the mutex, `q_idx` doesn't need to be an atomic.

framework/sandstone.cpp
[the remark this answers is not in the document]
thiagomacieira: I meant `num_cpus()`, which is what you've now used.

framework/sandstone_p.h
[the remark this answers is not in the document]
busykai: I actually think that the schedule cannot be created when the option is parsed. It should be created in `child_run()` which both fork and exec mode use to run the test. [After it called `restrict_topology()`](https://github.com/AthenasJimenez/opendcdiag/blob/athenas/fw/reschedule/framework/sandstone.cpp#L1846), that should solve it.

framework/sandstone_p.h
```diff
@@ -320,6 +324,67 @@ class LoggingStream
...
+
+private:
+    std::atomic<int> q_idx;
+    std::vector<int> queue;
+    std::mutex q_mutex;
+
+    void shuffle_queue() {
+        // Must be called with mutex locked
+        std::default_random_engine rng(random32());
+        std::shuffle(queue.begin(), queue.end(), rng);
```
busykai, line 363: It needs to be randomly re-populated from the set of available CPUs. Otherwise, if we run less threads then available CPUs, then the workloads always be migrating among the same `num_cpus()` threads instead of all available.

framework/sandstone_p.h
```diff
@@ -320,6 +324,67 @@ class LoggingStream
...
+class QueueDeviceSchedule : public DeviceSchedule {
+public:
+    QueueDeviceSchedule(struct cpu_info *cpu_info, int q_size)
+    {
+        q_idx.store(0, std::memory_order_relaxed);
+        queue.reserve(q_size);
+
+        // Populate queue
+        for (int i=0; i<q_size; i++)
+            queue.push_back(cpu_info[i].cpu_number);
```
busykai, line 342: We cannot assume that the queue is the same size as the number of available CPUs. So while this can be populated with the first `q_size` ones. I would take a different approach and have 1. a queue of size of available cpus (which is >= `q_size`) and 2. track current index (as class member). Here, populate the queue (and, optionally, shuffle) when index reaches `q_size`, shuffle the entire queue and reset index to 0.

framework/sandstone_p.h
```diff
@@ -320,6 +324,67 @@ class LoggingStream
...
+
+class RandomDeviceSchedule : public DeviceSchedule {
+public:
+    RandomDeviceSchedule(struct cpu_info *cpu_info, int q_size)
+    {
+        queue.reserve(q_size);
+
+        // Populate queue
+        for (int i=0; i<q_size; i++)
+            queue.push_back(cpu_info[i].cpu_number);
```
busykai, line 376: Same comment here, this limits the scope of CPUs that we migrate between. If we run 4 software threads on 64 hw threads, they will always migrate between the first 4 hw threads, the rest will be always idle. In case of random, it should be populated with all available CPUs (to this slice).

framework/sandstone_p.h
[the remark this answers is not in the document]
thiagomacieira: The object can be created in `main()` but then updated by `child_run()`. What the class does is up to the implementation.

framework/sandstone_p.h
thiagomacieira: That's exactly what it should be doing. If the user told us to run only on a given set of CPUs using [taskset(1)](https://man7.org/linux/man-pages/man1/taskset.1.html) or `--cpuset=`, we should obey and only run on that set. We should not "leak" and go outside.

In any case, topology.cpp should already be obeying that: it should not have queried the CPUs not in the ambient set at all, should have restricted to `--cpuset=` before this is called in the first place.

What's important is that this be fixed to work with slicing.

framework/sandstone.cpp
AthenasJimenez: I prefer the former. The problem using `cpu_info` is that the slices only have the info of the cpus that belong to that specific slice, and I need to reschedule in all available cpus no matter what slice they belong to.

framework/sandstone_p.h
busykai: Slicing is a separate issue. I'm talking about this scenario:

``` bash
$ nproc
24
$ opendcdiag -n 4 --schedule=queue
# this will migrate only between the first 4.
```

framework/sandstone_p.h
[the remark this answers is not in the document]
AthenasJimenez: The purpose of having a `queue` and an `index` was to have a single control point for all threads, no matter the slices. If we are going to have a `queue` per slice and an `index` per slice, I think  we lost the whole purpose of having a `queue`.

framework/sandstone.cpp
```diff
@@ -3394,6 +3407,23 @@ int main(int argc, char **argv)
...
+                fprintf(stderr, "%s: --reschedule is only useful with at least 2 threads\n", argv[0]);
+                return EX_USAGE;
+            }
+
+            if (strcmp(optarg, "none") == 0 ) {
+                // Default option, so do nothing
+            } else if (strcmp(optarg, "queue") == 0) {
+                sApp->device_schedule = std::make_unique<QueueDeviceSchedule>(sApp->shmem->cpu_info, num_cpus());
+            } else if (strcmp(optarg, "random") == 0) {
+                sApp->device_schedule = std::make_unique<RandomDeviceSchedule>(sApp->shmem->cpu_info, num_cpus());
```
thiagomacieira, line 3421: The `-n` restriction hasn't taken effect yet at this point. However, the fact we're processing things immediately means the order of `--cpuset` and `--reschedule` matters. Please make it so they don't.

thiagomacieira (changes requested):

framework/sandstone_p.h
busykai: With `-n <num>` the framework restricts the CPU set to the first `<num>` number of threads in the same way `--cpuset` option or `taskset` would do. Here is the [current code](https://github.com/opendcdiag/opendcdiag/blob/85f9251f519e943a446a9a640aed929068cfe436/framework/sandstone.cpp#L3637). So my comment here withdrawn. However, it also means that `q_size` argument to the constructor is excessive. Since it's always `num_cpus()`.

framework/sandstone.cpp
AthenasJimenez: I still would like to jump across slices, not sure about jumping across NUMA. But that would be for another upgrade PR.

thiagomacieira: The test that wants to do that could declare itself `isolate_socket`.

framework/sandstone.cpp
```diff
@@ -1039,6 +1041,16 @@ int num_packages() {
     return Topology::topology().packages.size();
 }
 
+void reschedule()
+{
+    if (sApp->device_schedule == nullptr) return;
+    int next_cpu = sApp->device_schedule->get_next_cpu();
+    if (!pin_to_logical_processor(LogicalProcessor(cpu_info[0].cpu_number))) {
```
thiagomacieira, line 1048: You didn't use `next_cpu` here. You probably meant `cpu_info[next_cpu].cpu_number`.

thiagomacieira (changes requested):
That's why we need a test.

framework/sandstone.cpp
AthenasJimenez: Indeed. While writing the selftest I was so confused. lol

framework/sandstone_p.h
AthenasJimenez: I think this will be pending for a future PR.

framework/sandstone_p.h
AthenasJimenez: Agree, so removed.

framework/sandstone_p.h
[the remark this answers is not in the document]
AthenasJimenez: Yeah, still need to have a better understanding about how the slices works.

bats/sanity-check/20-selftests.bats
```diff
@@ -1905,6 +1905,33 @@ selftest_cpuset_negated() {
     done
 }
 
+# Confirm we are rescheduling properly
+@test "thread queue reschedule" {
+    local -a cpuset=(`$SANDSTONE --dump-cpu-info | awk '/^[0-9]/ { print $1 }'`)
+    nproc=${#cpuset[@]}
+
+    declare -A yamldump
+    sandstone_selftest -e selftest_logs_reschedule -s AES:79a6a544fabb099aae15df8b23a5ef7086595abb0544f66551ea2074dc5a108f --reschedule=queue
```
thiagomacieira, line 1914: Could you use an LCG seed? The AES generator is not available in all CPUs.

framework/selftest.cpp
```diff
@@ -179,6 +180,70 @@ static int selftest_logs_getcpu_run(struct test *test, int cpu)
...
+#if defined(_WIN32)
+    PROCESSOR_NUMBER number;
+    GetCurrentProcessorNumberEx(&number);
+    // see win32/cpu_affinity.cpp
+    cpu_number = number.Group * 64 + number.Number;
+#elif defined(__linux__) || defined(__FreeBSD__)
+    cpu_number = sched_getcpu();
+#else
+    log_skip(OsNotSupportedSkipCategory, "No API to get the CPU number on this OS");
+#endif
```
thiagomacieira, line 211: Please deduplicate from the `logs_getcpu` code - just write local `static` function to get the CPU.

For the `#else` case, you can do:
```c
    errno = ENOSYS;
    return -1;
```
and then just leave the "OS failed:" text below.

framework/selftest.cpp
```diff
@@ -179,6 +180,70 @@ static int selftest_logs_getcpu_run(struct test *test, int cpu)
     return EXIT_SUCCESS;
 }
 
+static int selftest_logs_reschedule_init(struct test *test)
+{
+    // In order to always get the same result and avoid race conditions,
+    // we use semaphores to synchronize the access to reschedule()
+    int sem_size = num_cpus() - 1;
+    sem_t *reschedule_sem = (sem_t *) malloc(sizeof(sem_t)*sem_size);
```
thiagomacieira, line 188: `calloc(sem_size, sizeof(sem_t))`

Alternatively, you could use `std::binary_semaphore`. I've never used it, but it supposed to work.
```c++
    test->data = new std::binary_semaphore[sem_size];
// in cleanup:
    delete[] static_cast<std::binary_semaphore *>(test->data);
```

framework/selftest.cpp
```diff
@@ -179,6 +180,70 @@ static int selftest_logs_getcpu_run(struct test *test, int cpu)
...
+        log_skip(OSResourceIssueSkipCategory, "OS failed: %m");
+    log_info("%d", cpu_number);
+
+    return EXIT_SUCCESS;
+}
+
+static int selftest_logs_reschedule_cleanup(struct test *test)
+{
+    sem_t *reschedule_sem = (sem_t *) test->data;
+    free (reschedule_sem);
```
thiagomacieira, line 244: Need to `sem_destroy()` the entries.

thiagomacieira (changes requested):
Almost there!

thiagomacieira (approved):

PR #579, opened 2025-02-18
Framework extension to log a message without a use of logging facility

jposwiata:
PTS console (as passed with PTS environment variable) is used to log all the messages.
Default implementation to print (xbyak) code and to print buffer (as bytes) included.

`--pts` (or `PTS` environment variables) allows to define where the logs should be logged. Any text file, device file (e.g. any pseudo-terminal session /dev/pts/xxx, e.g. `tmux` pane tty) or open file descriptor (`./sandstone --pts=3 .... 3>log.txt`) works. Per-cpu syntax not available.
`PRINT_ASM` environment variable might be used to pass the script to parse binary into human readable assembly form (defaults to `print_asm.sh`). Script not included.

### DIFF
--- a/framework/logging.cpp
+++ b/framework/logging.cpp
@@ -105,6 +105,8 @@ static uint8_t progress_bar_needs_flush = false;
 
 static constexpr auto UsedKnobValueLoggingLevel = LOG_LEVEL_VERBOSE(1);
 
+static std::mutex immediate_message_mutex{};
+
 namespace {
 enum LogTypes {
     UserMessages = 0,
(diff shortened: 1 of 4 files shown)


thiagomacieira (commented):
I don't want this. Stuff going to the PTY causes a hash of the terminal when trying to use a full screen application, like Emacs shell mode, Emacs GUD, GDB TUI mode, etc.

For Xbyak content, we should just add a `log_jit_code` that saves a copy and prints it if the test failed.

jposwiata:
Idea here was to use separate pseudo-terminals for each process (or with `-n 1` to use default passed pts to log messages "just in time", to allow mixing with other messages, like telemetry data).
For printing out the code we'd rather need to link agains XED library and decode instructions in the code. I wonder if it would be accepted at all.
For `log_jit_code()` we can use /dev/stderr, which is handled to follow your needs by the framework.

thiagomacieira:
> Idea here was to use separate pseudo-terminals for each process (or with `-n 1` to use default passed pts to log messages "just in time", to allow mixing with other messages, like telemetry data). 

That's only valid for development of tests. I understand you want to get some feedback as the test is running, instead of waiting for it to finish and the framework logging to print out what happened. But that's a rather narrow niche case and doesn't remain in the test content even in debug mode builds.

The bigger problem I have is have it tied to a TTY. That messes things up. Can this be reduced to a simple "log to file descriptor" and be done with it? The person developing the test can easily open the PTYs if they want, `opendcdiag -e testname --extra-fd-log=3 3>/dev/pty/123`.

> For printing out the code we'd rather need to link agains XED library and decode instructions in the code. I wonder if it would be accepted at all. For `log_jit_code()` we can use /dev/stderr, which is handled to follow your needs by the framework.

I'd rather we simply logged the bytes and let post-processing decode them. We can do that in the framework, by calling out to other processes if we wanted to, but hex dump of bytes are smaller in the log than a full disassembly.

jposwiata:
> That's only valid for development of tests. I understand you want to get some feedback as the test is running, instead of waiting for it to finish and the framework logging to print out what happened. But that's a rather narrow niche case and doesn't remain in the test content even in debug mode builds.
> The bigger problem I have is have it tied to a TTY. That messes things up. Can this be reduced to a simple "log to file descriptor" and be done with it? The person developing the test can easily open the PTYs if they want, `opendcdiag -e testname --extra-fd-log=3 3>/dev/pty/123`.

Do you mean an extension of `log_debug()`? Makes sense.
In general, the output is tied to any file (it might be a device or regular file). The "interface" for the file has changed several times (env/knobs, single file/list of files/TTYs "per thread", with fopen/fd only, etc). Each has cons and pros. Simplest model with a mutex should do (hmm.. is process-shared mutex required here?)

> I'd rather we simply logged the bytes and let post-processing decode them. We can do that in the framework, by calling out to other processes if we wanted to, but hex dump of bytes are smaller in the log than a full disassembly.

For quick debugging it is usefull to see the assembly (instead of copy/paste to the script). Printing out the bytes could be fair when the script has failed (no script or the binary is too long to pass as a cmd line).

jposwiata:
Lets keep private debug capabilities private

PR #581, opened 2025-02-18
Meson: move setting of _FORTIFY_SOURCE=2

thiagomacieira:
The order was wrong: `-U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=2` is still setting it. That causes glibc to complain that we were using the option in non-optimised builds.

Amends f8dcbb86fb924032f835bba341bfc7965a5b06b0.


AthenasJimenez (approved):
LGTM

PR #582, opened 2025-02-18
FW: add a way for OpenDCDiag to ignore SIGBUS relating to MCEs

thiagomacieira:
This adds a `--ignore-mce-errors` option, which a) disables the `mce_check` test and b) makes the framework ignore SIGBUS errors that are related to MCEs. The Linux kernel sends those with `.si_code` equal to either `BUS_MCEERR_AO` or `BUS_MCEERR_AR`.

[ChangeLog][Framework] Added `--ignore-mce-errors` option, which will disable handling of Machine Check Exceptions. This option is useful if the user wants the tool to only react to Silent Data Errors (uncorrected MCEs are User-Observable Errors).

[ChangeLog][Framework] The --ignore-mce-errors will also cause the tool to ignore SIGBUS errors sent by the Linux kernel with signal codes indicating they were caused by MCEs. If those are detected, the test will terminate the test and mark it as skipped.

In debug builds, we also ignore SIGTRAP (Breakpoints), which allows this functionality to be tested. There's no way to send the required `si_code` for SIGBUS from userspace - only the OS kernel is allowed to set those values.

### DIFF
--- a/framework/sysdeps/unix/child_debug.cpp
+++ b/framework/sysdeps/unix/child_debug.cpp
@@ -1007,7 +1007,7 @@ void debug_init_child()
     action.sa_sigaction = child_crash_handler;
     action.sa_flags = SA_NODEFER;       // allow recursive signalling, so the child can raise()
     action.sa_flags |= SA_SIGINFO;
-    for (int signum : { SIGILL, SIGABRT, SIGFPE, SIGBUS, SIGSEGV }) {
+    for (int signum : { SIGILL, SIGTRAP, SIGABRT, SIGFPE, SIGBUS, SIGSEGV }) {
         sigaction(signum, &action, nullptr);
     }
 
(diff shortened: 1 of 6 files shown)


fenrus75 (dismissed):

fenrus75 (approved):

framework/sysdeps/unix/child_debug.cpp
```diff
@@ -187,8 +187,39 @@ print("..Done..")
...
-static void child_crash_handler(int, siginfo_t *si, void *ucontext)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wformat-security"
+template <typename... Args> [[noreturn]] static void
+terminate_with_ignored_crash(const char *msg, Args... params)
 {
+    // we can't really recover from a signal, so we terminate the
+    // process in a way the parent can tell it was an ignored error
+    log_message_skip(-1, IgnoredMceCategory, msg, params...);
+    log_warning("MCE was delivered to or is related to this thread");
```
thiagomacieira, line 198: Note, it's a horrible idea to use the logging framework from inside a SIGBUS handler. We may deadlock.  Without it, I'd need to find some other way to communicate what happened.

busykai (approved):

PR #583, opened 2025-02-20
FW/child_debug: fix returning from the signal handler if socket==-1

thiagomacieira:
This isn't supposed to happen, because we don't install the signal handler in the first place if we don't have both ends of the socket pair. But in case we do, the correct thing to do is to repeat the same suicide protocol.

Also fixing the detection of when to do so: we can detect a kernel-originated signal with `si_code >= 0`.


framework/sysdeps/unix/child_debug.cpp
```diff
@@ -209,21 +213,19 @@ static void child_crash_handler(int, siginfo_t *si, void *ucontext)
...
         raise(si->si_signo);
     }
 
-    // wait forever
-    struct pollfd dummy;
-    while (true)
-        poll(&dummy, 0, -1);
+    // If the signal was generated by the OS and we return, we'll execute
+    // the same instruction and crash again. Without the signal handler in
+    // place, the core dump should point to to the exact locus now.
```
busykai, line 228: typo "to to"

thiagomacieira: Fixed.

busykai (approved):

PR #585, opened 2025-02-24
FW: Store worker threads process ID

AthenasJimenez:


framework/sandstone_thread.cpp
```diff
@@ -89,6 +89,8 @@ void SandstoneTestThread::start(RunnerFunction *f, int cpu)
     auto runner = +[](void *ptr) {
         auto self = static_cast<SandstoneTestThread *>(ptr);
         ::thread_num = self->thread_num;
+        sApp->test_thread_data(self->thread_num)->tid.store(gettid());
+        auto clear_tid = scopeExit([self] { sApp->test_thread_data(self->thread_num)->tid.store(0); });
```
thiagomacieira, line 93: These should be `std::memory_order_relaxed`.

AthenasJimenez: Fixed

thiagomacieira (approved):

PR #587, opened 2025-02-24
CI: Upgrade the Windows build to Fedora 41

thiagomacieira:
We need to do these once in a while.

To do that, we need to stop calling `SetThreadIdealProcessorEx` because that breaks the CI.

### DIFF
--- a/framework/sysdeps/windows/cpu_affinity.cpp
+++ b/framework/sysdeps/windows/cpu_affinity.cpp
@@ -30,6 +30,14 @@ static void set_thread_name(const char *thread_name)
     (void) thread_name;
 }
 
+[[maybe_unused]] static bool running_in_wine()
+{
+    HMODULE hMod = GetModuleHandleW(L"ntdll.dll");
+    if (hMod)
+        return GetProcAddress(hMod, "wine_get_version") != nullptr;
+    return false;
+}
+
 LogicalProcessorSet ambient_logical_processor_set()
 {
     LogicalProcessorSet result = {};        // memsets to zero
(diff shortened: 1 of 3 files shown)


framework/sysdeps/windows/meson.build
```diff
@@ -22,3 +22,7 @@ framework_files += \
         '../generic/msr.c',
         '../generic/physicaladdress.c',
     )
+
+if run_command([shell, '-c', 'echo $ENABLE_WINE_WORKAROUNDS'], capture: true).stdout().strip() == '1'
+    default_cpp_flags += [ '-DENABLE_WINE_WORKAROUNDS', ]
```
busykai, line 27: maybe we need to start using `add_project_arguments()` for this, like grown-ups, but that's a separate story.

busykai (approved):

framework/sysdeps/windows/meson.build
thiagomacieira: I thought of that, but `-Dsomething` requires adding something to meson_options.txt and I didn't want this option to be documented.

framework/sysdeps/windows/cpu_affinity.cpp
```diff
@@ -51,12 +59,20 @@ bool pin_to_logical_processor(LogicalProcessor n, const char *thread_name)
     if (n == LogicalProcessor(-1))
         return true;            // don't change affinity
 
+    HANDLE hThread = GetCurrentThread();
+
+#ifdef ENABLE_WINE_WORKAROUNDS
+    // Work around WINE not implementing SetThreadIdealProcessorEx
+    static const bool is_wine = running_in_wine();
+    if (is_wine && SetThreadIdealProcessor(hThread, DWORD(n)) == 0)
```
thiagomacieira, line 67: This is probably causing the CI failures when run on a 2-core hyperthreaded CI (looks like AMD CPU).

PR #588, opened 2025-02-25
Idle injection in test execution

lukasz-tuz:
Commit introduces a new feature: injecting idle (sleep) into the execution of the test loop. This feature puts more stress on the cores executing the test content to create conditions in which silent data corruption can be detected more easily.

Injection of idle is triggered by a new command line option - `--inject-idle <T>,` where `T` determines how much sleep time should be injected into the test loop (expressed as a percentage of test duration). The allowed range is (0, 50), inclusive, and saturates when an out-of-bounds value is supplied.

When `inject-idle` is set to a non-zero value, OpenDCDiag will calculate the test's idle duration for each fracture and adjust the test's run duration accordingly to avoid modifying the overall test timings.

Idle cycles are injected when evaluating timing conditions inside the `TEST_LOOP`. OpenDCDiag will inject the `usleep` command with duration determined as the pre-calculated test's idle duration divided by number of test iterations.

The graph below shows core C6 residency measured by `turbostat` while running OpenDCDiag with different values provided to the `--inject-idle` option:

image

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -724,7 +724,7 @@ extern "C" void test_loop_iterate() noexcept;    // see below
 
 /* returns 1 if the test should keep running, useful for a while () loop */
 #undef test_time_condition
-bool test_time_condition() noexcept
+bool test_time_condition(int N) noexcept
 {
     test_loop_iterate();
     sApp->test_tests_iteration(current_test);
@@ -733,6 +733,10 @@ bool test_time_condition() noexcept
     if (max_loop_count_exceeded(current_test))
         return 0;  // end the test if max loop count exceeded
 
+    if (sApp->shmem->current_test_sleep_duration != 0us) {
+        usleep(sApp->shmem->current_test_sleep_duration.count() / N);
+    }
+
     return !wallclock_deadline_has_expired(sApp->shmem->current_test_endtime);
 }
 
@@ -740,16 +744,6 @@ bool test_is_retry() noexcept
 {
     // negative values indicate a retry
     return sApp->current_iteration_count < 0;
-}
-
-bool test_inject_idle(int N) noexcept
-{
-    if (sApp->shmem->current_test_sleep_duration > 0) {
-        usleep(sApp->shmem->current_test_sleep_duration / N);
-    }
-    // Always return false so that the condition check in TEST_LOOP
-    // doesn't create an infinite loop.
-    return false;
 }
 
 // Creates a string containing all socket temperatures like: "P0:30oC P2:45oC"
@@ -2414,11 +2408,14 @@ run_one_test(const test_cfg_info &test_cfg, SandstoneApplication::PerCpuFailures
     first_iteration_target = MonotonicTimePoint::clock::now() + 10ms;
 
     if (sApp->inject_idle > 0) {
-        sApp->shmem->current_test_sleep_duration = duration_to_usec(
-            sApp->current_test_duration * sApp->inject_idle / 100
+        float idle_rate = sApp->inject_idle / 100.0;
+        sApp->shmem->current_test_sleep_duration = duration_cast<microseconds>(
+            sApp->current_test_duration * idle_rate
         );
         // When injecting idle time, reduce the test duration accordingly
-        sApp->current_test_duration = sApp->current_test_duration * (1 - sApp->inject_idle / 100);
+        sApp->current_test_duration = duration_cast<ShortDuration>(
+            sApp->current_test_duration * (1 - idle_rate)
+        );
     }
 
     if (test->fracture_loop_count) {
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -724,7 +724,7 @@ extern "C" void test_loop_iterate() noexcept;    // see below
 
 /* returns 1 if the test should keep running, useful for a while () loop */
 #undef test_time_condition
-bool test_time_condition(int N) noexcept
+bool test_time_condition() noexcept
 {
     test_loop_iterate();
     sApp->test_tests_iteration(current_test);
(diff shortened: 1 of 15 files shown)


thiagomacieira (changes requested):
Good idea.

framework/sandstone.cpp
```diff
@@ -2344,6 +2355,14 @@ run_one_test(const test_cfg_info &test_cfg, SandstoneApplication::PerCpuFailures
     sApp->current_test_duration = test_duration(test_cfg);
     first_iteration_target = MonotonicTimePoint::clock::now() + 10ms;
 
+    if (sApp->inject_idle > 0) {
+        sApp->shmem->current_test_sleep_duration = duration_to_usec(
+            sApp->current_test_duration * sApp->inject_idle / 100
+        );
+        // When injecting idle time, reduce the test duration accordingly
+        sApp->current_test_duration = sApp->current_test_duration * (1 - sApp->inject_idle / 100);
```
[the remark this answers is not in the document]
thiagomacieira: Wait, the first one will work, because we multiply first, before dividing.

The second won't work. It will always be zero.

We could also just use a simple `float` with a value from 0 to 0.5. An [FP multiply](https://uops.info/html-instr/VMULSS_XMM_XMM_XMM.html) is 4 cycles, whereas a [32-bit integer divide](https://uops.info/html-instr/IDIV_R32.html) can be up to 14 cycles (on modern P-cores).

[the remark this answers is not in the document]
thiagomacieira: Even if you add the [CVTSS2SI](https://uops.info/html-instr/VCVTSS2SI_R32_XMM.html) of 7 cycles and register file change, we're mostly winning.

framework/sandstone.cpp
```diff
@@ -739,6 +740,16 @@ bool test_is_retry() noexcept
     return sApp->current_iteration_count < 0;
 }
 
+bool test_inject_idle(int N) noexcept
+{
+    if (sApp->shmem->current_test_sleep_duration > 0) {
+        usleep(sApp->shmem->current_test_sleep_duration / N);
```
thiagomacieira, line 746: While working on the Windows changes, I found a very old commit of mine that adds a `clock_sleep_for(Duration d)`. I think it's worthwhile resurrecting it.

I can do that after your change.

thiagomacieira (changes requested):

thiagomacieira (changes requested):

framework/sandstone_p.h
```diff
@@ -470,6 +474,7 @@ struct SandstoneApplication::SharedMemory
     // test execution
     MonotonicTimePoint current_test_endtime = {};
     int current_max_loop_count = 0;
+    std::chrono::duration<int, std::micro> current_test_sleep_duration = 0us;
```
thiagomacieira, line 477: Can you do with milliseconds here? If so, we have `ShortDuration`

Please replace the `0us` with `{}`, so you don't need the `using namespace` above.

framework/selftest.cpp
```diff
@@ -926,6 +926,25 @@ static const kvm_config_t *selftest_kvm_config_real_16bit_fail()
 }
 #endif // __linux__ && x86-64
 
+static int selftest_inject_idle(struct test *test, int cpu)
+{
+    auto loop_start = std::chrono::steady_clock::now();
+    TEST_LOOP(test, 1) {
+        for (int i = 0; i < 10000; ++i) {
+            int a_variable = cpu * 2;
+            a_variable += 1;
```
thiagomacieira, line 935: The compiler will optimise this out of existence, along the entire loop.

Append:
```c++
    // use the result so the compiler doesn't optimize this away
    __asm__ ("" : "+gv" (a_variable));
```

framework/selftest.cpp
```diff
@@ -926,6 +926,25 @@ static const kvm_config_t *selftest_kvm_config_real_16bit_fail()
...
+        }
+    }
+    auto loop_end = std::chrono::steady_clock::now();
+    auto elapsed_time = duration_cast<microseconds>(loop_end - loop_start);
+    auto sleep_duration = sApp->shmem->current_test_sleep_duration;
+
+    if ((elapsed_time > 1.01 * sleep_duration) || (elapsed_time < 0.99 * sleep_duration)) {
```
thiagomacieira, line 942: The less-than is fine.

The more-than is likely going to be a source of flaky failures due to VMs in GitHub being slow for any reason. I think that's why the Windows run failed, not due to Windowsnesses. Also, it's a debugoptimized build, so the loop above will have been eliminated.

framework/sandstone.h
```diff
@@ -546,6 +546,14 @@ extern void test_loop_end(void) noexcept;
 extern bool test_time_condition() noexcept;
 #define test_time_condition(test)       test_time_condition()
 
+/// Called from the TEST_LOOP macro to determine if loop should continue.
+/// Argument N is the requested number of loop iterations. If idle cycle
+/// injection is configured, this function will issue usleep() for calculated time
+/// (configured idle time / N) before calling test_time_condition().
+/// This way compatibility with tests that call test_time_condition() directly
+/// is maintained.
```
thiagomacieira, line 554: Superfluous information that is likely going to become stale soon.

thiagomacieira (changes requested):

framework/sandstone_p.h
lukasz-tuz: > Can you do with milliseconds here? If so, we have `ShortDuration`

Sadly, no. With milliseconds we don't get enough control over C-state residency.

thiagomacieira: Ok, understood.

thiagomacieira (approved):

PR #589, opened 2025-02-25
FW/frequency_manager: simple clean ups and fixes

thiagomacieira:
First, include what you need. Otherwise, IDEs show everything as underlined in this file.

Second, use `AutoClosingFile` to avoid having to remember to `fclose`. This amends commit c837a1d2e52e027fbe6165cb322827248f93e708.

Finally, refer to the object indirectly in sApp. This is a 192-byte structure, which is not used by default in the majority of runs. Move it off to a pointer so we don't need to occupy space in the sApp global for the majority of users.


lukasz-tuz (approved):
Looks ok.

PR #590, opened 2025-02-25
FW: implement TEST_QUALITY_OPTIONAL

thiagomacieira:
This is meant for tests that are production-quality, but for whatever reason we don't want to have enabled by default. If the test is explicitly named, it will get run (it won't match wildcards).

A good case for this may be an obsolete test, which may still be useful in some circumstances, like for an A-B comparison with the updated version.

### DIFF
--- a/tests/mce_check/mce_check.cpp
+++ b/tests/mce_check/mce_check.cpp
@@ -106,7 +106,7 @@ struct test mce_test = {  // This variable is used in the framework!
         .test_run = mce_check_run,
         .desired_duration = -1,
         .fracture_loop_count = -1,
-        .quality_level = INT_MAX,
+        .quality_level = TEST_QUALITY_PROD,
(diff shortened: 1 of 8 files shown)


tests/mce_check/mce_check.cpp
```diff
@@ -106,7 +106,7 @@ struct test mce_test = {  // This variable is used in the framework!
         .test_run = mce_check_run,
         .desired_duration = -1,
         .fracture_loop_count = -1,
-        .quality_level = INT_MAX,
+        .quality_level = test_quality(INT_MAX),
```
busykai, line 109: Let's just use `TEST_QUALITY_PROD` here instead of casting it to enum.

thiagomacieira: Sure.

busykai (approved):

PR #591, opened 2025-02-26
Meson: silence warnings about -Wl options during compilation

thiagomacieira:
GCC isn't complaining, but Clang is:

```
clang++-19: warning: -Wl,-z,relro: 'linker' input unused [-Wunused-command-line-argument]
clang++-19: warning: -Wl,-z,noexecstack: 'linker' input unused [-Wunused-command-line-argument]
clang++-19: warning: argument unused during compilation: '-pie' [-Wunused-command-line-argument]
```

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -94,10 +94,16 @@ nasm_system_flags = [
 default_c_flags = []
 default_cpp_flags = []
 
-add_project_arguments([
+foreach ccflag : [
+    '-fstack-clash-protection',
     '-fstack-protector-strong',
     '-fcf-protection=full',
-], language: [ 'c', 'cpp' ])
+    '-mcet-switch',
+]
+    if cc.has_argument(ccflag)
+        add_project_arguments(ccflag, language: [ 'c', 'cpp' ])
+    endif
+endforeach
 
 foreach lflag : [
     '-Wl,-z,relro',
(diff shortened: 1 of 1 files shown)


AthenasJimenez (approved):

meson.build
```diff
@@ -95,13 +95,14 @@ default_c_flags = []
...
-    '-pie',
     '-fstack-protector-strong',
     '-fcf-protection=full',
 ], language: [ 'c', 'cpp' ])
-
+add_project_link_arguments([
+    '-Wl,-z,relro',
+    '-Wl,-z,noexecstack',
+    '-pie',
+], language : [ 'c', 'cpp' ])
```
AthenasJimenez, line 105: In a second though, we need to clarify this options are only for `native` compilations and not for `cross` compilations to not break Windows build.
```
101 add_project_link_arguments([
102     '-Wl,-z,relro',
103     '-Wl,-z,noexecstack',
104     '-pie',
105 ], language : [ 'c', 'cpp' ],
106 native: true)
```

AthenasJimenez (changes requested):

meson.build
thiagomacieira: Or, rather, they are for not-windows. They do apply to FreeBSD, for example. They wouldn't apply to macOS.

There must be a way to dynamically check them. Will update.

busykai (approved):

AthenasJimenez (approved):

PR #593, opened 2025-02-26
CI: Fix runs on Windows with WINE

thiagomacieira:
During one of our tests under WINE, we're getting the outputs:
```
error: XDG_RUNTIME_DIR is invalid or not set in the environment.
0690:err:winediag:nodrv_CreateWindow Application tried to create a window, but no driver could be loaded.
```

The former is coming from libwayland-client, probably through Gtk as a dependency of WINE:
https://gitlab.freedesktop.org/wayland/wayland/-/blob/1.23.1/src/wayland-client.c?ref_type=tags#L1154-1160

The second happens when WINE attempts to launch the debugger, but there's no GUI in the CI.

Both began to be a problem with 3432e1190e343d71a516ceadedb6f31cfef62e63 (PR #587). So we now set `XDG_RUNTIME_DIR` and suppress all WINE output.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -78,7 +78,7 @@ jobs:
     container: fedora:41
     env:
       ENABLE_WINE_WORKAROUNDS: 1
-      WINEDEBUG: fixme-all,-dbghelp_stabs
+      WINEDEBUG: -all
       WINEPREFIX: /tmp/wine_root
       WINEPATH: d:/bin
       XDG_RUNTIME_DIR: /tmp/wine_root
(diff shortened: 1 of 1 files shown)


thiagomacieira:
https://github.com/thiagomacieira/opendcdiag/actions/runs/13555591124 shows how this runs, but it's not working.

I think it's time to ditch WINE and accept running on Windows.

busykai (approved):
Pushing this through. I believe the CI will still be broken since one of the self tests is broken in wine. And yes, we need to move to use a Windows runner.

PR #597, opened 2025-03-07
FW: initialize rescheduling on first use

thiagomacieira:
This fixes running it in `--fork=no` mode, because we don't require something in `child_run()`. Moreover, it also makes the start up of tests that don't ever `reschedule()` be faster.

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -1049,17 +1049,12 @@ public:
     {
         // Returns next CPU from the queue
         std::lock_guard lock(q_mutex);
-        if (q_idx >= queue.size())
+        if (q_idx == 0)
             shuffle_queue();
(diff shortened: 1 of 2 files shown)


framework/sandstone.cpp
```diff
@@ -1041,6 +1041,52 @@ int num_packages() {
...
+
+    void shuffle_queue()
+    {
+        if (queue.size() == 0) {
+            // First use: populate queue with the indexes available
+            for (int i=0; i<num_cpus(); i++)
+                queue.push_back(i);
+        }
+
+        // Must be called with mutex locked
```
thiagomacieira, line 1073: This comment should have remained at the top of the file...

AthenasJimenez (approved):

PR #598, opened 2025-03-07
FW: make it clear --reschedule doesn't work in exec mode (Windows)

thiagomacieira:
Because the `sApp->device_schedule` class hasn't been created in the child process.

This can be fixed by delaying the creation until `exec_mode_run()`, with some information passed to the child via the shmem block. This isn't necessary for now.


AthenasJimenez (approved):
LGTM

PR #602, opened 2025-03-19
FW: Implement barrier CPU rescheduling

AthenasJimenez:


AthenasJimenez:
I still have pending to deal with `num_cpus()` not divisible by `members_per_group`, but I'd like to some feedback before keep moving.

framework/sysdeps/windows/cpu_affinity.cpp
```diff
@@ -64,6 +64,25 @@ bool pin_to_logical_processor(LogicalProcessor n, const char *thread_name)
...
+
+    HANDLE hThread = OpenThread(THREAD_ALL_ACCESS, FALSE, thread_id);
+    if (!SetThreadGroupAffinity(hThread, &groupAffinity, nullptr)) {
+        win32_perror("SetThreadGroupAffinity");
+        return false;
+    }
+    return true;
+}
+
 bool pin_to_logical_processors(CpuRange range, const char *thread_name)
```
thiagomacieira, line 86: Similar here, but I'd recommend an internal function that takes a `HANDLE` parameter instead of the `tid_t`.

framework/sysdeps/windows/cpu_affinity.cpp
```diff
@@ -64,6 +64,25 @@ bool pin_to_logical_processor(LogicalProcessor n, const char *thread_name)
...
+    GROUP_AFFINITY groupAffinity = {};
+    groupAffinity.Group = processorNumber.Group;
+    groupAffinity.Mask = KAFFINITY(1) << processorNumber.Number;
+
+    HANDLE hThread = OpenThread(THREAD_ALL_ACCESS, FALSE, thread_id);
+    if (!SetThreadGroupAffinity(hThread, &groupAffinity, nullptr)) {
+        win32_perror("SetThreadGroupAffinity");
+        return false;
+    }
+    return true;
```
thiagomacieira, line 83: `CloseHandle` missing.

framework/sandstone.cpp
```diff
@@ -1045,18 +1045,24 @@ namespace {
 class QueueDeviceSchedule : public DeviceSchedule
 {
 public:
-    int get_next_cpu() override
-    {
-        // Returns next CPU from the queue
+    void device_reschedule() override {
```
thiagomacieira, line 1134: Nitpick: position of the `{`

framework/sandstone.cpp
```diff
@@ -1079,21 +1085,24 @@ class QueueDeviceSchedule : public DeviceSchedule
 class RandomDeviceSchedule : public DeviceSchedule
 {
 public:
-    int get_next_cpu() override
-    {
-        // return random cpu
-        return random32() % num_cpus();
+    void device_reschedule() override {
+        // Select a random cpu index among the ones available
+        int next_cpu = random32() % num_cpus();
```
thiagomacieira, line 1176: Nitpick: `random()` because it's *slightly* faster. `unsigned(random())` for best performance.

Eventually we'll need to have a better ranged random, because this is not very uniform.

framework/sandstone.cpp
```diff
@@ -1045,18 +1045,24 @@ namespace {
...
             shuffle_queue();
-        int result = queue[q_idx];
+
+        int next_cpu = queue[q_idx];
         if (++q_idx == queue.size())
             q_idx = 0;
-        return result;
+
+        if (!pin_to_logical_processor(LogicalProcessor(cpu_info[next_cpu].cpu_number))) {
+            log_debug("Failed to reschedule %d (%ld) to CPU %d", thread_num, pthread_self(), next_cpu);
```
thiagomacieira: (Pre-existing but please fix)
This should be a `log_warn`.

The formatter for `pthread_self()` should be `%tu` and you should cast it to `uintptr_t` so it will compile everywhere.
Linux: `typedef unsigned long int pthread_t;`
FreeBSD: `typedef struct  pthread                 *pthread_t;`
Winpthreads: `typedef uintptr_t pthread_t;`

You may want to put the actual pinning and warning in the base class, in a protected function, to avoid duplication.

framework/sandstone.cpp
```diff
@@ -1042,6 +1047,87 @@ int num_packages() {
...
+        // group->next_cpu.erase(group->next_cpu.begin() + thread_info_idx); // Should I remove the cpus as well?
+        group->pid.erase(group->pid.begin() + thread_info_idx);
+        lock.unlock();
+
+        group->barrier->arrive_and_drop();
+    }
+
+private:
+    struct GroupInfo {
+        std::unique_ptr<std::barrier<std::function<void()>>> barrier;
```
thiagomacieira, line 1115: Why is the barrier indirectly held in a pointer?

framework/sandstone.cpp
```diff
@@ -1042,6 +1047,87 @@ int num_packages() {
...
+            }
+        };
+
+        std::unique_lock lock(groups_mutex);
+
+        // Initialize groups on first run
+        if (groups.size() == 0) {
+            groups.reserve(num_cpus()/members_per_group);
+            for (int i=0; i<num_cpus(); i+=members_per_group) {
+                groups.push_back(new GroupInfo(members_per_group, on_completion));
```
thiagomacieira, line 1077: Any reason to use pointers here? If you can, make `groups` be a `vector<GroupInfo>` and here you can initialise each member with `emplace_back`:
```c++
    groups.emplace_back(members_per_group, on_completion);
```

framework/sandstone.cpp
```diff
@@ -1042,6 +1047,87 @@ int num_packages() {
...
+            }
+        }
+
+        // Fill thread info if not done already
+        int group_idx = thread_num / members_per_group;
+        GroupInfo *group = groups[group_idx];
+        int thread_info_idx = thread_num % members_per_group;
+        if (group->pid[thread_info_idx] == 0) {
+            group->pid[thread_info_idx] = getpid();
```
thiagomacieira, line 1086: `getpid()` is probably wrong here. Did you mean `gettid()` and thus use a  `tid_t`?

We've discussed (but haven't implemented) the framework knowing and storing the `tid_t` for each thread it starts, so you don't have to do it here. That way, the `pin_thread_to_logical_processor` function would take an internal CPU number instead of a `tid_t`. But let's think about that with a later patch.

framework/sandstone.cpp
```diff
@@ -1042,6 +1047,87 @@ int num_packages() {
...
+    void finish_reschedule() override {
+        // When thread finishes, unsubscribe it from barrier
+        // this avoid partners deadlocks
+        int group_idx = thread_num / members_per_group;
+        GroupInfo *group = groups[group_idx];
+
+        // Remove thread info from groups
+        std::unique_lock lock(groups_mutex);
+        int thread_info_idx = thread_num % members_per_group;
+        // group->next_cpu.erase(group->next_cpu.begin() + thread_info_idx); // Should I remove the cpus as well?
```
thiagomacieira, line 1106: Good question. We can leave the CPU there, so the remaining threads would run there. But we may not want to do that if the execution has failed for this thread... that's probably a bad device.

You can query if we have failed using
```c++
    sApp->test_thread_data(thread_number)->has_failed()
```

Explain the logic in a comment in the source, for future reference.

framework/sandstone.cpp
```diff
@@ -1018,6 +1019,10 @@ static uintptr_t thread_runner(int thread_number)
 
     cleanup.run_now();
 
+    // If rescheduling, do cleanup
+    if (sApp->device_schedule != nullptr)
+        sApp->device_schedule->finish_reschedule();
```
thiagomacieira, line 1024: This needs to be inside of the cleanup lambda, so it will be executed even if the test function throws. Put it after the `test_end` call, so it's after the `logging_mark_thread_failed()` and `ud2()` calls too.

thiagomacieira (changes requested):
Very good progress.

framework/sandstone.cpp
AthenasJimenez: Storing the TID was implemented some time ago in commit 726f637, but I kind find it useless for this because `sched_setaffinity`, on `pin_thread_to_logical_processor()`, expects a PID not a TID.

framework/sysdeps/linux/cpu_affinity.cpp
```diff
@@ -61,6 +61,27 @@ bool pin_to_logical_processor(LogicalProcessor n, const char *thread_name)
...
+    LogicalProcessorSetOps::setInArray({ cpu_set, size }, n);
+
+    if (sched_setaffinity(thread_id, sizeof(cpu_set), reinterpret_cast<cpu_set_t *>(cpu_set))) {
+        perror("sched_setaffinity");
+        return false;
+    }
+    return true;
+}
+
 bool pin_to_logical_processors(CpuRange range, const char *thread_name)
```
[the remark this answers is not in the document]
AthenasJimenez: Yes, added

framework/sandstone.cpp
AthenasJimenez: Oh got it!

framework/sandstone.cpp
AthenasJimenez: I was thinking that probably would be better to simply print the `tid_t tid` that each thread stores. In terms of debugging, I don't know which would be more useful

framework/sandstone.cpp
thiagomacieira: You should be able to hold the `std::barrier` directly and just initialise it in the constructor initialisation list:
```c++
        GroupInfo(int members_per_group, std::function<void()> on_completion)
          : barrier(members_per_group, std::move(on_completion))
        {
```

framework/sandstone.cpp
```diff
@@ -1042,6 +1047,90 @@ int num_packages() {
...
+            std::unique_lock lock(groups_mutex);
+            int g_idx = thread_num / members_per_group;
+
+            // Rotate cpus vector so reschedule group members to a different cpu
+            std::rotate(groups[g_idx].next_cpu.begin(), groups[g_idx].next_cpu.begin() + 1, groups[g_idx].next_cpu.end());
+            lock.unlock();
+
+            // Reschedule group members
+            for (int i=0; i<members_per_group; i++) {
+                pin_to_next_cpu(cpu_info[groups[g_idx].next_cpu[i]].cpu_number, groups[g_idx].tid[i]);
```
thiagomacieira, line 1065: Hint for brevity:
```c++
    GroupInfo &my_group = groups[g_idx];
```

thiagomacieira (changes requested):
Looking good, minor things remaining now.

framework/sandstone.cpp
AthenasJimenez: Much cleaner, thank you!

framework/sysdeps/linux/cpu_affinity.cpp
```diff
@@ -51,10 +56,10 @@ bool pin_to_logical_processor(LogicalProcessor n, const char *thread_name)
     size_t size = size_t(n) / ProcessorsPerWord + 1;
 
     Word cpu_set[size];         // -Wvla
-    memset(cpu_set, 0, sizeof(cpu_set));
+    memset(cpu_set, thread_id, sizeof(cpu_set));
```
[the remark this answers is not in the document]
AthenasJimenez: Ops... fixed!

AthenasJimenez:
Almost done, but I do still have pending some stuff:

- Figure out what to do when we don't have a `num_cpus()` divisible by `members_per_group`. Should I skip the test? or reorganize the groups to include available cpus? I.e. `num_cpus() = 7` and `members_per_group = 2`, so `barrier0` and `barrier1` will have 2 members, and `barrier2` will have 3 members.
- I'm still thinking about using a mutex per group to improve a little the performance, but not sure if it will be effective.
- Figure out if I can make the number of members configurable by user input, but this is low priority.

Regarding to this suggestion:
```
 GroupInfo(int members_per_group, std::function<void()> on_completion)
          : barrier(members_per_group, std::move(on_completion))
        {
```
It does not work. I guess is because `std::barrier` is non-copyable so compiler complains.

framework/sandstone.cpp
AthenasJimenez: I tried, but it fails to compile. I do guess is has to do with `std::barrier` being not assignable.

```
/usr/local/include/c++/14/bits/stl_construct.h:96:17: error: use of deleted function '{anonymous}::BarrierDeviceSchedule::GroupInfo::GroupInfo(const {anonymous}::BarrierDeviceSchedule::GroupInfo&)'
   96 |     -> decltype(::new((void*)0) _Tp(std::declval<_Args>()...))
      |                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../framework/sandstone_opts.cpp:366:12: note: '{anonymous}::BarrierDeviceSchedule::GroupInfo::GroupInfo(const {anonymous}::BarrierDeviceSchedule::GroupInfo&)' is implicitly deleted because the default definition would be ill-formed:
  366 |     struct GroupInfo {
      |            ^~~~~~~~~
../framework/sandstone_opts.cpp:366:12: error: use of deleted function 'std::barrier<_CompletionF>::barrier(const std::barrier<_CompletionF>&) [with _CompletionF = std::function<void()>]'
```

framework/topology.h
```diff
@@ -15,6 +15,7 @@
 #include <string>
 #include <vector>
 
+#include <gettid.h>
```
busykai, line 18: if you need `tid_t`, it's `<sys/types.h>`, not `<gettid.h>` (there's no standard header of that name).

AthenasJimenez: Good catch, I was meant to use `include "gettid.h"`. I'll fix it.

thiagomacieira: It can't be sys/types.h because that won't define the type on Windows. We need our gettid.h

thiagomacieira (dismissed):

thiagomacieira (approved):

PR #607, opened 2025-04-11
FW: remove dead code (--test-range option)

mcisowsk:
[ChangeLog][Framework] Removed `--test-range` option, which was not handled anyway.


lukasz-tuz (approved):
+1

PR #612, opened 2025-04-16
FW/ContextDump: Update dump_xsave() to print the APX Extended GPRs

thiagomacieira:
Untested.

### DIFF
--- a/framework/sandstone_context_dump.cpp
+++ b/framework/sandstone_context_dump.cpp
@@ -13,6 +13,7 @@
 #include "fp_vectors/Floats.h"
 
 #include <algorithm>
+#include <array>
 #include <limits>
 
 #ifdef __unix__
@@ -29,6 +30,8 @@
 #ifdef _WIN32
 #  include <windows.h>
 #endif
+
+using ApxState = std::array<int64_t, 16>;
 
 union xmmreg
 {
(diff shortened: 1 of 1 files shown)


framework/sandstone_context_dump.cpp
```diff
@@ -334,6 +337,27 @@ void dump_gprs(std::string &f, SandstoneMachineContext mc)
...
+        return;
+
+    auto base = reinterpret_cast<const uint8_t *>(state);
+    auto egprs = reinterpret_cast<const ApxState *>(base + offset);
+    char regname[] = "r16";
+    for (int64_t value : *egprs) {
+        print_gpr(f, regname, value);
+
+        // increment the register name
+        if (regname[2]++ == '9') {
```
busykai, line 353: Why do this if it's a fix arch register set? Let's just have an array of names (`static const *char egpr_names[16] = { "r16", ... };`). Or even counter and `sprintf()` to `char[3]`.

thiagomacieira: That would be far more expensive than this, requiring static memory for the register names and an array, or for the `sprintf` call, the latter of which would do exactly what we're doing here.

busykai (approved):

PR #614, opened 2025-04-24
FW/topology: Adding numa awareness to topology and restructuring topology with  hwloc

rsagashe03:


framework/sandstone.h
```diff
@@ -364,9 +364,11 @@ struct cpu_info {
     uint64_t ppin;          ///! Processor ID read from MSR
     uint64_t microcode;     ///! Microcode version read from /sys
     int cpu_number;         ///! Logical processor number as seen by OS
-    int thread_id;          ///! Topology info from APIC
-    int core_id;            ///! Topology info from APIC
-    int package_id;         ///! Topology info from APIC
+    int thread_id;          ///! Topology info from hwloc
+    int core_id;            ///! Topology info from hwloc
+    int package_id;         ///! Topology info from hwloc
+    uint32_t apic_id;       ///! APIC ID from hwloc
```
thiagomacieira, line 370: Use `int`. You are initialising this to -1.

.github/workflows/build.yaml
```diff
@@ -34,12 +34,12 @@ jobs:
             elfutils \
             file \
             git \
-            hwloc \
             libboost-dev \
             libeigen3-dev \
             libgtest-dev \
             libhwloc-dev \
             libssl-dev \
+            libudev-dev \
```
thiagomacieira, line 42: Needs an explanation in the commit message: what is failing to build?

framework/topology.cpp
```diff
@@ -656,19 +534,65 @@ static bool fill_ppin_sysfs(struct cpu_info *info)
     return false;
 }
 
+static unsigned  get_apic_id_from_cpuid() {
+    #ifdef _WIN32
+        unsigned int cpu_info[4];
+        __cpuid(cpu_info, 1);
+        return (cpu_info[1] >> 24) & 0xFF;
```
thiagomacieira, line 541: Doesn't make sense. `__cpuid` comes from GCC's cpuid.h header, which defines it as:
```c
#define __cpuid(level, a, b, c, d)					\
  __asm__ __volatile__ ("cpuid\n\t"					\
			: "=a" (a), "=b" (b), "=c" (c), "=d" (d)	\
			: "0" (level))
```

framework/topology.cpp
```diff
@@ -656,19 +534,65 @@ static bool fill_ppin_sysfs(struct cpu_info *info)
...
+        int can_bind = hwloc_topology_get_support(topology)->cpubind->set_thisproc_cpubind;
+        if (can_bind) {
+            hwloc_cpuset_t cpuset = hwloc_bitmap_dup(pu->cpuset);
+            hwloc_bitmap_singlify(cpuset);
+            if (hwloc_set_cpubind(topology, cpuset, HWLOC_CPUBIND_PROCESS) == 0) {
+                info->apic_id = get_apic_id_from_cpuid();
+                apic = 1;
+            }
+            hwloc_bitmap_free(cpuset);
+        }
```
thiagomacieira, line 579: We're already bound to the right PU. But I'd rather you left the APIC ID set to -1 here and added another function to fill the missing bits in.

framework/topology.cpp
```diff
@@ -965,7 +967,7 @@ static void init_topology_internal(const LogicalProcessorSet &enabled_cpus)
             }
 
             pin_to_logical_processor(lp);
-            try_detection<topo_impls>(&cpu_info[i]);
+            try_detection<topo_impls>(&cpu_info[i], topology);
```
thiagomacieira, line 893: Please run the hwloc topology detection before the pthread code and this lambda, because it doesn't need to run on the PU like this code does. Just add a topology function to the `topo_impls` list that returns true if the information has been set and therefore we shouldn't overwrite.

thiagomacieira (changes requested):

framework/topology.cpp
```diff
@@ -24,6 +25,7 @@
 #endif
 #if defined(_WIN32)
 #   include <windows.h>
+#   #include <intrin.h>
```
thiagomacieira, line 28: Rule of thumb: `<windows.h>` is always last.

framework/topology.cpp
```diff
@@ -728,24 +347,85 @@ static bool fill_ppin_sysfs(struct cpu_info *info)
...
+        }
+        if (cache && cache->attr) {
+            if (cache->attr->cache.type == HWLOC_OBJ_CACHE_UNIFIED || cache->attr->cache.type == HWLOC_OBJ_CACHE_DATA) {
+                info->cache[level].cache_data = cache->attr->cache.size;
+            }
+            if (cache->attr->cache.type == HWLOC_OBJ_CACHE_UNIFIED || cache->attr->cache.type == HWLOC_OBJ_CACHE_INSTRUCTION) {
+                info->cache[level].cache_instruction = cache->attr->cache.size;
+            }
+        }
+    }
```
thiagomacieira, line 407: Future thought is required here:
* cache topology is always the same inside of a given package (and usually is for the entire system)
* but we also need to know what other cores a given cache level is shared with

Nothing to do now, just leaving the comment for discussion later.

framework/topology.cpp
```diff
@@ -754,18 +434,30 @@ template <auto &fnArray> static bool try_detection(struct cpu_info *cpu)
...
+    // Check if the fn matches the regular signature
+    if constexpr (std::is_invocable_v<CallDetector, struct cpu_info *>) {
+        return fn(cpu);
+    }
+    // If it matches the specialized signature
+    else if constexpr (std::is_invocable_v<CallDetector, struct cpu_info *, hwloc_topology_t>) {
+        return fn(cpu, topology);
+    }
+    return false;
+}
```
thiagomacieira, line 447: Move the `call_detector` function above its use.

But better yet, remove the need for it. Just pass `hwloc_topology_t` to all functions; if they don't need it, then fine.

thiagomacieira (changes requested):

meson.build
```diff
@@ -252,10 +252,13 @@ if get_option('coverage') == true
     ]
 endif
 
+hwloc_dep = dependency('hwloc', static : true)
```
thiagomacieira, line 255: tests/meson.build has:
```
dep_static = get_option('dependency_link') == 'static'
...
eigen3_dep = dependency('eigen3', static : dep_static)
```

Please move the `dep_static` variable to the top-level meson and then use it here.

thiagomacieira (changes requested):

meson.build
thiagomacieira: Or move the variable assignment here.  I think that works too.

.github/workflows/build.yaml
```diff
@@ -37,12 +37,14 @@ jobs:
...
             libgtest-dev \
+            libhwloc-dev \
             libssl-dev \
+            libudev-dev \
             libzstd-dev \
             meson \
             ninja-build \
             python3-yaml \
-            zlib1g-dev
+            zlib1g-dev
```
busykai, line 47: Nitpick: do not add spaces.

framework/sandstone.h
```diff
@@ -364,9 +364,10 @@ struct cpu_info {
     uint64_t ppin;          ///! Processor ID read from MSR
     uint64_t microcode;     ///! Microcode version read from /sys
     int cpu_number;         ///! Logical processor number as seen by OS
-    int thread_id;          ///! Topology info from APIC
-    int core_id;            ///! Topology info from APIC
-    int package_id;         ///! Topology info from APIC
+    int thread_id;          ///! Topology info from hwloc
+    int core_id;            ///! Topology info from hwloc
+    int package_id;         ///! Topology info from hwloc
```
thiagomacieira, line 369: Can you add the module ID between core and package?

framework/topology.cpp
```diff
@@ -765,8 +826,8 @@ static const fill_ppin_func ppin_impls[] = { fill_ppin_sysfs, fill_ppin_msr };
 /* prefer sysfs, fallback to MSR. the latter is not reliable and may require
  * root. */
 static const fill_ucode_func ucode_impls[] = { fill_ucode_sysfs, fill_ucode_msr };
-/* prefer CPUID, fallback to sysfs. */
-static const fill_topo_func topo_impls[] = { fill_topo_cpuid, fill_topo_sysfs };
+/* prefer hwloc. */
```
thiagomacieira, line 442: It's the only option :-)

meson.build
```diff
@@ -252,10 +254,13 @@ if get_option('coverage') == true
     ]
 endif
 
+hwloc_dep = dependency('hwloc', static : dep_static)
+
 target_deps += [
     boost_dep,
     dl_dep,
     pthread_dep,
+    hwloc_dep,
```
thiagomacieira, line 263: This portion of the change should probably be moved a few commits up, so that `<hwloc.h>` is found and linked to.

Nitpick: `hwloc_dep` sorts before `pthread_dep`

thiagomacieira (changes requested):
This looks good, thanks! Please address the nitpicks. You can add the module ID that I'm requesting in a new PR.

PR #616, opened 2025-04-28
Framework: Add `TestObsolete` skip category.

dawid-sabat:
There are multiple reasons one may want to remove or make obsolete an existing content.
This commit purpose is to give means to developer to leave message for the user to explain what happened to the test.
We can not rely on release notes on this matter.
    
Making the obsolete test quality to `SKIP` is not an answer either as we do not leave any message when skipping due to test quality.
    
[ChangeLog] Added new runtime skip category called `TestObsolete`.

### DIFF
--- a/framework/sandstone.h
+++ b/framework/sandstone.h
@@ -94,6 +94,7 @@ typedef enum SkipCategory {
     UnknownSkipCategory,
     IgnoredMceCategory,
     RuntimeSkipCategory,
+    TestObsoleteSkipCategory,
     SelftestSkipCategory,
 } SkipCategory;
 
--- a/framework/logging.cpp
+++ b/framework/logging.cpp
@@ -475,6 +475,8 @@ static const char *char_to_skip_category(int val)
         return "OSResourceIssue";
     case IgnoredMceCategory:
         return "IgnoredMceCategory";
+    case TestObsoleteSkipCategory:
+        return "TestObsolete";
     }
 
     return "NO CATEGORY PRESENT";


framework/sandstone.h
```diff
@@ -95,6 +95,7 @@ typedef enum SkipCategory {
     IgnoredMceCategory,
     RuntimeSkipCategory,
     SelftestSkipCategory,
+    TestDepreciatedSkipCategory,
```
thiagomacieira, line 98: Please leave Selftest as the last.

thiagomacieira (changes requested):

thiagomacieira:
> We have `TEST_QUALITY_OPTIONAL`. Why do we need this skip mode?

Looking at the test change you want to use this for, I don't think Depreciated is the right term here. We're talking about tests that **cannot** be run any more for reasons outside of our control, not just something we suggest not running. Can we find a better term? @busykai

dawid-sabat:
> We have `TEST_QUALITY_OPTIONAL`. Why do we need this skip mode?

I find `TEST_QUALITY_OPTIONAL` a good idea if we want to keep the test content as it can be valuable sometimes. For example: we'd like to keep a test that is finding issues but you have to do additional preparations to device under test to run it.
To me this quality type says, that the test is valuable to someone, but not for everyone and it can't be ran in general test list.

In my case I've got tests that depend on 3rd party library. The library content moved forward and the maintainers removed the feature making the test content broken beyond repair. The only option is to remove broken code.
However, I'd like to leave message for the users which basically say: this test missing is not a bug, it has been removed deliberately.
That said, we can't tell the user, that these tests are somehow useful by making them `OPTIONAL` which would be a lie as they're gone and there is no point running them.
I think the removed content should do a runtime skip for some time to leave a message and then we can safely remove all remnants.

Let me think about a better skip category. What about `TestObsolete`?

thiagomacieira (approved):

PR #615, opened 2025-04-25
KVM: do not set AMX bits in XCRS0.

busykai:
Otherwise Linux kernel refuses to set them. This workarounds the issue of the initialization failing on systems with AMX hardware support.

PR #622, opened 2025-05-05
FW/KVM: AND the XCR0 bits with what KVM says it supports

thiagomacieira:
Instead of hardcoding the non-AMX bits, which also disabled the APX bit.

Amends commit a1fde0ee55c75d7a63fc62a874d73dce93cb6319 (PR #615).


thiagomacieira:
We're still getting the syscall failing, even with this change.

PR #624, opened 2025-05-09
'device_type' build option.

lukasz-tuz:
PR adds a `device_type` option to the OpenDCDiag. This option allows for
specifying the target device during the build (currently does nothing and defaults to
'cpu').

In addition to adding this option to the 'meson_option.txt' file, a variable with the
same name is added to the framework. In addition, the framework now checks if all
variables it expects to be set are indeed created -- including the `device_type`.

Additionally, build files were reformatted using a Meson linter.


framework/meson.build
```diff
@@ -224,15 +274,15 @@ framework_a = static_library(
...
     'unit-tests/sandstone_utils_tests.cpp',
-    'unit-tests/tests_dummy.cpp',
     'unit-tests/test_knob_tests.cpp',
+    'unit-tests/tests_dummy.cpp',
     'unit-tests/thermal_monitor_tests.cpp',
     'unit-tests/topology_tests.cpp',
+    'sandstone_chrono.cpp',
+    'sandstone_data.cpp',
+    'sandstone_utils.cpp',
+    'test_knobs.cpp',
```
thiagomacieira, line 326: What sorting order is this?

framework/meson.build
```diff
@@ -47,6 +47,12 @@ endif
 framework_config = configuration_data()
 framework_config.set10('SANDSTONE_STATIC', false) # sysdeps may override
 
+if device_type == 'cpu'
+    framework_config.set10('SANDSTONE_DEVICE_CPU', true)
+else
```
thiagomacieira, line 52: Probably `device_type == 'gpu'` else error

framework/sysdeps/linux/meson.build
```diff
@@ -1,17 +1,26 @@
...
+if device_type == 'cpu'
+    if (host_machine.cpu_family() == 'x86_64')
+        framework_files += files(
+            'kvm.c',
+            'interrupt_monitor.cpp',
+            'msr.c',
+        )
+    else
+        framework_files += files(
+            '../generic/kvm.c',
```
thiagomacieira, line 13: What requires even the generic version? Ideally we'd not compile anything with "kvm" in the name in the first place.

thiagomacieira (changes requested):
Minor change only. The rest looks good.

framework/sysdeps/linux/meson.build
thiagomacieira: Oh, it's already there.

framework/sandstone.cpp
```diff
@@ -777,7 +777,7 @@ void prepare_test(/*nonconst*/ struct test *test)
     if (test->groups)
         apply_group_inits(test);
 
-#ifdef SANDSTONE
+#if defined(SANDSTONE) and defined(SANDSTONE_DEVICE_CPU)
```
thiagomacieira, line 780: Note to self: find out if we still need this first macro.

thiagomacieira (approved):

PR #625, opened 2025-05-14
FW sandstone_p: stop using deprecated std::aligned_storage

thac0:
It is deprecated as of C++23. So, create a static storage buffer, with alignas(SandstoneApplication), to approximate its behavior.


thiagomacieira (approved):

busykai:
@thac0 please use branches so that the workflows could run.

PR #626, opened 2025-05-15
FW: Do allow test groups for obfuscate-test-name mode

thiagomacieira:
The old repository said

> When a replacement macro is defined, we'll suppress the test description and test groups too.

But I didn't write in the commit message why I wanted to suppress them. This is now causing problems for the tests that rely on the group's `group_init` doing something, like detecting whether certain functionality is possible or not (e.g., SELinux enforcing no SMC).


dawid-sabat:
I saw that issue with my setup, let me check this change and get back to you with the verification.

dawid-sabat (approved):
Looks good on my workstation build.

PR #627, opened 2025-05-16
FW: Fix empty messages in main threads when a test was skipped.

busykai:
Adjust expected size by the size of the init skip message for main threads which run init(). There's still a small potential issue with the same code potentially printing a different skip message, but it needs to be addressed separately and is, in fact, an indication of non-SMP machine or something badly broken.

Before:
```
  result: skip
  skip-category: CpuTopologyIssue
  skip-reason: '2 threads on the same core are required'
  time-at-end:   { elapsed:      3.999, now: !!timestamp '2025-05-16T18:39:26Z' }
  test-runtime: 2.678
  threads:
  - thread: main
    messages:
- test: mce_check
```
After:
```
  result: skip
  skip-category: CpuTopologyIssue
  skip-reason: '2 threads on the same core are required'
  time-at-end:   { elapsed:      4.000, now: !!timestamp '2025-05-16T18:38:44Z' }
  test-runtime: 2.515
- test: mce_check
```

### DIFF
--- a/framework/logging.cpp
+++ b/framework/logging.cpp
@@ -184,7 +184,7 @@ private:
     static void format_and_print_message(int fd, std::string_view level, std::string_view message);
     static void format_and_print_skip_reason(int fd, std::string_view message);
     int print_one_thread_messages(int fd, mmap_region r, int level);
-    void print_result_line();
+    void print_result_line(int &init_skip_message_bytes);
 
     enum TestHeaderTime { AtStart, OnFirstFail };
     static void print_tests_header(TestHeaderTime mode);
@@ -2358,7 +2358,7 @@ inline int YamlLogger::print_one_thread_messages(int fd, mmap_region r, int leve
     return lowest_level;
 }
 
-void YamlLogger::print_result_line()
+void YamlLogger::print_result_line(int &init_skip_message_bytes)
 {
     int loglevel = this->loglevel();
     if (loglevel == LOG_LEVEL_QUIET && file_log_fd != real_stdout_fd && sApp->shmem->verbosity < 1) {
@@ -2390,6 +2390,7 @@ void YamlLogger::print_result_line()
                 if (loglevel <= sApp->shmem->verbosity && file_log_fd != real_stdout_fd)
                     format_and_print_skip_reason(real_stdout_fd, message);
                 format_and_print_skip_reason(file_log_fd, message);
+                init_skip_message_bytes = init_skip_message.size() + 2;
             } else {
                 logging_printf(loglevel, "  skip-category: %s\n", "Unknown");
                 logging_printf(loglevel, "  skip-reason: %s\n",
@@ -2455,7 +2456,8 @@ void YamlLogger::print()
 {
     Duration test_duration = MonotonicTimePoint::clock::now() - sApp->current_test_starttime;
 
-    print_result_line();
+    int init_skip_message_bytes = 0;
+    print_result_line(init_skip_message_bytes);
     if (should_print_fail_info())
         logging_printf(LOG_LEVEL_QUIET, "%s", fail_info_details().c_str());
     logging_printf(LOG_LEVEL_VERBOSE(1), "  time-at-end:   %s\n", get_current_time().c_str());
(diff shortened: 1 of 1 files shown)


framework/logging.cpp
```diff
@@ -2483,17 +2485,20 @@ void YamlLogger::print()
     }
 
     // print the thread messages
-    auto doprint = [this](PerThreadData::Common *data, int i) {
+    auto doprint = [this, init_skip_message_bytes](PerThreadData::Common *data, int s_tid) {
         struct mmap_region r = maybe_mmap_log(data);
 
-        if (r.size == 0 && !data->has_failed() && sApp->shmem->verbosity < 3)
+        /* for main threads (negative ids) adjust the message size to account
```
thiagomacieira, line 2491: s/negative/non-positive/

framework/logging.cpp
```diff
@@ -2483,17 +2485,20 @@ void YamlLogger::print()
...
+         * for skip message. this is to avoid empty messages printed to the log.
+         */
+        if (r.size - (s_tid < 0 ? init_skip_message_bytes : 0) == 0 && !data->has_failed() && sApp->shmem->verbosity < 3)
```
thiagomacieira, line 2494: Keep the current line as-is, because if `r.size` is zero, there definitely isn't anything to send. Then add:
```c++
/* explaining comment here */
if (s_tid < 0 && r.size <= init_skip_message_bytes)
    return;
```

This will also prevent a problem in the case `init_skip_message_bytes > r.size` (see above) that would result in a non-zero value here.

framework/logging.cpp
```diff
@@ -2390,6 +2390,7 @@ void YamlLogger::print_result_line()
                 if (loglevel <= sApp->shmem->verbosity && file_log_fd != real_stdout_fd)
                     format_and_print_skip_reason(real_stdout_fd, message);
                 format_and_print_skip_reason(file_log_fd, message);
+                init_skip_message_bytes = init_skip_message.size() + 2;
```
thiagomacieira, line 2393: The skip message may not be the first message.

thiagomacieira (changes requested):
This won't work if the skip message isn't the first (e.g. a `log_debug` or `log_info` before skipping).

thiagomacieira:
> This won't work if the skip message isn't the first (e.g. a `log_debug` or `log_info` before skipping).

It's also corrupting the log, somehow, even in tests that shouldn't be doing that:
```
# - test: selftest_log_platform
#   details: { quality: production, description: "Logs platform messages (not related to the test)" }
#   state: { seed: 'LCG:1509968380', iteration: 0, retry: false }
#   time-at-start: { elapsed:    322.287, now: !!timestamp '2025-05-16T18:44:45Z' }
#   result: pass
#   time-at-end:   { elapsed:    346.446, now: !!timestamp '2025-05-16T18:44:45Z' }
#   test-runtime: 7.229
#   threads:
#   - thread: main
#     messages:
#     - { level: info, text: 'I> Platform issue: This is an informational platform message' }
# -beta option    - level: info
#       text: |1
#        I> This is a multiline info message from init function
#        Second line.
#     - { level: warning, text: 'W> This is a warning message from init function ending in newline' }
```

framework/logging.cpp
busykai: It does not matter. We do not see if it's first. We only adjust for size here.

busykai:
> > This won't work if the skip message isn't the first (e.g. a `log_debug` or `log_info` before skipping).
> 
> It's also corrupting the log, somehow, even in tests that shouldn't be doing that:
> 
> ```
> # - test: selftest_log_platform
> #   details: { quality: production, description: "Logs platform messages (not related to the test)" }
> #   state: { seed: 'LCG:1509968380', iteration: 0, retry: false }
> #   time-at-start: { elapsed:    322.287, now: !!timestamp '2025-05-16T18:44:45Z' }
> #   result: pass
> #   time-at-end:   { elapsed:    346.446, now: !!timestamp '2025-05-16T18:44:45Z' }
> #   test-runtime: 7.229
> #   threads:
> #   - thread: main
> #     messages:
> #     - { level: info, text: 'I> Platform issue: This is an informational platform message' }
> # -beta option    - level: info
> #       text: |1
> #        I> This is a multiline info message from init function
> #        Second line.
> #     - { level: warning, text: 'W> This is a warning message from init function ending in newline' }
> ```

That's only on Windows and I guess because it ends in newline. :(

framework/logging.cpp
busykai: Negative. `0` is a test thread.

framework/logging.cpp
busykai: It absolutely must print stuff that is not a skip message, right? Skip message will not be printed if the test was skipped, but all the rest will be. So to me the logic seems correct.

framework/logging.cpp
thiagomacieira: Ah, right.

framework/logging.cpp
```diff
@@ -2483,17 +2485,20 @@ void YamlLogger::print()
...
 
-        print_thread_header(file_log_fd, i, INT_MAX);
+        print_thread_header(file_log_fd, s_tid, INT_MAX);
         int lowest_level = print_one_thread_messages(file_log_fd, r, INT_MAX);
 
         if (lowest_level <= sApp->shmem->verbosity && file_log_fd != real_stdout_fd) {
-            print_thread_header(real_stdout_fd, i, sApp->shmem->verbosity);
+            print_thread_header(real_stdout_fd, s_tid, sApp->shmem->verbosity);
             print_one_thread_messages(real_stdout_fd, r, sApp->shmem->verbosity);
         }
```
thiagomacieira, line 2504: If you return when `r.size != 0`, then `munmap_and_truncate_log` doesn't truncate the log. I wonder if the existing second part of the conditional is a latent bug too.

thiagomacieira (changes requested):

thiagomacieira (approved):

PR #629, opened 2025-05-22
Fix SegFaults when building with `restricted-cmdline`

dawid-sabat:
This is a fix for side effect from https://github.com/opendcdiag/opendcdiag/pull/626. I reviewed that PR wrong building my test build without `restricted-cmdline` option. Building with it causes the application to Seg Fault.

### DIFF
--- a/framework/sandstone_test_groups.cpp
+++ b/framework/sandstone_test_groups.cpp
@@ -5,11 +5,8 @@
 
 #include "sandstone_p.h"
 
-#if SANDSTONE_RESTRICTED_CMDLINE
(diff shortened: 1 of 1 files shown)


busykai (approved):

thiagomacieira (commented):
I'd rather have #631

thiagomacieira:
> @thiagomacieira I think we can have both.

We have the ability to have both. But I do not *want* both, because that will cause the group name and descriptions to be added to the binary, where they can be decoded using `strings` or other such tools.

PR #633, opened 2025-05-22
FW: don't require test descriptions if we're not logging

thiagomacieira:
When we're not producing a log file, we don't need the descriptions. This is the operating mode for Intel DCDiag.

PR #632, opened 2025-05-22
FW: remove unreachable code in sandstone_tests.cpp in restricted cmdline

thiagomacieira:
Without a full command-line, one can't select tests to enable or disable, nor provide a new test list with the list of tests to enable. So help the compiler by removing the code paths.

PR #630, opened 2025-05-22
Meson: fix warning about using set01() with value 0

thiagomacieira:
Isn't that what the name implies?

```
framework/meson.build:80: DEPRECATION: configuration_data.set10 with number. The `set10` method should only be used with booleans
framework/meson.build:81: DEPRECATION: configuration_data.set10 with number. The `set10` method should only be used with booleans
framework/meson.build:82: DEPRECATION: configuration_data.set10 with number. The `set10` method should only be used with booleans
```

PR #635, opened 2025-05-27
Fix: Populate all_tests but not all_group_map.

dawid-sabat:
The change in https://github.com/opendcdiag/opendcdiag/pull/631 was too excessive. This resulted in test failures with exit error: `EX_DATAERR`

We hit it in `exec_mode_run`

See: `if (tests_to_run.size() != 1) return EX_DATAERR;`

The affected build was Windows build but not Linux.

### DIFF
--- a/framework/sandstone_tests.cpp
+++ b/framework/sandstone_tests.cpp
@@ -33,11 +33,10 @@ struct test mce_test = {
 void SandstoneTestSet::load_all_tests()
 {
     std::span<struct test> known_tests = cfg.is_selftest ? selftests : regular_tests;
-
-    if (!SandstoneConfig::RestrictedCommandLine) {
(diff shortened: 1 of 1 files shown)


dawid-sabat:
A summary of debug:

Looks like on windows we hit this code: `if (tests_to_run.size() != 1) return EX_DATAERR;` but no on Linux.
This is `exec_mode_run` .
I've got a fix that I'm about to push as it works on both Linux and Windows. Seems like we have to populate `all_tests` in `load_all_tests()`.
I don't understand why it was working on Linux though.

thiagomacieira:
> I ran it with a couple of other pending PRs, looks good:
> 
> #636 #635 #633 #632 #630

The one we really need in is #614. The rest are just nice-to-have cleanups.

PR #645, opened 2025-06-03
FW cpu_features: width now 128 bits, add future features

thac0:
In addition to some typos, ...

```
Ran out of feature bits, so widened the bitfields. Modified headers and
script to use standard `uint64_t` xor compiler builtin `unsigned
__int128`, depending on how many features are enumerated. Also needed
new CPUID leaf awareness for `0x1e`.

Feature bits added for:
  * SM3 and SM4 (hashes and block ciphers, respectively)
  * LASS (linear address space separation)
  * FRED (flexible return and event delivery)
  * LKGS (load into kernel GS)
  * LAM (linear address masking)
  * MSRLIST (supporting IA32_BARRIER_MSR)
  * MOVRS (MOV instructions with read-shared hint)
  * UIRET_UIF (flexible UIRET)
  * AVX-VNNI with
    * INT8, INT16 data types
  * AMX with
    * MOVRS, TRANSPOSE, TILE operations
    * COMPLEX, FP8, TF32 data types

Architectures added:
  * PNL
  * LNC, DMR
  * SKT, DKT

Architectures updated:
  * ARL, LNL
  * EMR, GNR, CWF

Referenced two documents (URLs below):
  * Intel(R) Architecture Instruction Set Extensions and Future Features
  * Intel(R) 64 and IA-32 Architectures Software Developer’s Manual

See: https://cdrdv2.intel.com/v1/dl/getContent/671368 (revision 057)
See: https://cdrdv2.intel.com/v1/dl/getContent/671200 (March 2025)
```

### DIFF
--- a/framework/simd.conf
+++ b/framework/simd.conf
@@ -81,11 +81,12 @@ avx512fp16		Leaf07_00EDX	    23	avx512f,f16c	# AVX512 16-bit Floating Point
 amx-tile		Leaf07_00EDX	    24		# Advanced Matrix Extensions Tile support
 amx-int8		Leaf07_00EDX	    25	amx-tile	# AMX Tile multiplication for Int8
 sha512			Leaf07_01EAX	    0	avx	# SHA-512 instructions
-#sm3			Leaf07_01EAX	    1	avx	# SM3 instructions
-#sm4			Leaf07_01EAX	    2	avx	# SM4 instructions
+sm3			Leaf07_01EAX	    1	avx	# SM3 instructions
+sm4			Leaf07_01EAX	    2	avx	# SM4 instructions
 raoint			Leaf07_01EAX	    3		# Remote Atomic Operations, Integer
 avxvnni			Leaf07_01EAX	    4	avx	# AVX (VEX-encoded) versions of the Vector Neural Network Instructions
 avx512bf16		Leaf07_01EAX	    5	avx512f	# AVX512 Brain Float16
+#lass			Leaf07_01EAX	    6		# Linear Address Space Seperation
 cmpccxadd		Leaf07_01EAX	    7		# CMPccXADD instructions
 #zlmovsb		Leaf07_01EAX	    10		# Zero-length MOVSB
 #fsrs			Leaf07_01EAX	    11		# Fast Short (REP?) STOSB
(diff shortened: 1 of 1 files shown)


thiagomacieira (changes requested):
Hmm... the OpenDCDiag isn't building with a test list - something we need to fix, so it didn't catch this error:
```
cpu_features.h:234:34: error: narrowing conversion of '0x4000030dcdffdd3ef77a' from 'cpu_feature_t' {aka '__int128 unsigned'} to 'uint64_t' {aka 'long long unsigned int'} [-Wnarrowing]
sandstone_test_lists.cpp:4180:43: note: in expansion of macro 'cpu_spr'
 4180 | static const BuiltinTestSet spr_test_set{ cpu_spr, "spr", test_set_0 };
      |                                           ^~~~~~~
```
The `BuiltinTestSet` structure needs to be updated too.

framework/scripts/x86simd_generate.pl
```diff
@@ -98,7 +100,9 @@
         { name => $name, depends => $depends, id => $id, bit => $bit, leaf => $function, comment => $comment };
         $feature_ids{$name} = $i;
         ++$i;
-        die("Too many features to fit a 64-bit integer") if $i > 64;
+
+        die("Too many features to fit a 128-bit integer") if $i > 128;
+        $feature_type = "unsigned __int128" if $i > 64;
```
busykai, line 105: why its on if instead of right away?

thiagomacieira (changes requested):
This is probably why:
https://github.com/opendcdiag/opendcdiag/blob/deafe2987b0086941eb9a152683011089b4162b1/framework/sandstone.cpp#L1955

The bit must be higher than 64, so we end up with 0 and don't skip the test.

Other places to fix:
https://github.com/opendcdiag/opendcdiag/blob/deafe2987b0086941eb9a152683011089b4162b1/framework/sandstone.cpp#L1259

framework/simd.conf
```diff
@@ -81,34 +81,48 @@ avx512fp16		Leaf07_00EDX	    23	avx512f,f16c	# AVX512 16-bit Floating Point
...
 avx512bf16		Leaf07_01EAX	    5	avx512f	# AVX512 Brain Float16
+lass			Leaf07_01EAX	    6		# Linear Address Space Seperation
 cmpccxadd		Leaf07_01EAX	    7		# CMPccXADD instructions
 #zlmovsb		Leaf07_01EAX	    10		# Zero-length MOVSB
 #fsrs			Leaf07_01EAX	    11		# Fast Short (REP?) STOSB
 #fsrc			Leaf07_01EAX	    12		# Fast Short (REP?) CMPSB, SCASB
-#fred			Leaf07_01EAX	    17		# Flexible Return and Event Delivery
-#lkgs			Leaf07_01EAX	    18		# Load into Kernel GS
+fred			Leaf07_01EAX	    17		# Flexible Return and Event Delivery
+lkgs			Leaf07_01EAX	    18		# Load into Kernel GS
```
thiagomacieira, line 95: Can we make use of these and `lass` in userspace (`lam` makes sense)? If not, leave commented out.

framework/simd.conf
```diff
@@ -81,34 +81,48 @@ avx512fp16		Leaf07_00EDX	    23	avx512f,f16c	# AVX512 16-bit Floating Point
...
 amx-fp16		Leaf07_01EAX	    21	amx-tile	# AMX Tile multiplication in FP16
 avxifma			Leaf07_01EAX	    23	avx	# AVX-IFMA instructions
-#lam			Leaf07_01EAX	    26		# Linear Address Masking
+lam			Leaf07_01EAX	    26		# Linear Address Masking
+msrlist			Leaf07_01EAX	    27		# RDMSRLIST, WRMSRLIST, IA32_BARRIER_MSR
```
thiagomacieira, line 99: `user_msr` makes sense. `msrlist` does not - it's ring 0.

framework/simd.conf
```diff
@@ -81,34 +81,48 @@ avx512fp16		Leaf07_00EDX	    23	avx512f,f16c	# AVX512 16-bit Floating Point
...
+avxvnniint8		Leaf07_01EDX	    4	avxvnni	# AVX Vector Neural Network Instructions, Int8
 avxneconvert		Leaf07_01EDX	    5	avx	# AVX Non-Exception BF16/FP16/FP32 Conversion instructions
-#amx-complex		Leaf07_01EDX	    8	amx-tile	# AMX Complex Matrix multiplication
-avxvnniint16		Leaf07_01EDX	    10	avx	# AVX Vector Neural Network Instruction, Int16
+amx-complex		Leaf07_01EDX	    8	amx-tile	# AMX Tile multiplication for complex matrices
+avxvnniint16		Leaf07_01EDX	    10	avxvnni	# AVX Vector Neural Network Instruction, Int16
 #prefetchiti		Leaf07_01EDX	    14		# PREFETCHIT0/1 instructions
+user_msr		Leaf07_01EDX	    15		# Userspace MSR support (URDMSR and UWRMSR)
 #avx10			Leaf07_01EDX	    19	avx	# AVX10 Converged Vector ISA
+uiret			Leaf07_01EDX	    17		# User-Interrupt Return
```
thiagomacieira, line 109: In theory, we could use UIRET because it can be run from any CPL. But I don't think Linux has enabled support for it, so leave it commented out.

framework/simd.conf
```diff
@@ -168,12 +182,15 @@ arch=WLC	SNC	shstk,movdiri,movdir64b,ibt,keylocker		# avx512vp2intersect
 arch=GLC	WLC	avx512bf16,avxvnni,cldemote,waitpkg,serialize,uintr	# pks,tsxldtrk
 arch=RPC	GLC
 arch=RWC	RPC	avx10,prefetchiti
+arch=LNC	RWC	cmpccxadd,lam,avxifma,avxneconvert,avxvnniint8,sha512,sm3,sm4
```
thiagomacieira, line 185: Move the crypto instructions to a comment: they may be optional in some SKUs (see pclmul, aesni and shani above). The script currently has no way to declare optional content.

Repeats below.

thiagomacieira (changes requested):

framework/simd.conf
thac0: Primarily kernel-space, the deeper I dig. I'll leave them commented.

busykai:
Please make sure to squash before or when merging. The original commit should have been split in two (one for doubling the size, the other for adding features). As it is, the second commit is not needed.

PR #647, opened 2025-06-04
FW/SSL: include the engines and providers in the header

thiagomacieira:
One important engine or provider is the one for Intel's QAT, so it's important to know if it is being enabled. On my regular laptop, this now prints:
```
os: Linux 6.14.6-1-default
runtime: glibc 2.41
openssl: { version: OpenSSL 3.5.0 8 Apr 2025, providers: ['default 3.5.0','sandstone-rand'], engines: ['dynamic','rdrand'] }
```

### DIFF
--- a/bats/sanity-check/20-selftests.bats
+++ b/bats/sanity-check/20-selftests.bats
@@ -296,6 +296,7 @@ tap_negative_check() {
         os=`wine cmd /c ver | sed -n "s/\r$//;s/.*Windows /Windows v/p"`
     fi
     [[ "${yamldump[/os]}" = "$os"* ]]
+    test_yaml_regexp "/openssl" "{'version':.*|None"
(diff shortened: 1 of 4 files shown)


busykai (dismissed):

AthenasJimenez (approved):
I think we should omit the engines, they have been deprecated since 2021. At some point the APIs won't work anymore and we may need to remove the code either way.

Otherwise, LGTM.

AthenasJimenez:
> > I think we should omit the engines, they have been deprecated since 2021. At some point the APIs won't work anymore and we may need to remove the code either way.
> > Otherwise, LGTM.
> 
> I disagree. If the user's system has engines (and QAT can be built as an engine), then they can affect the result.

They will need to migrate at some point as well. Anyway, I'm ok keeping the `engines` part until OpenSSL removes it.

busykai (approved):

PR #648, opened 2025-06-04
FW/Topology: fix extracting the module_id when Linux doesn't know either

thiagomacieira:
The topology/cluster_id file is always present (on x86), but it may contain -1. See `parse_topology()`[1] for where it gets first initialized. Includes test to ensure this works.

[1] https://codebrowser.dev/linux/linux/arch/x86/kernel/cpu/topology_common.c.html#parse_topology

### DIFF
--- a/bats/sanity-check/20-selftests.bats
+++ b/bats/sanity-check/20-selftests.bats
@@ -306,6 +306,8 @@ tap_negative_check() {
         fi
         test_yaml_numeric "/cpu-info/$i/logical" 'value >= 0'
         test_yaml_numeric "/cpu-info/$i/package" 'value >= 0'
+        test_yaml_numeric "/cpu-info/$i/numa_node" 'value >= -1'
+        test_yaml_numeric "/cpu-info/$i/module" 'value >= 0'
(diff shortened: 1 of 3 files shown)


busykai (approved):

bats/sanity-check/20-selftests.bats
```diff
@@ -306,6 +306,8 @@ tap_negative_check() {
         fi
         test_yaml_numeric "/cpu-info/$i/logical" 'value >= 0'
         test_yaml_numeric "/cpu-info/$i/package" 'value >= 0'
+        test_yaml_numeric "/cpu-info/$i/numa_node" 'value >= -1'
+        test_yaml_numeric "/cpu-info/$i/module" 'value >= 0'
```
busykai, line 310: the CI is right. this should be >= -1.

thiagomacieira: No, that's a bug in the code. Linux is indeed telling us that the `cluster_id` is 65535 but that's a bad solution.

busykai (approved):

PR #652, opened 2025-06-05
FW/Selftests: add a data-comparison test that randomly fails

thiagomacieira:
For didactic purposes.


dawid-sabat:
Do you have any practical use in mind?

thiagomacieira:
> Do you have any practical use in mind?

I wrote it for a demonstration of what an SDE event would look like, for a presentation of OpenDCDiag. I thought it would be useful for others.

dawid-sabat (approved):

PR #659, opened 2025-06-09
FW/Win32: Add strerror function.

AthenasJimenez:
The function returns a last-error code as a string.


thiagomacieira (approved):
Can you also move `win32_perror` into the new .cpp and make it use the newly-created function?

PR #674, opened 2025-06-24
FW: permit whitespaces in the test list files when specifying duration

### LINKED ISSUE #672: Spaces are not permitted in the test list files when specifying duration
With the test list mgmt refactoring, the test duration specification is no longer allowing spaces. E.g. before it was possible to specify

```
eigen_svd: 4000
```
but the current implementation will reject it while allowing:

```
eigen_svd:4000
```

Space characters should be allowed in the test list files.

gibasm:
In test list files, whitespaces between terminator and duration are producing syntax error as described in #672.
These changes fix the issue by skipping the whitespaces before encountering token contents.

Fixes #672

### DIFF
--- a/bats/sanity-check/20-selftests.bats
+++ b/bats/sanity-check/20-selftests.bats
@@ -1030,7 +1030,8 @@ test_list_file() {
 }
 
 @test "--test-list-file with duration" {
-    test_list_file selftest_pass:default selftest_timedpass:250 selftest_timedpass:10
+    test_list_file selftest_pass:default selftest_timedpass:250 selftest_timedpass:10 \
+        '' 'selftest_timedpass: 10' '' "$(printf "selftest_timedpass:\t10")"
 }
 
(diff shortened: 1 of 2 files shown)


bats/sanity-check/20-selftests.bats
```diff
@@ -1030,7 +1030,8 @@ test_list_file() {
 }
 
 @test "--test-list-file with duration" {
-    test_list_file selftest_pass:default selftest_timedpass:250 selftest_timedpass:10
+    test_list_file selftest_pass:default selftest_timedpass:250 selftest_timedpass:10 \
+        '' 'selftest_timedpass:   250'
```
dawid-sabat, line 1034: Are we supposed to accept `\t` as well? If you put this change into a context `is_ignored` returns `true` with ` ` and `\t`.
Can we add a selftest for that?

gibasm: Done

busykai (approved):

PR #675, opened 2025-06-24
Framework extended with device assignment, it allows to not start

jposwiata:
always-skip threads and reduce the number of logged messages. WIP


jposwiata:
This was a proposal to skip execution on "unwanted" CPUs without skip messages.
As the solution doesn't match agreed one, lets discard it.

PR #680, opened 2025-07-02
CI: enable building on ARM

thiagomacieira:
This PR includes a number of simple fixes and disables a number of tests that weren't passing (and I won't investigate).

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -14,13 +14,14 @@ permissions:
 
 jobs:
   Linux:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


bats/sanity-check/20-selftests.bats
```diff
@@ -1330,6 +1330,9 @@ selftest_crash_common() {
 }
 
 @test "selftest_sigsegv_instruction" {
+    if [[ "`uname -m`" != x86_64 ]]; then
+        skip "On some architectures, this causes signals other than SIGSEGV"
```
busykai, line 1334: Maybe "Only tested on x86_64 since the test may signals other than SIGSEGV".

busykai (approved):

PR #682, opened 2025-07-02
Docs: add a schema file for our YAML output

thiagomacieira:
I've manually verified our output passes its check. Running it in the CI is non-obvious, because [jsonschema](https://github.com/python- jsonschema/jsonschema) no longer has a working CLI and I can't make heads or tails of how to report errors with its Python API.


busykai (approved):
We need to give it a use otherwise it'll become stale pretty quick.

PR #689, opened 2025-07-04
CPU-separation: move `cpu_info` and start moving topology to separate files

mcisowsk:


thiagomacieira (approved):

thiagomacieira (dismissed):

framework/devicedeps/cpu/topology_cpu.h
```diff
@@ -0,0 +1,138 @@
...
+    Data &operator=(Data &&) = default;
+
+    std::vector<Package> packages;
+    std::vector<Topology::Thread> all_threads;
+};
+
+bool pin_thread_to_logical_processor(LogicalProcessor n, tid_t thread_id, const char *thread_name = nullptr);
+
+// CPU reschedule
+void reschedule();
```
AthenasJimenez, line 81: This change breaks `reschedule()` calls done inside the tests.

```
../tests/zlib/test.c: In function 'zlib_run_common':
../tests/zlib/test.c:146:9: error: implicit declaration of function 'reschedule' [-Wimplicit-function-declaration]
  146 |         reschedule();
      |         ^~~~~~~~~~
ninja: build stopped: subcommand failed.
```

AthenasJimenez (changes requested):

framework/devicedeps/cpu/topology_cpu.h
mcisowsk: seems like we don't do any calls to `reschedule()` inside tests? Would we want to?

framework/topology.h
```diff
@@ -21,100 +21,37 @@
...
-    {
+    static Word constWordForInArray(std::span<const Word> array, LogicalProcessor n) {
         int idx = int(n) / ProcessorsPerWord;
         return idx < array.size() ? array[idx] : 0;
     }
 };
 
-class LogicalProcessorSet : private LogicalProcessorSetOps
-{
+class LogicalProcessorSet : private LogicalProcessorSetOps {
```
thiagomacieira, line 54: Please don't change the coding style. The opening brace for top-level classes and functions is always in alone in a line or in the same line as the closing brace.

thiagomacieira (changes requested):

framework/topology.h
mcisowsk: since I've noticed the style being mixed, I took the liberty to choose the one I like more. I will switch fully to this second variant of braces then.

framework/devicedeps/cpu/topology_cpu.h
thiagomacieira: Move it back to device_cpu.h and make sure it's inside of the `__cplusplus` `extern "C"` block

thiagomacieira (changes requested):
Otherwise looks good.

thiagomacieira (dismissed):

AthenasJimenez (dismissed):

busykai (commented):
Can we call the directory `device` instead of `devicedeps`? If it is named after `sysdeps`, it's not quite right analogy.

framework/meson.build
```diff
@@ -55,10 +57,16 @@ else
     error('Unknown device type: ' + device_type)
 endif
 
-sysdeps_dir = 'sysdeps/' + target_machine.system()
+sysdeps_dir = 'sysdeps' / target_machine.system()
+devicedeps_dir = 'devicedeps' / device_type
+devicedeps_sys_dir = ''
+if fs.is_dir(devicedeps_dir / target_machine.system())
```
busykai, line 63: Is this needed for GPU? Cause there's no need for this at the moment.

framework/meson.build
```diff
@@ -55,10 +57,16 @@ else
...
+    devicedeps_sys_dir = devicedeps_dir / target_machine.system()
+endif
 
 if target_machine.system() == 'cygwin'
     sysdeps_dir = 'sysdeps/windows'
+    devicedeps_sys_dir = 'devicedeps' / device_type / 'windows'
```
busykai, line 69: Nor for this one. I'd prefer to not add infrastructure for hypothetical cases.

mcisowsk: ok, removed

framework/meson.build
mcisowsk: yes, it will come with the following PRs, but changes in meson file are already there

PR #694, opened 2025-07-07
FW/Topology: reorganize code to use a TopologyDetector class

thiagomacieira:
This will allow us to do some pre-processing before launching the detection and save the topology CPUID in the heap instead of in static (.bss) variables.

### DIFF
--- a/framework/topology.cpp
+++ b/framework/topology.cpp
@@ -27,6 +27,40 @@
 #   include <windows.h>
 #endif
 
+namespace {
+struct TopologyDetector
+{
+    TopologyDetector();
+    void detect(const LogicalProcessorSet &enabled_cpus);
+    void sort();
+
+private:
+    bool detect_via_cpuid(Topology::Thread *info);
+    bool detect_via_os(Topology::Thread *info);
+    bool detect_numa();
+
+#ifdef __x86_64__
+    bool detect_cache_via_cpuid(Topology::Thread *info, uint32_t *max_cpus_sharing_l2);
+    bool detect_topology_via_cpuid(Topology::Thread *info);
+#endif
+
+    void detect_family_via_cpuid();
+    bool detect_ppin_via_msr(Topology::Thread *info);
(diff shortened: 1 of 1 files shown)


framework/topology.cpp
```diff
@@ -1175,27 +1230,31 @@ static void init_topology_internal(const LogicalProcessorSet &enabled_cpus)
...
+        pthread_join(detection_thread, nullptr);
+    } else {
+        // Use OS-level detection only (no pinning, no threads).
+        for (Topology::Thread &cpu : std::span(cpu_info, count)) {
+            detect_via_os(&cpu);
+        }
+    }
 
-    fill_numa();
+    detect_numa();
```
AthenasJimenez, line 1257: Maybe a corner case, but there is no definition for `detect_numa()` when OS is different from Linux and Windows.

framework/topology.cpp
```diff
@@ -27,6 +27,72 @@
...
+struct TopologyDetector
+{
+    TopologyDetector();
+    void detect(const LogicalProcessorSet &enabled_cpus);
+    void sort();
+
+private:
+    bool setup_cpuid_detection();
+    bool detect_via_cpuid(Topology::Thread *info);
+    bool detect_via_os(Topology::Thread *info);
```
AthenasJimenez, line 40: I find confusing the names `detect_via_os()` and `detect_topology_via_os()`, like if they do the same thing. Either way, this is only a comment and no changes are requested.

AthenasJimenez (changes requested):

framework/topology.cpp
thiagomacieira: I haven't compiled FreeBSD yet. Thanks for noticing.

framework/topology.cpp
thiagomacieira: Not necessarily. "detect_via_os" is "detect everything", not just the topology (see other functions).

AthenasJimenez (approved):

PR #699, opened 2025-07-15
CPU-separation: move cpu-related `sysdeps` to separate directory

mcisowsk:
- parent: https://github.com/opendcdiag/opendcdiag/pull/698


busykai:
> > I think the entire current `sysdeps` tree needs to be moved to `device/cpu/sysdeps`, not to `device/cpu` (one level lower). So that you don't get directories like `device/darwin`.
> 
> The majority of `sysdeps/unix` is actually _not_ specific to testing CPU:
> 
> ```
> framework_files += files(
>     'child_debug.cpp',
>     'resource.cpp',
>     'signals.cpp',
>     'splitlock_detect.c',
>     'stacksize.cpp',
>     'tmpfile.c',
> )
> ```
> 
> The Windows counterparts of the above are likewise not.

I misspoke. Whatever is being moved needs to be moved to sysdeps inside `device/(cpu|gpu)`.

mcisowsk:
> > > I think the entire current `sysdeps` tree needs to be moved to `device/cpu/sysdeps`, not to `device/cpu` (one level lower). So that you don't get directories like `device/darwin`.
> > 
> > 
> > The majority of `sysdeps/unix` is actually _not_ specific to testing CPU:
> > ```
> > framework_files += files(
> >     'child_debug.cpp',
> >     'resource.cpp',
> >     'signals.cpp',
> >     'splitlock_detect.c',
> >     'stacksize.cpp',
> >     'tmpfile.c',
> > )
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The Windows counterparts of the above are likewise not.
> 
> I misspoke. Whatever is being moved needs to be moved to sysdeps inside `device/(cpu|gpu)`.

Got it. Makes sense.

busykai (approved):

PR #720, opened 2025-07-22
CPU-separation: move selftests

mcisowsk:
- parent: #719 

Let's have a discussion what else can be moved as "CPU-only" selftest


thiagomacieira:
> * parent: [CPU-separation: move unit-tests targeting cpu #719](https://github.com/opendcdiag/opendcdiag/pull/719)
> 
> Let's have a discussion what else can be moved as "CPU-only" selftest

This one I think went too far. Too many of the selftests are just testing the framework itself. Even the ones about crashing are useful for other types of devices, in case our code or the device's runtime libraries crash (something we have seen). Can you identify which selftests do *not* make sense for a non-CPU type of device? We talked about thread affinity and pinning, for example.

Finally, unlike regular tests, the `selftests_array` doesn't use a linker trick, so all tests must be defined inside of the same .cpp. We can change that, but it might be simpler — at least initially — to simply `#if` tests that aren't applicable. We do that for OS-specific tests (KVM and not), and architecture-specific ones.

mcisowsk:
> > * parent: [CPU-separation: move unit-tests targeting cpu #719](https://github.com/opendcdiag/opendcdiag/pull/719)
> > 
> > Let's have a discussion what else can be moved as "CPU-only" selftest
> 
> This one I think went too far. Too many of the selftests are just testing the framework itself. Even the ones about crashing are useful for other types of devices, in case our code or the device's runtime libraries crash (something we have seen). Can you identify which selftests do _not_ make sense for a non-CPU type of device? We talked about thread affinity and pinning, for example.
> 
> Finally, unlike regular tests, the `selftests_array` doesn't use a linker trick, so all tests must be defined inside of the same .cpp. We can change that, but it might be simpler — at least initially — to simply `#if` tests that aren't applicable. We do that for OS-specific tests (KVM and not), and architecture-specific ones.

KVM and testing individual sockets are such an example. We can stay with `#ifdef` approach though, no problem. Especially, since there is already many ifdefs in this file.

thiagomacieira:
> KVM and testing individual sockets are such an example. We can stay with `#ifdef` approach though, no problem. Especially, since there is already many ifdefs in this file.

Thanks. Let's revisit separating if and when this becomes too cumbersome.

PR #721, opened 2025-07-22
CPU-separation: move `main.cpp` and `cpuid_internal.cpp`

mcisowsk:


busykai (dismissed):

framework/meson.build
```diff
@@ -270,13 +271,6 @@ configure_file(
...
-premain_file = files()
-if device_type == 'cpu'
-    premain_file += files(
-        'main.cpp',
-    )
-endif
-
 framework_main_a = static_library(
     'framework_main',
     premain_file,
```
thiagomacieira, line 276: I wonder if we should move the entire `framework_main_a` library

thiagomacieira (changes requested):
I wonder if we should move the entire `framework_main_a` library. Meson is already complaining about a library with no content, as in:
```
WARNING: Build target tests_hsw has no sources. This was never supposed to be allowed but did because of a bug, support will be removed in a future release of Meson
```

So my suggestion is that instead of having a file list, you instead have an object list. Above, declare:
```
extra_objects = []
```
and below:
```
    objects: extra_objects,
```

Then move to cpu/meson.build:
```
framework_main_a = static_library(
    'framework_main',
    premain_file,
    build_by_default: false,
    include_directories: [
        framework_incdir,
    ],
    cpp_args: [
        debug_c_flags,
        march_generic_flags,
        default_cpp_flags,
        default_cpp_warn,
    ],
)

extra_objects += framework_main_a.extract_all_objects(recursive: false)
```

linux/meson.build may also want to put its `glibc_malloc.o` in this list.

framework/meson.build
mcisowsk: it's a very good idea

PR #725, opened 2025-07-23
CPU-separation: use `device_info` alias in common code

mcisowsk:
Other devices would introduce their own struct representing a device (GPU)

### DIFF
--- a/framework/logging.cpp
+++ b/framework/logging.cpp
@@ -1820,7 +1820,7 @@ void KeyValuePairLogger::print_thread_header(int fd, int cpu, const char *prefix
         return;
     }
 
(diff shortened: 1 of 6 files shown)


framework/logging.cpp
```diff
@@ -1820,7 +1820,7 @@ void KeyValuePairLogger::print_thread_header(int fd, int cpu, const char *prefix
         return;
     }
 
-    struct cpu_info *info = cpu_info + cpu;
+    device_info *info = cpu_info + cpu;
```
thiagomacieira, line 1823: We'll eventually need to rename the `cpu_info` global variable too.

thiagomacieira (approved):

PR #726, opened 2025-07-23
CPU-separation: use `num_devices()` interface instead of `num_cpus()`

mcisowsk:


mcisowsk:
migrated to https://github.com/opendcdiag/opendcdiag/pull/754

PR #727, opened 2025-07-23
CPU-separation: move `slice_plan_init` as an interface

mcisowsk:
Highly dependent on the CPU topology.
Other devices would not need slice plan to be that sophisticated.


thiagomacieira (changes requested):
Looks good, but can you move it?

thiagomacieira (approved):

PR #729, opened 2025-07-23
CPU-separation: move `init_topology` as an interface

mcisowsk:


framework/sandstone.cpp
```diff
@@ -2750,11 +2734,10 @@ int main(int argc, char **argv)
...
 
-    {
-        LogicalProcessorSet enabled_cpus = init_cpus();
-        init_shmem();
-        init_topology(std::move(enabled_cpus));
-    }
+    // this order is required
+    init_num_devices();
+    init_shmem(); // this uses num_devices
+    init_topology(); // this uses shmem
```
thiagomacieira, line 2740: Can you say more why you've removed the variable from here? I was expecting something like:
```c++
    DeviceSet enabled_devices = detect_devices();
    init_shmem(enabled_devices.count());
    setup_devices(std::move(enabled_devices));
```
and `restrict_topology` below becomes `restrict_devices`

mcisowsk: that's one of the things that arose during POC GPU implementation. I don't think `init_topology()` would need any externally provided parameter in case of the GPU impl. The `LogicalProcessorSet` passed from `init_cpus()` is CPU-specific implementation detail, that's why I've hidden it. 
So, `init_num_devices()` initializes `sApp->thread_count`, and `init_shmem()` needs `sApp->thread_count`, and nothing else

mcisowsk: then, which "state" is used by `init_topology()`, is implementation detail

mcisowsk: I also like the "topology" nomenclature, so I kept the original function names

PR #730, opened 2025-07-23
CPU-separation: add `deviceset` alias for `cpuset`, use the former in code

mcisowsk:


framework/sandstone_opts.cpp
```diff
@@ -227,7 +228,7 @@ Common command-line options are:
      limit to the number of loop iterations.  This special value can be
      used to disable test fracturing.  When specified tests will not be
      fractured and their execution will be time limited.
- --cpuset=<set>
+ --cpuset=<set>, --deviceset=<set>
```
mcisowsk, line 231: not sure whether we should edit this description at this point?

PR #737, opened 2025-07-24
CPU-separation: decouple `analyze_test_failures`s dependency on CPU topology

mcisowsk:
Part of it's implementation can be in a common code, the rest should be
    implementation-specific for CPU/GPU as it relies heavy on topology.


framework/sandstone.cpp
```diff
@@ -1965,120 +1965,35 @@ static TestResult run_one_test_once(const struct test *test)
...
-        // valid topology, we can do more a interesting analysis
-        logging_printf(LOG_LEVEL_VERBOSE(1), "# Topology analysis:\n");
-        for (size_t p = 0; p < topology.packages.size(); ++p) {
-            Topology::Package *pkg = &topology.packages[p];
-            for (size_t c = 0; c < pkg->cores.size(); ++c) {
-                Topology::Core *core = &pkg->cores[c];
-                bool all_threads_failed_once = true;
-                bool all_threads_failed_equally = true;
-                int nthreads = 0;
-                fail_pattern = 0;
```
mcisowsk, line 2044: `fail_pattern` is re-initialized to 0 in each loop, so I do not pass it to `analyze_test_failures_for_topology`

framework/device/cpu/topology.cpp
```diff
@@ -1562,6 +1562,97 @@ void restrict_topology(DeviceRange range)
...
+        // valid topology, we can do more a interesting analysis
+        logging_printf(LOG_LEVEL_VERBOSE(1), "# Topology analysis:\n");
+        for (size_t p = 0; p < topology.packages.size(); ++p) {
+            Topology::Package *pkg = &topology.packages[p];
+            for (size_t c = 0; c < pkg->cores.size(); ++c) {
+                Topology::Core *core = &pkg->cores[c];
+                bool all_threads_failed_once = true;
+                bool all_threads_failed_equally = true;
+                int nthreads = 0;
+                uint64_t fail_pattern = 0;
```
mcisowsk, line 1620: defined again

framework/sandstone_p.h
```diff
@@ -357,7 +357,6 @@ struct SandstoneApplication : public InterruptMonitor, public test_the_test_data
         std::array<Slices, 2> plans;
     };
 
-    using PerThreadFailures = std::vector<__uint128_t>;
```
thiagomacieira, line 360: If possible, keep an alias here to wherever this ends up, to facilitate updating the internal code that currently uses this type.

framework/sandstone.cpp
```diff
@@ -1965,120 +1965,35 @@ static TestResult run_one_test_once(const struct test *test)
...
+    if (all_threads_failed_equally && nfailures == num_cpus()) {
+        logging_printf(LOG_LEVEL_VERBOSE(1), "# All threads failed equally. This is highly unlikely (SW bug?)\n");
         return;
     }
 
-    Topology topology  = Topology::topology();
-    if (!topology.isValid()) {
-        // can't use this information
-        if (all_cpus_failed_equally)
-            logging_printf(LOG_LEVEL_VERBOSE(1), "# All failing CPUs failed equally.\n");
```
thiagomacieira, line 1996: This was dead code.

framework/device/cpu/topology.cpp
```diff
@@ -1562,6 +1562,97 @@ void restrict_topology(DeviceRange range)
         topo = build_topology();
 }
 
+void analyze_test_failures_for_topology(const struct test *test,
+        const PerThreadFailures &per_thread_failures, bool all_threads_failed_equally)
+{
+    Topology topology  = Topology::topology();
+    if (!topology.isValid()) {
+        // can't use this information
+        if (all_threads_failed_equally)
```
thiagomacieira, line 1571: `all_threads_failed_equally = true` is impossible, because the caller never reaches the point of calling us with that value. Don't pass it to this function and remove the impossible conditions.

We redefine a variable with the same name below.

framework/sandstone.cpp
thiagomacieira: I agree on not passing it, but see above.

framework/sandstone.cpp
```diff
@@ -1965,120 +1965,35 @@ static TestResult run_one_test_once(const struct test *test)
...
-    bool all_cpus_failed_equally = true;
+    // First, determine if all threads failed the exact same way
+    bool all_threads_failed_equally = true;
     uint64_t fail_pattern = 0;
     int nfailures = 0;
-    for (size_t i = 0; i < num_cpus() && all_cpus_failed_equally; ++i) {
+    for (size_t i = 0; i < num_cpus() && all_threads_failed_equally; ++i) {
         if (per_thread_failures[i]) {
             if (++nfailures == 1)
                 fail_pattern = per_thread_failures[i];
```
thiagomacieira, line 1982: Oops, bug here. Commit 3db9c3914b80e797fc879bfbe96447dc9bb44207 forgot to change `fail_pattern` type to `__uint128` too. Change it to `PerThreadFailures::value_type`.

framework/device/cpu/topology.cpp
thiagomacieira: Change to `PerThreadFailures::value_type`. `uint64_t` is wrong here.

framework/sandstone_p.h
thiagomacieira: It can, but it'll require an update to the internal code to do so. I was hoping to leave it unchanged. You'll see the issue if you build the internal tool with this version of OpenDCDiag.

framework/device/cpu/topology.cpp
```diff
@@ -1562,6 +1562,95 @@ void restrict_topology(DeviceRange range)
         topo = build_topology();
 }
 
+void analyze_test_failures_for_topology(const struct test *test, const PerThreadFailures &per_thread_failures)
+{
+    Topology topology  = Topology::topology();
+    if (!topology.isValid()) {
+        // can't use this information
+        return;
```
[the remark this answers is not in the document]
thiagomacieira: I don't think it'll be removed because the compiler can't see that the topology is invalid. It will be valid in most cases, actually. And frequently it won't be.

But this is a good thing to do.

thiagomacieira (approved):

PR #738, opened 2025-07-24
CPU-separation: disable FrequencyManager for non-cpu builds

mcisowsk:
- parent: https://github.com/opendcdiag/opendcdiag/pull/735


framework/sandstone_opts.cpp
```diff
@@ -84,8 +84,10 @@ enum {
     triage_option,
     ud_on_failure_option,
     use_builtin_test_list_option,
+#ifdef SNADSTONE_DEVICE_CPU
```
thiagomacieira, line 87: Spelling. Repeats below.

thiagomacieira (commented):
Does it make sense to make this a framework option in framework/meson.build?

thiagomacieira (changes requested):
I think I'd prefer that it, actually: make it an option that can be selectively disabled.

thiagomacieira:
> Do you mean setting `SANDSTONE_FREQUENCY_MANAGER` variable, which in our case would be activated for case `get_option('device_type') == 'cpu'`? This way we won't close ourselves to other devices supporting any kind of "frequency management" in future. Do I get this right?

Kinda. I meant two things:

1. it should be something that goes into sandstone_config.h, which will resolve the problem of the spelling because you'll use `#if SANDSTONE_FREQUENCY_MANAGER` instead of `#ifdef`;
2. you should enable/disable it based on `get_option('framework_options')`, applying a default of enabled for CPU device types. That will also allow us to disable it directly if we are in `restricted-cmdline` mode, because you can't activate the frequency manager.

PR #739, opened 2025-07-24
Expose test_requires_smt flag to tests

dilipshivaraju:
Why is this required?

Many tests in Sandstone need hyperthreading(smt) to be effective.
These tests need to write extra code to detect if there is a sibling
thread in the same core(HT). If no sibling thread present, they skip.
This flag will ease test writer's job where they can specify this flag
in test details and framework will skip this test if there are no hyperthreads.
If there is "atleast" 1 hyperthread, the test runs.

bats output:

```
SANDSTONE_BIN=build/opendcdiag bats -t -f "selftest test smt" bats/sanity-check
1..2
ok 1 selftest test smt skip
ok 2 selftest test smt run
```

This is still a draft PR since more tests which consider cases like slicing and Dual Core Module(DCM) needs to be added.

### DIFF
--- a/bats/sanity-check/25-cpu-topology.bats
+++ b/bats/sanity-check/25-cpu-topology.bats
@@ -525,3 +525,32 @@ selftest_cpuset_negated() {
     test_yaml_regexp "/tests/0/threads/3/messages/0/text" "I> ${cpuset[3]}\$"
     test_yaml_regexp "/tests/0/threads/3/messages/1/text" "I> ${cpuset[3]}\$"
 }
+
+@test "selftest test smt skip" {
+    declare -A yamldump
+    # if all thread ids are 0(--cpuset=t0) skip the test
+    sandstone_selftest -e selftest_requires_smt --cpuset=t0
+
+    [[ "$status" -eq 0 ]]
+    test_yaml_regexp "/tests/0/result" 'skip'
+    test_yaml_regexp "/tests/0/skip-category" 'CpuTopologyIssue'
+    test_yaml_regexp "/tests/0/skip-reason" 'Test requires SMT \(hyperthreading\)'
+}
+
+@test "selftest test smt run" {
+    # core0 has 2 threads(0 and 1) where as core1 and core2 have only thread0
+    export SANDSTONE_MOCK_TOPOLOGY="0:0:0 0:0:1 0:1:0 0:2:0"
+    echo "SANDSTONE_MOCK_TOPOLOGY=\"$SANDSTONE_MOCK_TOPOLOGY\""
+
+    declare -A yamldump
+    # run the test
+    sandstone_yq -e selftest_requires_smt --selftests --disable=mce_check
+
+    local threads1=`query_jq '."cpu-info"[] | select(.thread != 0) .logical'`
+
+    if [[ -z "$threads1" ]]; then
+        skip "Test only works with Debug builds to mock the topology or CPUs with hyperthreading"
+    else
+        test_yaml_regexp "/tests/0/result" 'pass'
+    fi
+}
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -1584,7 +1584,6 @@ static void wait_for_children(ChildrenList &children, const struct test *test)
     }
 }
 
-
 static TestResult child_run(/*nonconst*/ struct test *test, int child_number)
 {
     if (sApp->current_fork_mode() != SandstoneApplication::no_fork) {
@@ -1610,7 +1609,7 @@ static TestResult child_run(/*nonconst*/ struct test *test, int child_number)
 
         auto has_smt = []() -> bool {
             for(int idx = 0; idx < num_cpus() - 1; idx++) {
-                if(cpu_info[idx].thread_id + 1 == cpu_info[idx + 1].thread_id)
+                if (cpu_info[idx].core_id == cpu_info[idx + 1].core_id)
                     return true;
             }
             return false;
--- a/framework/sandstone.h
+++ b/framework/sandstone.h
@@ -294,8 +294,9 @@ typedef enum test_flag {
     /// package and not to threads or cores.
     test_failure_package_only       = 0x1000,
 
+    /// Indicates test requires to have Simultaneous Multi-Threading(SMT)/Hyperthreading(HT)
+    /// support
     test_requires_smt               = 0x4000,
-
 } test_flags;
 
 struct test_data_per_thread
--- a/framework/selftest.cpp
+++ b/framework/selftest.cpp
@@ -1343,7 +1343,15 @@ static struct test selftests_array[] = {
     .desired_duration = 750,
     .quality_level = TEST_QUALITY_PROD,
 },
-
+{
+    .id = "selftest_requires_smt",
+    .description = "Test the flag test_requires_smt",
+    .groups = DECLARE_TEST_GROUPS(&group_positive),
+    .test_run = selftest_pass_run,
+    .desired_duration = -1,
+    .quality_level = TEST_QUALITY_PROD,
+    .flags = test_requires_smt,
+},
 {
     .id = "selftest_timedpass_busywait",
     .description = "Runs for the requested time, but busy-waiting", // or practically so


bats/sanity-check/25-cpu-topology.bats
```diff
@@ -525,3 +525,37 @@ selftest_cpuset_negated() {
...
+
+@test "selftest test smt skip" {
+
+    declare -A yamldump
+    # if all thread ids are 0(--cpuset=t0) skip the test
+    sandstone_selftest -e selftest_test_requires_smt --cpuset=t0
+
+    if (( status != 0 )); then
+        skip "Test only works with Debug builds (to mock the topology) or multi-socket systems"
+    fi
```
thiagomacieira, line 537: Remove these lines, this should not be a problem and you didn't set the mock topology anyway. The only problem would be on a hypothetical system with only non-zero thread IDs, which I don't think exist.

bats/sanity-check/25-cpu-topology.bats
```diff
@@ -525,3 +525,37 @@ selftest_cpuset_negated() {
...
+
+@test "selftest test smt run" {
+
+    # core0 has 2 threads(0 and 1) where as core1 has only thread0
+    export SANDSTONE_MOCK_TOPOLOGY="0:0:0 0:0:1 0:1:0 0:2:0"
+    echo "SANDSTONE_MOCK_TOPOLOGY=\"$SANDSTONE_MOCK_TOPOLOGY\""
+
+    declare -A yamldump
+    # run the test
+    sandstone_selftest -e selftest_test_requires_smt
```
thiagomacieira, line 553: Please confirm after the run that the `cpu-info` array sees a non-zero thread ID, in case this test gets run on a non-HT system in a release build. If it doesn't, `skip`.

bats/sanity-check/25-cpu-topology.bats
```diff
@@ -525,3 +525,37 @@ selftest_cpuset_negated() {
...
+
+    if (( status != 0 )); then
+        skip "Test only works with Debug builds (to mock the topology) or multi-socket systems"
```
thiagomacieira, line 556: Bad copy/paste: not about multi-socket.

bats/sanity-check/25-cpu-topology.bats
```diff
@@ -525,3 +525,37 @@ selftest_cpuset_negated() {
...
+    fi
+
+    test_yaml_absent "/tests/0/skip"
```
thiagomacieira, line 559: Remove: we never have a "skip" entry in the dict.

fenrus75:
I would also prefer to be named "test_prefers_smt" since it's not a
technical "requires" -- the test is correct without it.

framework/sandstone.h
```diff
@@ -293,6 +293,9 @@ typedef enum test_flag {
     /// Indicates that a test can only attribute failure to a particular
     /// package and not to threads or cores.
     test_failure_package_only       = 0x1000,
```
busykai, line 295: on a side note, why `test_failure_package_only` and `test_flag_ignore_do_while` have the same  value?

busykai: I should have worn glasses. 🤣

framework/sandstone.cpp
```diff
@@ -1604,6 +1605,21 @@ static TestResult child_run(/*nonconst*/ struct test *test, int child_number)
         init_per_thread_data();
 
         sApp->test_tests_init(test);
+
+        auto has_smt = []() -> bool {
+            for(int idx = 0; idx < num_cpus() - 1; idx++) {
+                if(cpu_info[idx].thread_id + 1 == cpu_info[idx + 1].thread_id)
```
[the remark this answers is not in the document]
dilipshivaraju: I think this requires both `core_id` and `thread_id`.

```
if((cpu_info[idx].core_id == cpu_info[idx+1].core_id) && (cpu_info[idx].thread_id + 1 == cpu_info[idx + 1].thread_id))
        return true;
```

For `0:0:0 0:0:1 0:1:0 0:2:0` since core_ids are equal for core0, test *doesn't* skip which is what is needed.

For `0:0:0 0:1:1` since core_ids are different, test skips which is what is needed.

However, I think if a Dual Code Module(DCM) has 2 physical cores inside the same module and 2 logical threads inside each of these physical cores, then this messes up the ordering of cpus in `cpu_info` data structure. On Linux, there will no longer be thread0 and thread1 next to each other. This is where this feature might break.

For e.g., before `cpu_info` sorting, this might be like this:

image


After sorting, thread0 and thread1 might not be next to each other.

image

Am i wrong anywhere here?

bats/sanity-check/25-cpu-topology.bats
dilipshivaraju: I don't understand this. Is this what is required when you mean confirm in `cpu-info`?

`test_yaml_numeric "/cpu-info/1/thread" 'value == 1'`

framework/sandstone.h
```diff
@@ -293,6 +293,9 @@ typedef enum test_flag {
...
+
+    test_requires_smt               = 0x4000,
+
```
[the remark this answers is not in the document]
dilipshivaraju: Changed this to as Arjan suggested `test_prefers_smt`. Please let me know if this needs to be changed.

busykai:
> I would also prefer to be named "test_prefers_smt" since it's not a technical "requires" -- the test is correct without it.

I actually made a comment after @dilipshivaraju implemented the change you suggested. This flag does imply "required" since the test is supposed to skip on lack of support or if there are no siblings. The tests that do not require HT should not set this flag at all.

framework/sandstone.h
```diff
@@ -293,6 +293,10 @@ typedef enum test_flag {
     /// Indicates that a test can only attribute failure to a particular
     /// package and not to threads or cores.
     test_failure_package_only       = 0x1000,
+
+    /// Indicates test prefers to have Simultaneous Multi-Threading(SMT)/Hyperthreading(HT)
+    /// support
+    test_prefers_smt               = 0x4000,
```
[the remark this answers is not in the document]
dilipshivaraju: I am fine with either `test_prefers_smt` or `test_requires_smt`.

thiagomacieira (changes requested):

thiagomacieira:
> > I would also prefer to be named "test_prefers_smt" since it's not a technical "requires" -- the test is correct without it.
> 
> I actually made a comment after @dilipshivaraju implemented the change you suggested. This flag does imply "required" since the test is supposed to skip on lack of support or if there are no siblings. The tests that do not require HT should not set this flag at all.

We have some tests that **do** require it and will misbehave without. For example, if one thread is waiting for a wake up (like in a semaphore) from the hyperthread partner; without HT the test would deadlock.

If we have any such tests, they currently have a hand-rolled check for the topology.

framework/sandstone.cpp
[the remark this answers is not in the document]
thiagomacieira: The Linux order is irrelevant. We sort everything in an OS-agnostic way in framework/devices/cpu/topology, such that sibling threads are always next to each other.

bats/sanity-check/25-cpu-topology.bats
```diff
@@ -525,3 +525,31 @@ selftest_cpuset_negated() {
...
+
+    declare -A yamldump
+    # run the test
+    sandstone_selftest -e selftest_requires_smt
+
+    if (( status != 0 )); then
+        skip "Test only works with Debug builds to mock the topology"
+    fi
+
+    test_yaml_numeric "/cpu-info/1/thread" 'value == 1'
```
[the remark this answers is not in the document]
dilipshivaraju: Changed it

```
SANDSTONE_BIN=./release_build/opendcdiag taskset -c 0-3 bats -p --tap -f "selftest test smt" bats/sanity-check
1..2
ok 1 selftest test smt skip
ok 2 selftest test smt run # skip Test only works with Debug builds to mock the topology
dshivara@bakelly-desk2:~/opendcdiag/opendcdiag$ SANDSTONE_BIN=./debug_build/opendcdiag bats -p --tap -f "selftest test smt" bats/sanity-check
1..2
ok 1 selftest test smt skip
ok 2 selftest test smt run
```

thiagomacieira (approved):

PR #751, opened 2025-08-08
Prevent clang "built-in" malloc() optimization

jposwiata:
Clang assumes some pre-optimizations for some compile-time conditions. One of these are extremely large allocations. As the framework overrides default implementation, one of the tests (selftest_malloc_fail) doesn't work as expected. The optimization must be forced out in this particular case.


thiagomacieira (changes requested):
Does it help to compile this file or test with` -fno-builtin-malloc`?

jposwiata:
> Does it help to compile this file or test with` -fno-builtin-malloc`?

File scope pragma `optimize` and function attribute doesn't help; no warnings/etc are generated during compilation.
Having it for whole app forces this test to run as expected.

PR #780, opened 2025-09-12
CI/Bats: add tests to ensure the KVM tests do run without crashing

thiagomacieira:
This tests all three run modes: -fno, -fexec (default on Windows), and -fyes (default elsewhere). This is extra pedantic, because of course we can't test the Linux KVM API on Windows, just in case in the future someone adds an equivalent way to test on other OSes.

### DIFF
--- a/bats/sanity-check/29-kvm.bats
+++ b/bats/sanity-check/29-kvm.bats
@@ -1,0 +1,77 @@
+#!/usr/bin/bats
+# -*- mode: sh -*-
+# Copyright 2022 Intel Corporation.
+# SPDX-License-Identifier: Apache-2.0
+load ../testenv
+load helpers
+
+function setup_file() {
+    run $SANDSTONE --selftests --list-group-members @kvm
+    export kvm_tests=$output
+    if [[ $status -ne 0 ]]; then
+        skip "No KVM tests in this build"
+    fi
+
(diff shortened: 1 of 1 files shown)


bats/sanity-check/29-kvm.bats
```diff
@@ -0,0 +1,82 @@
...
+    local -a positive negative
+    for t in $kvm_tests; do
+        if printf '%s\n' $positive_tests | grep -Fqx "$t"; then
+            positive+=($t)
+        else
+            negative+=($t)
+        fi
+    done
+    export kvm_tests_positive="${positive[@]}"
+    export kvm_tests_negative="${negative[@]}"
```
busykai, line 26:
```bash
kvm_tests_positive=$(grep -Fxf <(echo "${kvm_tests}") <(echo "${positive_tests}"))
kvm_tests_negative=$(grep -Fxvf <(echo "${kvm_tests}) <(echo "${kvm_tests_positive}")
export kvm_tests_positive=$(echo "${kvm_tests_positive}" | xargs)
export kvm_tests_negative=$(echo "${kvm_tests_negative}" | xargs)
```

thiagomacieira: Ha! I didn't know of `grep -f`. Neat.

busykai (approved):

PR #790, opened 2025-10-01
FW/Test: preinit the tests in the main process

thiagomacieira:
This is how it was meant to be, how it was designed back in day (commit 7c2b9cd8cc03828503da1526825dc5944ad1a3b6 in the internal repository prior to open-sourcing), but was lost in refactoring with PR #495 (commit 7c2b9cd8cc03828503da1526825dc5944ad1a3b6).

This new design is stricter in that the preinit function can only affect things that the parent process will see, for example, the timing and looping constraints or replacing the init function (so long as it runs in the parent).

No self-test for group inits because in the current design doesn't allow the selftests to have group init, therefore we can't test the side- effect.

### DIFF
--- a/framework/selftest.cpp
+++ b/framework/selftest.cpp
@@ -110,6 +110,33 @@ static int selftest_timedpass_noloop_run(struct test *test, int cpu)
     return EXIT_SUCCESS;
 }
 
+static int selftest_preinit_preinit(struct test *test)
+{
+    check_is_main_process();
+    test->desired_duration = -1;
+    test->maximum_duration = 1; // 1ms only
+    return EXIT_SUCCESS;
+}
+
+static int selftest_preinit_init(struct test *test)
+{
+    // If this is an -fexec run, we can only check things in sApp->shmem for
+    // side-effects from the preinit function.
+
+    // test->desired_duration influences the loop counter (though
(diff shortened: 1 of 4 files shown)


framework/selftest.cpp
```diff
@@ -110,6 +110,33 @@ static int selftest_timedpass_noloop_run(struct test *test, int cpu)
...
+}
+
+static int selftest_preinit_init(struct test *test)
+{
+    // If this is an -fexec run, we can only check things in sApp->shmem for
+    // side-effects from the preinit function.
+
+    // test->desired_duration influences the loop counter (though
+    // --max-test-loop-count takes precedence...)
+    if (sApp->shmem->current_max_loop_count == INT_MAX)
```
mcisowsk, line 128: why would it be `INT_MAX`? Can't we just check `if (sApp->shmem->current_max_loop_count != -1)` ?

mcisowsk (approved):
a very needed change

framework/selftest.cpp
thiagomacieira: Because the test may be getting run with `--max-test-loop-count=2` or some other value. We can't know whether something _else_ changed the max loop count. But we can know if nothing did.

PR #792, opened 2025-10-02
FW: remove dead option `--mce-check-every`

### LINKED ISSUE #586: --mce-check-every has no effect
The `--mce-check-every` option sets `sApp->mce_check_period` but that variable is not used anywhere. I guess some refactoring of the test selection code dropped it. We need to either remove the command-line option or we should restore the behaviour.

I would prefer the latter. It would be useful in some conditions to fail faster in case an MCE interrupt was delivered than at the end of the iteration.

mcisowsk:
Unless we plan to use it in future.

Fixes #586

### DIFF
--- a/framework/sandstone_opts.cpp
+++ b/framework/sandstone_opts.cpp
@@ -49,7 +49,6 @@ enum {
     max_test_loop_count_option,
     max_messages_option,
     max_logdata_option,
-    mce_check_period_option,
     mem_sample_time_option,
     mem_samples_per_log_option,
     no_mem_sampling_option,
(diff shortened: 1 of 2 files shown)


dawid-sabat:
We have a bug for that: https://github.com/opendcdiag/opendcdiag/issues/586

So we have to make a decision if we want to get it back or remove forever.

thiagomacieira (approved):
> We have a bug for that: #586
> 
> So we have to make a decision if we want to get it back or remove forever.

I'd say let's remove. If we need this again, we can re-develop it.

PR #797, opened 2025-10-03
FW: bring back `mce_counts_start` as `SandstoneApplication`s field

mcisowsk:


dawid-sabat (approved):
Let's bring back previous capabilities. At least for now...

PR #803, opened 2025-10-03
CPU-separation: use universal name `device_features_t`

mcisowsk:
This commit also brings back definition of `device_features` (former `cpu_features`) variable back to the common code.

### DIFF
--- a/framework/sandstone.h
+++ b/framework/sandstone.h
@@ -26,7 +26,6 @@
 
 #include "device/device.h"
 
-#include "cpu_features.h"
 #include "sandstone_config.h"
 #include "sandstone_data.h"
 #include <sandstone_test_groups.h>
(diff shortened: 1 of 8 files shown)


framework/sandstone.h
```diff
@@ -345,7 +344,7 @@ struct test {
     /* filled in by framework, used by framework and tests */
 
     /// minimum CPU required to be run, skipped if too old
-    cpu_features_t minimum_cpu;
+    device_features_t minimum_cpu; // we need to keep that name for legacy reasons, ideally should be `minimum_device_to_run`
```
mcisowsk, line 347: sadly

framework/sandstone.h
```diff
@@ -323,7 +322,7 @@ typedef const kvm_config_t *(*kvmconfigfunc)(void);
 struct test {
     /* metadata */
     /// filled in by the DECLARE_TEST macro
-    cpu_features_t compiler_minimum_cpu;
+    device_features_t minimum_device_to_compile;
```
thiagomacieira, line 325: I think `compiler_minimum_device` still, because it's the information provided by the compiler, not for compiling.

thiagomacieira (changes requested):

framework/sandstone.h
mcisowsk: done

thiagomacieira (approved):

PR #818, opened 2025-10-08
FW: do not call `num_cpus()` in device-agnostic code

mcisowsk:


framework/sandstone.cpp
```diff
@@ -1432,6 +1433,7 @@ static TestResult child_run(/*nonconst*/ struct test *test, int child_number)
             state = TestResult::Skipped;
             break;
         }
+#endif
```
thiagomacieira, line 1436: Should we move this to cpu_device.cpp and call it
```c++
    state = device_preinit_test(test);
    if (state != TestResult::Passed)
        break;
```

thiagomacieira (approved):

PR #836, opened 2025-10-10
unittests: test `ProgramOptionsParser` class

mcisowsk:


thiagomacieira (changes requested):
Look ok, but the running the test produces a bunch of stderr output (see https://github.com/opendcdiag/opendcdiag/actions/runs/18585441289/job/53146887850?pr=836). Ideally, you'd modify the test so we can inspect what got written to stderr, but if you can't do that, make sandstone_opts.cpp not print anything.

It can probably be done by making sandstone_opts.cpp not use `printf` and `fprintf` directly, but instead a wrapper that in the test writes to a temp file or a memstream whose contents you will also test. See unix/child_debug.cpp before 2fe1ca18699d21a6347019a28157f55464fa7e9e for a use of `open_memstream()` and that on windows we had to use `tmpfile()`.

framework/unit-tests/opts_parser_tests.cpp
```diff
@@ -0,0 +1,879 @@
...
+        test_stream = open_memstream(&buffer, &buflen);
+#endif
+    }
+
+    ~StreamBuffer() {
+        fclose(test_stream);
+        free(buffer);
+    }
+
+    void flush() {
```
mcisowsk: rename to `commit()` or something similar?

framework/unit-tests/opts_parser_tests.cpp
```diff
@@ -0,0 +1,914 @@
...
+
+        auto ret = opts.parse(2, argv, &cfg);
+        EXPECT_EQ(ret, EX_USAGE);
+        sb.flush();
+        const std::regex pattern(R"(\/unittests: error: value out of range for option '-Y \/ --yaml': 161 \(minimum is 0, maximum 160\).*)");
+        std::smatch match;
+        auto str = std::string(sb.buffer);
+        std::regex_search(str, match, pattern);
+        ASSERT_TRUE(match.size() > 0);
+        EXPECT_STREQ(match[0].str().data(), "/unittests: error: value out of range for option '-Y / --yaml': 161 (minimum is 0, maximum 160)");
```
mcisowsk: should it look like this ultimately?

framework/meson.build
```diff
@@ -311,4 +312,9 @@ unittests_sources += files(
     'unit-tests/test_knob_tests.cpp',
     'unit-tests/tests_dummy.cpp',
     'unit-tests/topology_tests.cpp',
+    'unit-tests/opts_parser_tests.cpp',
 )
+
+unittests_flags += [
+    '-DSANDSTONE_UNITTESTS',
+]
```
thiagomacieira: cc @jgasparakis

thiagomacieira (approved):

PR #847, opened 2025-10-15
CPU-separation: introduce `special_tests` vector to store `mce_test` & others

mcisowsk:
Non-CPU builds would be able to prepend this vector with its' own device-specific tests expected to work the same way as `mce_check`. That is, placed at the end of the test lists, and not being able to be added manually (serving a purpose of a "monitor"). `mce_test` is still guaranteed to be executed as the very last one.

### DIFF
--- a/framework/sandstone_tests.cpp
+++ b/framework/sandstone_tests.cpp
@@ -27,8 +27,9 @@ void SandstoneTestSet::load_all_tests()
         }
     }
 
-    /* add "special" mce_check as well */
-    all_tests.push_back(&mce_test);
+    /* add "special" tests */
+    special_tests = special_tests_for_device();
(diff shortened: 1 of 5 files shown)


framework/sandstone_tests.cpp
```diff
@@ -27,8 +27,11 @@ void SandstoneTestSet::load_all_tests()
         }
     }
 
-    /* add "special" mce_check as well */
-    all_tests.push_back(&mce_test);
+    /* add "special" tests */
+    for (auto special_test : special_tests_for_device()) {
+        special_tests.push_back(special_test);
+    }
```
thiagomacieira:
```c++
    special_tests = special_tests_for_device();
```
no?

dawid-sabat (approved):

PR #849, opened 2025-10-16
FW: add flag `test_in_parent` to execute whole test in parent

mcisowsk:
This allows tests to be run in parent regardless of the command line mode selected. It's a preparation step for device-specific special tests that may want to spawn a 'monitoring' thread in their preinit (to listen for driver events, for example) and join it in cleanup (test added at the end of the test list).

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -1387,17 +1387,8 @@ static void wait_for_children(ChildrenList &children, const struct test *test)
     }
 }
 
-static TestResult child_run(/*nonconst*/ struct test *test, int child_number)
+static TestResult run_one_test_inner(struct test *test, bool init_in_aux_thread = false)
 {
-    if (sApp->current_fork_mode() != SandstoneApplication::ForkMode::no_fork) {
-        protect_shmem();
-        sApp->select_main_thread(child_number);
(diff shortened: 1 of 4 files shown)


framework/sandstone.cpp
```diff
@@ -1773,6 +1773,44 @@ static void run_one_test_init_in_parent(ChildrenList &children, const struct tes
...
+                state = TestResult::Failed;
+            } else if (ret == EXIT_SKIP || thr->has_skipped()) {
+                log_skip(RuntimeSkipCategory, "SKIP requested in cleanup");
+                state = TestResult::Skipped;
+            } else if (ret < EXIT_SUCCESS) {
+                log_skip(RuntimeSkipCategory, "Unexpected OS error in cleanup: %s", strerror(-ret));
+                state = TestResult::Skipped;
+            }
+        }
+    } while(false);
```
thiagomacieira: This is the main portion of `child_run()`, with the difference that the init function is run in the main process' main thread, instead of an auxiliary thread. Instead of C&P, please merge the two, in a function named something like `inner_run_one_test` or `run_one_test_inner`.

thiagomacieira (approved):

lukasz-tuz (approved):
Re-approving after a rebase.

PR #851, opened 2025-10-16
Selftests: Add tests with required minimum CPU.

dawid-sabat:
The idea here is to add a bunch of tests that passes but require various minimum CPU to run. 
This can be useful in development to check if we detect CPU correctly in OpenDCDiag.

### DIFF
--- a/bats/sanity-check/35-cpu-specific.bats
+++ b/bats/sanity-check/35-cpu-specific.bats
@@ -1,0 +1,67 @@
+#!/usr/bin/bats
+# -*- mode: sh -*-
+# Copyright 2025 Intel Corporation.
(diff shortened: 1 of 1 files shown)


bats/sanity-check/35-cpu-specific.bats
```diff
@@ -0,0 +1,67 @@
...
+
+function setup_file() {
+    setup_cpu_features_tests
+}
+
+function run_min_cpu_test() {
+    local test=$1
+
+    declare -A yamldump
+    sandstone_selftest --beta -e $test
```
thiagomacieira: They're all production; `--beta` isn't required.

thiagomacieira (changes requested):

PR #859, opened 2025-10-20
CPU-separation: create interface function `print_fixed_for_device()`

mcisowsk:
This commit adds also ifdef around `CPUTimeFreqStamp`. Ultimately, we can implement a set of device-specific "monitors" catching a snapshot before/after each test, and updating appropriate stuff in thread data.

### DIFF
--- a/framework/logging.h
+++ b/framework/logging.h
@@ -62,6 +62,7 @@ public:
     // device-specific interfaces
     static std::string thread_id_header_for_device(int device, int verbosity);
     static void print_thread_header_for_device(int fd, PerThreadData::Test *thr);
+    static void print_fixed_for_device();
 
     const struct test *test;
     MonotonicTimePoint earliest_fail = MonotonicTimePoint::max();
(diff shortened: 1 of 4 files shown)


framework/logging.h
```diff
@@ -116,11 +116,14 @@ class YamlLogger : public AbstractLogger
 
 std::string thread_id_header_for_device(int device, int verbosity);
 void print_thread_header_for_device(int fd, PerThreadData::Test *thr);
+void print_fixed_for_device();
```
thiagomacieira: Before this commit, can we move the two functions above into `AbstractLogger` as static members?

mcisowsk: ok, we can

framework/sandstone.cpp
```diff
@@ -892,8 +892,10 @@ static uintptr_t thread_runner(int thread_number)
     // indicate to SIGQUIT handler that we're running
     this_thread->thread_state.store(thread_running, std::memory_order_relaxed);
 
+#if SANDSTONE_DEVICE_CPU
     CPUTimeFreqStamp before;
     before.Snapshot(thread_number);
+#endif
```
thiagomacieira: This can probably be factored out to somewhere in devices/cpu.

thiagomacieira (approved):

PR #869, opened 2025-10-22
FW/Logging: add a mode to log raw YAML, not in a string

thiagomacieira:
It is the responsibility of the test author to ensure that the output is valid YAML.

### DIFF
--- a/bats/sanity-check/20-selftests.bats
+++ b/bats/sanity-check/20-selftests.bats
@@ -621,6 +621,21 @@ function selftest_log_skip_init_common() {
     check_test 1
 }
 
+@test "selftest_log_raw_yaml" {
+    declare -A yamldump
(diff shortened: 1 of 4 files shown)


thiagomacieira:
> what about we use nlohmann-json to format the message?

Overkill, IMHO. I don't expect we're going to need that that often. It can also be done by the caller code, without the framework needing to support anything.

I did think of and may still add on top a simple API like:
```c++
using Value = std::variant<std::nullptr_t, bool, int64_t, double, std::string>;
log_yaml(const char *level, std::span<const Value> array);
log_yaml(const char *level, const std::map<std::string, Value> &map);
```
I don't think we'll need nested content.

busykai (approved):
Looks good, but we should be very careful allowing use of this by the tests.

PR #872, opened 2025-10-23
Selftests: do not define `hw_features` test group for non-cpu builds

mcisowsk:


dawid-sabat (approved):
This is fair. We should not have this outside of CPU builds.

PR #873, opened 2025-10-23
unittests: fix stack smashing by stdio's `open_memstream()`

thiagomacieira:
We passed a stack variable as `buflen`, but `open_memstream()`'s routines remember that address and will update it as the buffer grows or shrinks. That causes stack smashing.

Also, as we always free the buffer with `free()`, so we must allocate it with `malloc()` and friends.


mcisowsk (approved):
Image

PR #885, opened 2025-10-28
FW: headers cleanup

mcisowsk:


framework/device/cpu/topology.cpp
```diff
@@ -6,20 +6,24 @@
...
 
+#include <algorithm>
+#include <cassert>
+#include <cinttypes>
+#include <map>
+#include <mutex>
+#include <random>
+#include <string>
+#include <utility>
+#include <vector>
```
thiagomacieira: I prefer the C++ headers above the C ones.

mcisowsk: makes sense

thiagomacieira (approved):

PR #888, opened 2025-10-28
FW/CPU: rename `cpu_info` to `cpu_info_t`

mcisowsk:
This way we can drop the `struct` keyword when referring to this type, as it's not interferring with `cpu_info` variable anymore.

### DIFF
--- a/framework/device/cpu/cpu_device.h
+++ b/framework/device/cpu/cpu_device.h
@@ -122,7 +122,7 @@
 /// This macro is provided in case tests need more fine grained control.
 #define cpu_has_feature(f)      ((device_compiler_features & (f)) == (f) || (device_features & (f)) == (f))
 
-/// Describes the field @c native_core_type in @ref cpu_info
+/// Describes the field @c native_core_type in @ref cpu_info_t
 // use typed enums
 enum NativeCoreType
 #if defined(__cplusplus) || __STDC_VERSION__ >= 202311L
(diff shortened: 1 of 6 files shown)


framework/device/cpu/logging_cpu.cpp
```diff
@@ -9,7 +9,7 @@
 
 #include <cinttypes>
 
-static constexpr const char *native_device_type(const struct cpu_info *info)
+static constexpr const char *native_device_type(const cpu_info_t *info)
```
thiagomacieira: In other words, replace here and in the rest of the files with `device_info_t`

framework/device/cpu/topology.cpp
```diff
@@ -111,7 +111,7 @@ struct TopologyDetector
 };
 } // unnamed namespace
 
-static void update_topology(std::span<const struct cpu_info> new_cpu_info,
+static void update_topology(std::span<const cpu_info_t> new_cpu_info,
```
thiagomacieira: Or `Topology::Thread` in this file (they are the same type).

framework/device/cpu/cpu_device.h
```diff
@@ -190,17 +190,19 @@ struct cpu_info
 #endif
 };
 
-// Alias for use in common framework code
-typedef struct cpu_info device_info;
+// Aliase for use in common framework code.
+typedef struct cpu_info device_info_t;
+// Temporary alias until we rename this struct to cpu_info_t.
+typedef struct cpu_info cpu_info_t;
```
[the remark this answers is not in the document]
mcisowsk: I also wanted to first leave the `cpu_info` untouched, then redo internal code, then go back here and rename `cpu_info` -> `cpu_info_t` as the last step. This way we'd never loose compatibility.

PR #889, opened 2025-10-28
FW/CPU: rename `cache_info` to `cache_info_t`

mcisowsk:
Keep it consistent with `cpu_info_t`.

---
requires: https://github.com/opendcdiag/opendcdiag/pull/888


thiagomacieira (commented):
Don't see the point. Also, I'd prefer:
```c
typedef struct cache_info 
{
....
} cache_info_t;
```

mcisowsk:
wanted to keep it consistent with `struct cpu_info { /*...*/ };`. There is no `typedef` there.

PR #890, opened 2025-10-29
FW: do not call `restrict_topology()` inside `commit_shmem()`

mcisowsk:
All we need from this call is the update of `cpu_info` ptr, as `sApp->shmem` may have moved. A side effect of this is that `build_topology()` is always called, because `sApp->shmem` changed, even though device range has not changed and no update to the topology is required. The way it's been implemented was enforcing other implementations of `restrict_topology()` to know that they must provide update of the pointer inside of it as well. Since this step is completely device-agnostic, let's just call it directly at the end of `commit_shmem()`.

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -1036,8 +1036,8 @@ static void commit_shmem()
         sApp->shmemfd = -1;
     }
 
-    // sApp->shmem has probably moved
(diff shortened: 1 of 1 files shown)


mcisowsk:
I got this wrong, we rely on `cpu_info` pointer in `cached_topology()`, so anytime it changes, we have to re-`build_topology()`

PR #899, opened 2025-11-13
FW/Linux: widen the "wchan" field of ps in case of freeze/timeout

thiagomacieira:
The current output contains only 6 characters, which causes a lot of ambiguity with common prefixes that could in turn mean a lot of different things: "migrat" could be "migrate_pages" (a memory problem) or "migration_to_node" (NUMA?). 20 characters isn't enough to disambiguate all symbols, but hopefully the problem area should be clear by then.


mcisowsk (approved):
lgtm

PR #902, opened 2025-11-17
Fix and reword command-line help.

busykai:
This removes double spaces, "you" and provides better/shorter
explanation in places. Formatting is done in a separate commit to facilitate the review.

Signed-off-by: Arzhan Kinzhalin <arzhan.i.kinzhalin@intel.com>

### DIFF
--- a/framework/sandstone_opts.cpp
+++ b/framework/sandstone_opts.cpp
@@ -300,9 +300,8 @@ Common command-line options are:
      Randomizes the order in which tests are executed.
  --test-delay <time in ms>
      Delay between individual test executions in milliseconds.
-  -Y, --yaml [<indentation>]
-     Use YAML for logging. The optional argument is the number of spaces to
-     indent each line by (defaults to 0).
+  -Y, --yaml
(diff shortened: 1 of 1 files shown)


mcisowsk (dismissed):
lgtm

thiagomacieira (dismissed):

framework/sandstone_opts.cpp
```diff
@@ -266,47 +264,45 @@ Common command-line options are:
...
  --test-list-randomize
      Randomizes the order in which tests are executed.
  --test-delay <time in ms>
      Delay between individual test executions in milliseconds.
-  -Y, --yaml [<indentation>]
-     Use YAML for logging. The optional argument is the number of spaces to
-     indent each line by (defaults to 0).
-For more options and information, please see the User Reference
-Guide.
+  -Y, --yaml
```
thiagomacieira: Aside from the fact that you're now no longer saying this takes an optional argument, this is ok.

`getopt` supports `-vY` because `-v` takes no options, but `-Yv` won't work.

PR #904, opened 2025-11-20
FW: rename `cpu_info` global array to `device_info`

mcisowsk:
Leave `cpu_info` alias for backward compatibility though.

### DIFF
--- a/framework/device/cpu/cpu_device.h
+++ b/framework/device/cpu/cpu_device.h
@@ -193,16 +193,18 @@ struct cpu_info_t
 // Alias for use in common framework code.
 typedef struct cpu_info_t device_info_t;
 
-/// cpu_info is an array of cpu_info_t structures.  Each element of the array
+/// device_info is an array of cpu_info_t structures.  Each element of the array
 /// contains information about a logical CPU that will be used to
 /// execute a test's test_run function.  The size of this array is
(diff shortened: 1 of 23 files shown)


framework/device/cpu/cpu_device.h
```diff
@@ -195,16 +195,16 @@ typedef struct cpu_info device_info_t;
 // Temporary alias until we rename this struct to cpu_info_t.
 typedef struct cpu_info cpu_info_t;
 
-/// cpu_info is an array of cpu_info_t structures.  Each element of the array
+/// device_info is an array of cpu_info_t structures.  Each element of the array
 /// contains information about a logical CPU that will be used to
 /// execute a test's test_run function.  The size of this array is
 /// equal to the value returned by num_cpus().
-extern struct cpu_info *cpu_info;
+extern struct cpu_info *device_info;
```
thiagomacieira: This one is a harder sell because we have quite a few tests depending on this and they'll all require update.
```
$ git grep '\bcpu_info\[' tests | wc -l
194
```

How about giving two names, with a macro for opting in or out?
```c
extern struct cpu_info *device_info;
#ifndef NO_COMPAT_CPU_INFO
extern struct cpu_info *cpu_info asm("device_info");
#endif
```
https://c.godbolt.org/z/Yd46WvnGz

PR #908, opened 2025-11-25
FW/Logging: fixes around `format_duration`

mcisowsk:
Noticed when smi_count skips. 
Overall, before:
```
      threads:
      - thread: main
        runtime: -96987658.00-
        resource-usage: { utime: 0.000, stime: 0.000, cpuavg: -0.0, maxrss: 0, majflt: 0, minflt: 0, voluntary-cs: 0, involutary-cs: 0 }
        messages:
```
after:
```
      threads:
      - thread: main
        runtime: 0.0
        resource-usage: { utime: 0.000, stime: 0.000, cpuavg: 0.0, maxrss: 0, majflt: 0, minflt: 0, voluntary-cs: 0, involutary-cs: 0 }
        messages:
```

### DIFF
--- a/framework/logging.cpp
+++ b/framework/logging.cpp
@@ -1834,7 +1834,7 @@ void YamlLogger::maybe_print_slice_resource_usage(int fd, int slice)
             indent_spaces().data(),
             format_duration(utime, FormatDurationOptions::WithoutUnit).c_str(),
             format_duration(stime, FormatDurationOptions::WithoutUnit).c_str(),
-            (utime + stime) * 100.0 / runtime,
+            std::abs((utime + stime) * 100.0 / runtime),
             (long long)usage.ru_maxrss, (long long)usage.ru_majflt
 #ifndef _WIN32
             , (long long)usage.ru_minflt, (long long)usage.ru_nvcsw, (long long)usage.ru_nivcsw


framework/logging.cpp
```diff
@@ -1834,7 +1834,7 @@ void YamlLogger::maybe_print_slice_resource_usage(int fd, int slice)
             indent_spaces().data(),
             format_duration(utime, FormatDurationOptions::WithoutUnit).c_str(),
             format_duration(stime, FormatDurationOptions::WithoutUnit).c_str(),
-            (utime + stime) * 100.0 / runtime,
+            std::abs((utime + stime) * 100.0 / runtime),
```
thiagomacieira: No, this doesn't look right. Why are the numbers negative in the first place?

`utime` and `stime` come from the `usage`, so they shouldn't be negative. So my guess it's `runtime`:
```c++
    auto runtime = slices[slice].endtime - sApp->current_test_starttime;
```
That's probably because we didn't set `endtime`. Something I thought I'd fixed... 606566be57deac84673f0cb307e65ba9a9cc7ca9, ah, interrupted tests only. I'd rather fix similarly there.

framework/sandstone_chrono.cpp
```diff
@@ -74,22 +74,28 @@ string format_duration(std::chrono::nanoseconds ns, FormatDurationOptions opts)
...
-    result.reserve(i + 7);
-    result.resize(i + 4);
-    result[i++] = '.';
-    if (us.count() < 100)
-        result[i++] = '0';
-    if (us.count() < 10)
-        result[i++] = '0';
-    std::to_chars(result.data() + i, result.data() + result.size(), us.count(), 10);
+    if (ns.count() < 0) {
+        result += "0.0";
```
thiagomacieira: No, this is wrong. True, the calculation below might be wrong for negative values, but this is not right either.

thiagomacieira (changes requested):

PR #918, opened 2025-12-18
build: use GitHub Actions for building targets

mcisowsk:
Follow the standard and define targets (cpu, cpu-win, gpu) as separate GitHub actions.
This enables changes in those files to be executed in the PR's CI checks.


lukasz-tuz:
Looks good to me.

We could go a step further and parametrize the build action so that a workflow provides all the necessary parameters (like `target_device`). This way, we could have just a single build composite action (well, two single build actions, one per OS). The build action would install the required additional packages as needed when building for the GPU; the rest of the build process is the same for both types of devices.

PR #923, opened 2026-01-07
CI: stucture the worflow so that it is easier to navigate

busykai:
1. Running of the tests moved from the "build" action back to the workflow;
2. Dependency installation (which has copious output on Linux) is foldable now.


.github/actions/build-cpu/action.yaml
```diff
@@ -37,6 +33,7 @@ runs:
       shell: bash
       run: |
         DEBIAN_FRONTEND=noninteractive apt-get -y update
+        echo "::group::Installing build dependencies"
```
thiagomacieira: Thanks!

PR #929, opened 2026-01-13
FW/GPU: implement `detect_`/`setup_devices()`

mcisowsk:


framework/device/gpu/topology.cpp
```diff
@@ -41,17 +44,114 @@ void print_temperature_of_device()
...
+            return EXIT_FAILURE;
+        }
+        enabled_devices.at(indices).zes_handle = device_handle; // matching indices must mean the exact same device
+        return EXIT_SUCCESS;
+    });
+
+    if (ret != EXIT_SUCCESS) [[unlikely]] {
+        fprintf(stderr, "%s: internal error: gpu enumeration failed!\n",
+                program_invocation_name);
+        exit(EX_OSERR);
```
mcisowsk: this causes exiting with error for systems without gpus, even for commands so benign like `opendcdiag --version`. Is this ok?

framework/device/gpu/topology.cpp
```diff
@@ -187,17 +190,111 @@ void print_temperature_of_device()
...
+    cpu_info = sApp->shmem->device_info;
+
+    assert(enabled_devices.size() == thread_count());
+    gpu_info_t* info = cpu_info;
+    const gpu_info_t* cend = cpu_info + thread_count();
+
+    auto enabled_cpus = find_enabled_logical_cpus();
+    auto enabled_cpu_index = 0;
+
+    std::ranges::for_each(enabled_devices, [&](auto& enabled_device) {
```
thiagomacieira: Why not a plain `for`:
```c++
    for (auto &enabled_device : enabled_devices) {
```

thiagomacieira (changes requested):

framework/device/gpu/topology.cpp
mcisowsk: no reason, and you're right - it does not fit here

framework/device/gpu/topology.cpp
```diff
@@ -187,17 +190,185 @@ void print_temperature_of_device()
...
+    std::ifstream infile;
+    infile.open(file.data());
+    if (!infile.is_open()) { [[unlikely]]
+        fprintf(stderr, "%s: internal error: unable to find physical_package_id file\n",
+                program_invocation_name);
+        return res;     // TODO: exit or return -1 and continue?
+    }
+    std::string str{5}; // TODO: simply max num digits of an int16_t
+    infile.read(str.data(), str.size());
+    infile.close();
```
lukasz-tuz: I think 4 digits would be a safe value. I suppose that 2 digits would suffice in 99.99% of cases, but let's add a few more.

framework/device/gpu/topology.cpp
```diff
@@ -187,17 +190,196 @@ void print_temperature_of_device()
...
+
+/// Tries to find an intersection of enabled_cpus and NUMA local CPUs. If no such exists,
+/// assigns first cpu from enabled_cpus and emmits a warning. Removes the assigned CPU from enabled_cpus
+/// to avoid duplicates.
+int try_assign_local_cpu(std::vector<int>& enabled_cpus, const ze_pci_address_ext_t& bdf)
+{
+    int res = -1;
+    auto local_cpus = find_numa_local_cpus(bdf);
+    std::vector<int> available_cpus;
+    std::ranges::set_intersection(enabled_cpus, local_cpus, std::back_inserter(available_cpus));
```
thiagomacieira: There's also a `numa_node`:
https://codebrowser.dev/linux/linux/drivers/pci/pci-sysfs.c.html#numa_node_show

Here, you wouldn't do a `set_intersection`, but `std::copy_if` if the `numa_id` matches.

framework/device/gpu/topology.cpp
thiagomacieira: Just use `>>`, no?
```c++
    infile >> res;
```

PR #933, opened 2026-01-14
FW: Set default rescheduling mode to Queue

AthenasJimenez:
Also fixes a minor issue where the "none" value was previously treated as invalid; it is now accepted as a valid option.

Previous:
```
$ ./sandstone --reschedule=none
./sandstone: unknown reschedule option: none
```

Now:
```
./sandstone --reschedule=none --disable=*
Ran 1 tests without error (0 skipped)
exit: pass
```

[ChangeLog][rescheduling] Set default rescheduling mode to Queue

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -2661,9 +2661,8 @@ int main(int argc, char **argv)
             return EX_USAGE;
         }
         if (sApp->device_scheduler) {
-            fprintf(stderr, "%s: error: --reschedule is not supported in this configuration\n",
-                    program_invocation_name);
-            return EX_USAGE;
+            sApp->device_scheduler = nullptr;
+            logging_printf(LOG_LEVEL_VERBOSE(1), "# WARNING: --reschedule is not supported in this configuration\n");
         }
     }
 
--- a/framework/sandstone_p.h
+++ b/framework/sandstone_p.h
@@ -245,7 +245,7 @@ struct TestConfig
     bool selftest = false;
     bool ud_on_failure = false;
     bool use_strict_runtime = false;
-    RescheduleMode reschedule_mode = RescheduleMode::none;
+    RescheduleMode reschedule_mode = RescheduleMode::queue;
 
     // logging parameters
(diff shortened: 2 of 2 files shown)


framework/sandstone_opts.cpp
```diff
@@ -874,9 +874,11 @@ struct ProgramOptionsParser {
             }
         }
 
-        if (auto value = string_opt_for(reschedule_option)) {
+        {
```
mcisowsk: I don't think we need a separate scope

framework/sandstone_opts.cpp
```diff
@@ -874,9 +874,11 @@ struct ProgramOptionsParser {
...
+            const char *value = string_opt_for(reschedule_option);
+            value = value ? value : "queue"; // If not specified, set default
             app_cfg->device_scheduler = make_rescheduler(value);
-            if (!app_cfg->device_scheduler) {
+            if (!app_cfg->device_scheduler && std::string_view{value} != "none") {
```
mcisowsk: maybe put checking of `std::string_view{value} != "none"` first

framework/sandstone_opts.cpp
AthenasJimenez: We don't, I was just trying to avoid a variable called `value_reschedule` only for this case. But I'm open to change it.

framework/device/cpu/topology.cpp
```diff
@@ -128,15 +128,21 @@ int num_packages()
...
+        case RescheduleMode::queue:
+            res = std::make_unique<QueueDeviceScheduler>();
+            break;
+        case RescheduleMode::barrier:
+            res = std::make_unique<BarrierDeviceScheduler>();
+            break;
+        case RescheduleMode::random:
+            res = std::make_unique<RandomDeviceScheduler>();
+            break;
+        default:
```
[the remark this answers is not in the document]
AthenasJimenez: At the end, I leaved it as if/else statement. It works better in my opinion.

Issue #784, opened 2025-09-19
Logging: report the virtualisation/containerisation state of the run


thiagomacieira:
There's a CPUID leaf for that and we report the virtualisation state... only in the TAP and key-value logging modes (forgot to add it to the YAML one).

Additionally, it may be useful to report too the containerisation state. For that, use [systemd-detect-virt](https://www.freedesktop.org/software/systemd/man/latest/systemd-detect-virt.html).

dawid-sabat:
As mentioned above, this is resolved. Closing.

PR #937, opened 2026-01-16
FW/Logging: report virtualisation/contenerisation state in YAML logs

gibasm:
Added container detection via 'systemd-detect-virt' (SysV only), and logging of the virtualisation/contenerisation state in the YAML format.

The proposed changes address a freature request described in #784.

### DIFF
--- a/framework/logging.cpp
+++ b/framework/logging.cpp
@@ -11,6 +11,8 @@
 #include "sandstone_p.h"
 #include "sandstone_iovec.h"
 #include "sandstone_utils.h"
+#include "sandstone_virt.h"
+
 #if SANDSTONE_SSL_BUILD
 #  include "sandstone_ssl.h"
 #endif
@@ -2134,6 +2136,21 @@ void YamlLogger::print()
     logging_flush();
 }
 
+void YamlLogger::maybe_print_virt_state() {
+    auto detected_vm = detect_running_vm();
+    auto detected_container = detect_running_container();
+
+    if ((!detected_vm.empty()) || (!detected_container.empty())) {
+        logging_printf(LOG_LEVEL_VERBOSE(1), "virtualization-state: { %s%s%s%s }\n",
+                detected_vm.empty() ? "" : "vm: ",
+                detected_vm.empty() ? "" : detected_vm.c_str(),
+                detected_container.empty() ? "" : detected_vm.empty()
+                    ? "container: " : ", container: ",
+                detected_container.empty() ? "" : detected_container.c_str()
+            );
+    }
+}
+
 void YamlLogger::print_header(std::string_view cmdline, Duration test_duration, Duration test_timeout)
 {
     using ::format_duration;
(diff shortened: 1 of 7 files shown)


framework/sandstone_containers.cpp
```diff
@@ -0,0 +1,67 @@
...
+std::optional<std::string> detect_running_container() {
+    auto path = find_detect_virt_path();
+
+    if (path.has_value()) {
+        std::string detected = detect_container_via_systemd(path.value());
+
+        if (detected.compare("none") == 0) {
+            return std::nullopt;
+        } else {
+            return std::optional<std::string>(detected);
```
mcisowsk: nit: `return detected;` should suffice

lukasz-tuz (approved):

framework/sandstone_containers.cpp
```diff
@@ -0,0 +1,71 @@
...
+ */
+#include "sandstone_containers.h"
+
+#if __linux__
+
+#include <unistd.h>
+
+static std::string detect_container_via_systemd(const std::string& path) {
+    std::string cmd = path + " -c";
+    FILE* fp = popen(cmd.c_str(), "r");
```
busykai: I really dislike the `popen` approach because of artisan command construction. could you use an `exec` variant instead?

framework/sandstone_containers.cpp
```diff
@@ -0,0 +1,71 @@
...
+        }
+    }
+
+    /* either failed to detect or not running inside of a container */
+    return std::nullopt;
+}
+
+#else
+
+std::optional<std::string> detect_running_container() {
```
thiagomacieira: Why `optional` instead of empty string?

framework/sysdeps/unix/virt.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+    if (fp == NULL) {
+        return "";
+    }
+
+    char container_type[max_line_len];
+
+    if(fgets(container_type, max_line_len, fp) == NULL) {
+        return "";
+    }
```
thiagomacieira: Hint: You can write:
```c++
    std::string type_str;
    str.resize_and_overwrite(max_line_len, [&](char *buf, size_t len) {
        return fgets(buf, len, fp) ? strlen(buf) : 0;
    });
```

framework/logging.cpp
```diff
@@ -2139,6 +2136,20 @@ void YamlLogger::print()
...
+        logging_printf(LOG_LEVEL_VERBOSE(1), "vm: %s\n",
+                detected_vm.c_str());
+    }
+
+    auto detected_container = detect_running_container();
+    if (!detected_container.empty()) {
+        logging_printf(LOG_LEVEL_VERBOSE(1), "container: %s\n",
+                detected_container.c_str());
+    }
+}
```
thiagomacieira: I'm going to approve, but can you think of a way to print this in a single line? That is, a good  name for the entry, instead of `system`:
```yaml
system: { vm: kvm, container: systemd-nspawn }
```

thiagomacieira (approved):

PR #939, opened 2026-01-16
Change reschedule() to an in-line API

AthenasJimenez:
This commit changes reschedule() to an inline function that checks reschedule is enabled before calling reschedule_internal(), minimizing performance overhead when not enabled.

### DIFF
--- a/framework/sandstone.h
+++ b/framework/sandstone.h
@@ -487,7 +487,13 @@ static inline bool write_msr(int cpu, uint32_t msr, uint64_t value)
 #endif
 
 /// Interface for C-based tests
-void reschedule();
+extern bool reschedule_enabled;
+void reschedule_internal(void);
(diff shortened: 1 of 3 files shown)


framework/sandstone.h
```diff
@@ -487,7 +487,13 @@ static inline bool write_msr(int cpu, uint32_t msr, uint64_t value)
 #endif
 
 /// Interface for C-based tests
-void reschedule();
+extern bool reschedule_enabled;
+extern void reschedule_internal(void);
```
mcisowsk: why do we declare it here like that (explicitly marked `extern`)? I think function declarations are implicitly `extern`

AthenasJimenez: No a real reason, something I put at some point while I was developing. It's redundant but not an actual problem. I'm ok removing it.

PR #945, opened 2026-01-19
Limited RNG usage functions

jposwiata:
Two families of functions to "consume" parts of RNG generated values: both given number of bits and values in given range.
WIP, unit tests in progress
The number of interface functions will be reduced too (to work with single set of state variables).


framework/random_utils.cpp
```diff
@@ -0,0 +1,222 @@
...
+
+    std::lock_guard<std::mutex> lock(mutex);
+    // with out-of-available-range values rejection
+    if (random_val_available < range) {
+        random_val_value = G();
+        random_val_available = get_mask<T>(B);
+    }
+    T val = random_val_value % range;
+    random_val_available /= range;
+    random_val_value /= range;
```
thiagomacieira: See this discussion for uniform generation: https://lore.kernel.org/lkml/e50884573b0f88ebd1f5b495476c13f28a5f0aa2@opteya.com/T/. 

For a commit on top, to simplify reviewing.

framework/random_utils.cpp
```diff
@@ -0,0 +1,222 @@
...
+} // anonymous namespace
+
+extern "C" {
+
+uint64_t get_random_bits31(uint32_t num_bits) {
+    return get_random_bits<uint64_t, random, BITS_FROM_RANDOM>(num_bits);
+}
+uint64_t get_random_bits128(uint32_t num_bits) {
+    return get_random_bits<uint64_t, random128, 128>(num_bits);
+}
```
thiagomacieira: What's the point of using `random128` but only returning 64 bits?

framework/random_utils.cpp
```diff
@@ -0,0 +1,222 @@
...
+        bits -= b;
+    }
+    return val;
+}
+
+template<typename T, auto G, uint32_t B>
+T get_random_value(T range) {
+    static std::mutex mutex{};
+    static T random_val_value = 0;
+    static T random_val_available = 0;
```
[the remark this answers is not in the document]
thiagomacieira: Sorry, the main issue is sharing the sample among threads. That is not acceptable. 

You *may* keep the last sample in a `thread_local` variable, but I think that is overkill.

Issue #634, opened 2025-05-23
Repeated --cpuset options are not accumulating


thiagomacieira:
This used to work in our tool's March 31 release:
```
$ /tmp/prevrelease/opendcdiag --cpuset=t1 --cpuset=c0 --dump-cpu-info 
Detected CPU: Skylake (Avx512); family-model-stepping (hex): 06-55-04; CPU features: clflush,sse2,pclmul,fma,sse4.2,movbe,popcnt,aes,avx,f16c,rdrnd,fsgsbase,bmi,avx2,bmi2,rtm,avx512f,avx512dq,rdseed,adx,clflushopt,clwb,avx512cd,avx512bw,avx512vl,lzcnt
# CPU   PkgID   CoreID  ThrdID  Microcode       PPIN
14      0       0       1       0x2007006
```
But now overriding the previous options:
```
$ ./opendcdiag --cpuset=t1 --cpuset=c0 --dump-cpu-info
Detected CPU: Skylake (Avx512); family-model-stepping (hex): 06-55-04; CPU features: clflush,sse2,pclmul,fma,sse4.2,movbe,popcnt,aes,avx,f16c,rdrnd,fsgsbase,bmi,avx2,bmi2,rtm,avx512f,avx512dq,rdseed,adx,clflushopt,clwb,avx512cd,avx512bw,avx512vl,lzcnt
# CPU   PkgID   CoreID  ThrdID  Microcode       PPIN
0       0       0       0       0x2007006
14      0       0       1       0x2007006
```

Probably caused by 324758092282f08f02aea80a7f951493b71eb6d1

PR #947, opened 2026-01-20
FW: fix repeated --deviceset/cpuset not accumulating

gibasm:
With multiple --cpuset/deviceset in the cli, the last were overriding the previous ones. 
The intended behavior is for the repeated --cpuset/device to accumulate as described in #634.

### DIFF
--- a/bats/sanity-check/25-cpu-topology.bats
+++ b/bats/sanity-check/25-cpu-topology.bats
@@ -490,6 +490,57 @@ selftest_cpuset_negated() {
     selftest_cpuset_negated \!p${cpuinfo[1]}c${cpuinfo[2]}t${cpuinfo[3]} ${cpuinfo[0]}
 }
 
+# Test if multiple --cpuset accumulate properly
+# e.g. --cpuset=p0 --cpuset=c0 should be the same as --cpuset=p0c0
+function selftest_cpuset_accumulate() {
+    local separate=$1
+    local accumulated=$2
+    local expression=$3
+
+    local cpuset=`$SANDSTONE $separate --dump-cpu-info \
+                 | awk '/^[0-9]/ { printf "%d,", $1; }'`
+    cpuset=${cpuset%,}          # remove last comma
+    local -a cpuinfo=(`$SANDSTONE --dump-cpu-info --cpuset=$accumulated \
+                     | awk '/^[0-9]/ && '"$expression"' { print $1; }'`)
+
+    selftest_cpuset_unsorted "$cpuset" "${cpuinfo[@]}"
+}
+
+@test "cpuset accumulate (p0 c0)" {
+    export SANDSTONE_MOCK_TOPOLOGY='p0c0 p0c1'
+
+    run $SANDSTONE --cpuset=p0c0,p1c1 --dump-cpu-info
+    if (( status != 0 )); then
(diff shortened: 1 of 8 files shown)


framework/sandstone_opts.cpp
```diff
@@ -696,7 +696,12 @@ struct ProgramOptionsParser {
         }
 
         // deviceset (before dump_cpu_info)
-        opts.deviceset = string_opt_for<std::string>(deviceset_option);
+        if (opts_map.contains(deviceset_option)) {
+            auto deviceset_vec = std::get<std::vector<const char*>>(opts_map.at(deviceset_option));
+            for (auto d : deviceset_vec) {
+                opts.deviceset += std::string(d);
```
busykai: It should be a comma-separated list in the `deviceset` since it's parsed later. This way it would result in `--cpuset=1` and `--cpuset=2` in `12` in the `opts.deviceset`.

gibasm: You're completely right, missed the comma, should be fixed now.

thiagomacieira (changes requested):

framework/sandstone_opts.cpp
```diff
@@ -696,7 +696,13 @@ struct ProgramOptionsParser {
         }
 
         // deviceset (before dump_cpu_info)
-        opts.deviceset = string_opt_for<std::string>(deviceset_option);
+        if (opts_map.contains(deviceset_option)) {
+            auto deviceset_vec = std::get<std::vector<const char*>>(opts_map.at(deviceset_option));
+            for (auto d : deviceset_vec) {
+                opts.deviceset += std::string(d) + ',';
+            }
```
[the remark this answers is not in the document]
busykai: Isn't this the issue of parsing? How it used to work before the accumulation was broken?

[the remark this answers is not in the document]
thiagomacieira: The way it was supposed to be was that separate `--cpuset` options were "AND" and options within one were "OR". So for example:
* `--cpuset=p0 --cpuset=c0,c1,c2` = cores 0, 1 and 2 (all threads) of package 0. Same as `p0c0,p0c1,p0c2`
* `--cpuset=p0 --cpuset=!c0` = all cores of package 0, except core 0. Not equivalent to `!p0c0` in multi-socket systems.

I think the way to go is to simply gather all of the options in a `std::vector` and call `apply_deviceset_param()` over it.

bats/sanity-check/25-cpu-topology.bats
```diff
@@ -490,6 +490,63 @@ selftest_cpuset_negated() {
...
 
+# Test if multiple --cpuset accumulate properly
+# e.g. --cpuset=p0 --cpuset=c0 should be the same as --cpuset=p0c0
+function selftest_cpuset_accumulate() {
+    local separate=$1
+    local accumulated=$2
+    local expression=$3
+
+    local cpuset=(`$SANDSTONE $separate --dump-cpu-info \
+                 | awk '/^[0-9]/ { printf "%d,", $1; }'`)
```
thiagomacieira: Note: the `( )` here are incorrect. You've created an array of one entry. No effect in the execution, though.

thiagomacieira (approved):
Approving so we unblock, but please firm up the bats expression.

PR #952, opened 2026-01-21
FW/Logging: pass the `PerThreadData::Common` pointer around more

thiagomacieira:
Instead of asking for it over and over again. The compiler will eliminate duplicate calls when they're nearby, but not all functions will get inlined.

### DIFF
--- a/framework/logging.cpp
+++ b/framework/logging.cpp
@@ -318,12 +318,12 @@ const char *AbstractLogger::char_to_skip_category(int val)
 }
 
 template <typename... Args> static ssize_t
-log_message_for_thread(int thread_num, LogTypes logType,
+log_message_for_thread(PerThreadData::Common *thread, LogTypes logType,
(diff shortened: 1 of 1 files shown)


framework/logging.cpp
```diff
@@ -1225,7 +1229,8 @@ static void logging_format_data(DataType type, std::string_view description, con
                   "remark:      'memcmp_offset() could not locate difference'";
     }
 
-    log_message_for_thread(thread_num, LogTypes::RawYaml, LOG_LEVEL_QUIET, buffer);
+    PerThreadData::Common *thread = sApp->thread_data(thread_num);
+    log_message_for_thread(thread, LogTypes::RawYaml, LOG_LEVEL_QUIET, buffer);
```
mcisowsk: nit: is `log_message_for_thread(sApp->thread_data(thread_num), LogTypes::RawYaml, LOG_LEVEL_QUIET, buffer);` too long?

PR #953, opened 2026-01-22
FW/GPU: add RAII wrappers around C-based L0 structures

mcisowsk:
Wrappers around structs that we plan to frequently use during test development, like data buffers, command queues, etc.

### DIFF
--- a/framework/device/gpu/ze_wrappers.h
+++ b/framework/device/gpu/ze_wrappers.h
@@ -1,0 +1,164 @@
+/*
+ * Copyright 2026 Intel Corporation.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef INC_ZE_WRAPPERS_H
+#define INC_ZE_WRAPPERS_H
+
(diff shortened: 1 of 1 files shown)


lukasz-tuz (dismissed):
Seems to be good to go now.

framework/device/gpu/ze_wrappers.h
```diff
@@ -0,0 +1,167 @@
...
+    {
+        ZE_CHECK_AND_LOG(zeMemAllocDevice(context, &desc, size, alignment, device, &data));
+    }
+
+    ~ZeDeviceDataPtr() {
+        ZE_CHECK_AND_LOG(zeMemFree(context, data));
+    }
+
+    // zeKernelSetArgumentValue requires ref
+    void*& get() noexcept {
```
thiagomacieira: Are you sure about returning a reference? That would allow the caller to modify the pointer.

lukasz-tuz (approved):

PR #961, opened 2026-01-26
[random_bits] Minimal interface to get optimized random value and set of bits

jposwiata:
Two families of functions to optimize RNG usage by the code. Functions to fetch given number of bits and generate random value in given range (ti [0..range-1]) provided, both as C functions and C++ templates.
The task is only to provide the interface with minimal implementations (updated in later commits).
Basic unit tests provided (with required definitions of random-generating functions defined)


thiagomacieira (dismissed):

framework/random_utils.cpp
```diff
@@ -0,0 +1,214 @@
...
+template<typename T, auto G, uint32_t B>
+T get_random_bits(uint32_t bits) {
+    decltype(G()) random_bits_value = 0;
+    uint32_t random_bits_available = 0;
+
+    static_assert(B <= 8 * sizeof(random_bits_value),
+        "Generator cannot provide more bits than the generated type can hold");
+
+    T val = 0;
+    while (bits != 0) {
```
thiagomacieira: Please help the compiler by breaking up the first loop, which is the common scenario:
```suggestion
    auto random_bits_value = G();
    static_assert(B <= 8 * sizeof(random_bits_value),
        "Generator cannot provide more bits than the generated type can hold");

    int random_bits_available = B - bits;
    T val = T(random_bits_value);
       
    while (random_bits_available < 0) {
        // get B more bits
        random_bits_value = G();
        random_bits_available += B;
        
        // 
    }
```

framework/sandstone.h
```diff
@@ -575,12 +573,26 @@ int thread_count() __attribute__((pure));
...
 
-extern void log_thread_context(std::string_view msg);
+template<typename T>
+inline T get_random_bits(uint32_t num_bits) {
+    static_assert((std::is_unsigned<T>::value), "get_random_bits() only supports unsigned types");
+    static_assert(sizeof(T) <= sizeof(uint64_t), "get_random_bits() only supports types up to 64 bits");
+    assert((num_bits <= sizeof(T) * 8) && "Requested number of bits must fit the type");
+
+    // TODO optimize for AES (use get_random_bits128 when AES is selected)
+    return static_cast<T>(get_random_bits31(num_bits));
```
thiagomacieira:
```suggestion
    if (sizeof(T) * 8 <= 31 || (__builtin_constant_p(num_bits) && num_bits <= 31)
        return static_cast<T>(get_random_bits31(num_bits));
    return static_cast<T>(get_random_bits128(num_bits));
```

framework/random_utils.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+        // as random data is random.. no need to keep endianness in mind, we can just
+        // store bits in any particular order
+        val ^= T(random_bits_value & get_mask<T>(b));
+        random_bits_available -= b;
+        if (random_bits_available != 0) {
+            random_bits_value >>= b;
+        }
+        bits -= b;
+    }
```
thiagomacieira: Much simpler, though makes the patch to achieve the caching later more complex.
```suggestion
T get_random_bits(int bits) {
    auto random_bits_value = G();
    static_assert(B <= 8 * sizeof(random_bits_value),
        "Generator cannot provide more bits than the generated type can hold");

    T val = T(random_bits_value);
    num_bits -= B;
       
    while (num_bits >= 0) {
        // get B more bits
        random_bits_value = G();
        
        // move them into val (as random data is random.. no need to keep endianness in 
        // mind, we can just store bits in any particular order)
        val = (val << B) ^ random_bits_value;
        num_bits -= B;
    }
    return val;
}
```

framework/random_utils.cpp
```diff
@@ -0,0 +1,77 @@
+/*
+ * Copyright 2022-2026 Intel Corporation.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "sandstone.h"
+#include <bit>
+
```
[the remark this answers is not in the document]
jposwiata: indeed, std::countl_zero() moved to another PR.
Won't fix it, as the file is intended to be removed.

jposwiata:

PR #964, opened 2026-01-26
FW/Logging: add `log_thread_context()`

thiagomacieira:
This function can be used by tests to add a line to the thread header that will be printed even in case of freezing, indicating extra context information.


thiagomacieira:
> I'm not seeing any additional changes. Is there a separate PR for the change in the docs?

No, I made a mistake in pushing.

lukasz-tuz (dismissed):
Looks good.

PR #972, opened 2026-01-28
FW/GPU: introduce GPU test class

mcisowsk:
- requires: #691


mcisowsk:
may revisit in future

PR #986, opened 2026-02-09
[random_bits] Value rejection pattern implemented for get_random_val()

jposwiata:
"Optimized" implementation moved to new PR on top of base #961


jposwiata:

PR #987, opened 2026-02-09
[random_bits] Optimized random bits fetching

jposwiata:
Optimization to reduce RNG usage in the context of multiple calls to fetch particular number of random bits.
`get_random_bits()` integrated with existing random generation framework, some optimizations applied.


thiagomacieira (dismissed):

framework/unit-tests/random_utils_mock.cpp
```diff
@@ -0,0 +1,137 @@
...
+    return &engine;
+}
+static thread_rng* rng_for_thread(int) {
+    static thread_rng rng{};
+    return &rng;
+}
+
+uint64_t RandomEngineWithMock::get_random_bits_8(thread_rng* rng, uint32_t bits) {
+    // let discard 24 bits from each mocked/fetched value, generate32/mock<uint32_t> is used.
+    // The cache is 32 bits, so we can fetch up to 4 random values (8bit)
```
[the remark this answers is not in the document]
jposwiata: No. bits 8..31 are never used, as number of available bits is increased by 8 only. The result is not masked, it is a kind of "optimization" only.

framework/random.cpp
```diff
@@ -110,13 +106,23 @@ void RandomEngineDeleter::operator()(RandomEngineWrapper *ptr) const
...
 namespace {
+
+#ifndef SANDSTONE_UNITTESTS
+
+static RandomEngineWrapper* random_engine(RandomEngineWrapper* engine = nullptr) {
+    if (engine) {
+        sApp->random_engine.reset(engine);
+    }
+    return sApp->random_engine.get();
+}
```
thiagomacieira: API design: please split the setter from the getter, and give a `set_` prefix to the setter.

framework/random.cpp
```diff
@@ -206,8 +240,135 @@ template <typename E> struct EngineWrapper : public RandomEngineWrapper
...
+            "RNG context must be big enough to hold engine state, cache and bits availability");
+
+        // it is not allowed to use this function in the context of different cache sizes (as the
+        // number of available bits might be bigger than the actual per-type cache holds).
+        // AES uses first 256 bits of the context for the seed/IV, therefore we must fit cache
+        // state in the remaining.
+        static_assert(sizeof(C) + sizeof(uint32_t) <= sizeof(*rng) / 2,
+            "Cache and bits availability must fit second half of the RNG context");
+        C& cache = *reinterpret_cast<C*>(&rng->u8[sizeof(*rng) / 2]);
+        uint32_t& available = *reinterpret_cast<uint32_t*>(&rng->u8[sizeof(*rng) - sizeof(uint32_t)]);
```
thiagomacieira: Design: please move the `cache` and `available` fields into the engine itself, so we avoid this nasty casting and the need for `static_cast` here. The `static_assert(sizeof(E)` elsewhere will take care ensuring it fits into the `thread_rng` block.

That will also reduce the number of template parameters to this function, because you can deduce from the members of `E`.

thiagomacieira (changes requested):
Ok, this is looking good now. 

The idea looks right, even if I'm not entirely sold on the idea of a cache. We can always disable it via some compile-time switch.

The implementation needs some clean-up. Those `reinterpret_cast` are mighty ugly and make reading the code a little difficult.

Nitpicks for later.

framework/random.cpp
```diff
@@ -50,7 +51,6 @@ DECLSPEC_IMPORT BOOLEAN WINAPI SystemFunction036(PVOID RandomBuffer, ULONG Rando
 #  define O_CLOEXEC 0
 #endif
 
-namespace {
 union alignas(64) thread_rng {
```
thiagomacieira: Why did you move them out of the unnamed namespace?

thiagomacieira (changes requested):

thiagomacieira (changes requested):

jposwiata:

PR #992, opened 2026-02-10
FW/Opts: treat non-arguments as test names

### LINKED ISSUE #979: opendcdiag runs with incorrect cmdline arguments if `-` or `--` are not present.
## How to reproduce:

Esure `-` or `--` are missing in cmdline.
Run opendcdiag with cmdline. Let's say the cmd is legitimate but you made an error typing it down. For example:

```
./opendcdiag max-test-loop-count=2 -o -
```

instead of

```
./opendcdiag --max-test-loop-count=2 -o -
```

I can run it with random string:

```
./opendcdiag abcdef
```

I'll get a correct error if I type down `-arg` or `--arg`. 

Which indicates OpenDCDiag looks for `-` or `--`. 

## What is expected:

OpenDCDiag returns an error due to incorrect cmdline argument and doesn't run the tests.

## Impact:

Users don't know their cmdline is incorrect and they're convinced they do everything right.
Unexpected results: for example I needed two loops and got something else.

mcisowsk:
So far they were always ignored which was confusing. Now they are attempted to be added to the enabled tests list, but only when they come after `--`. Otherwise a warning is emitted.
```
$ build/opendcdiag -e zlib -- zlib2
build/opendcdiag: Cannot find matching tests for 'zlib2'
$ build/opendcdiag -e zlib zlib2
build/opendcdiag: option 'zlib2' is ignored - positional arguments should be put after --
command-line: 'opendcdiag -e zlib zlib2'
version: opendcdiag-5ebccb2ff82b
exit: pass
```

### DIFF
--- a/framework/sandstone_opts.cpp
+++ b/framework/sandstone_opts.cpp
@@ -645,6 +645,20 @@ struct ProgramOptionsParser {
                 return EX_USAGE;
             }
         }
+
+        // attempt to treat positional arguments as test names
+        if (optind < argc) {
+            bool should_parse_positional_args = strcmp(argv[optind - 1], "--") == 0;
+            for (auto i = optind; i < argc; i++) {
+                if (should_parse_positional_args) {
+                    add_to_map_as_vec('e', argv[i]);
+                } else {
+                    fprintf(ERR_STREAM, "%s: option '%s' is ignored - positional arguments should be put after --\n",
+                        program_invocation_name, argv[i]);
+                }
+            }
+        }
+
         return EXIT_SUCCESS;
     }
 
--- a/framework/unit-tests/opts_parser_tests.cpp
+++ b/framework/unit-tests/opts_parser_tests.cpp
@@ -901,3 +901,61 @@ TEST(ProgramOptionsParser, repeated_opt_accumulates__cpuset_deviceset)
         sb.check_eq(EMPTY_STR); // no messages printed
     }
 }
+
+TEST(ProgramOptionsParser, positional_args__parsed_as_test_names_after_double_dash)
+{
+    {
+        StreamBuffer sb;
+        ProgramOptions opts;
+        SandstoneApplicationConfig cfg{};
+        char* argv[] = {
+            (char*)"foo-bar", // binary name
+            (char*)"-etest1",
+            (char*)"--enable=test2",
+            (char*)"--",
+            (char*)"test3",
+            (char*)"test4",
+            (char*)"test5",
+        };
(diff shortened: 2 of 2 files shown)


busykai:
This is bad idea, @mcisowsk. To me a free positional argument should be a warning first and then, after some time, an error. We had PR with similar functionality: #364. I do not see why this is needed and this is going to be more confusing, imo, than it is right now.

We can introduce a new syntax for such stuff, e.g. `+test_name1` is equivalent to `-e test_name1`, and `-test_name2` (already conflicting with option parsing) is equivalent to `--disable test_name`, but not just take any stuff as a test name.

thiagomacieira (changes requested):
> This is bad idea, @mcisowsk. To me a free positional argument should be a warning first and then, after some time, an error. We had PR with similar functionality: #364. I do not see why this is needed and this is going to be more confusing, imo, than it is right now.

I don't see why it should be an error. No one is using them now, and using the positional arguments as test names just makes sense. If nothing else, for scripts like I said, because expanding a shell array with `-e` prepended is non-obvious in bash:
```
$ set foo bar baz
$ echo "-e$@"'
-efoo bar baz
```
You need to do an in-array replacement instead:
```
$ echo "${@/#/-e}"
-efoo -ebar -ebaz
```
That fortunately does keep whitespaces intact:
```
$ set "foo   bar" baz
$ printf %s\\n "${@/#/-e}"'
-efoo   bar
-ebaz
```

Zsh has an interpolator (RC_EXPAND_PARAM mode) and retains whitespace by default, which is why I prefer it:
```
$ set "foo  bar" baz
$ printf %s\\n -e${^@}
-efoo  bar
-ebaz
```

But it is still much easier to not have to do any of that.
```bash
local testlist=(foo bar baz)
testlist+=(quux)
...
./opendcdiag --on-crash=core --retest-on-fail=0 -- "${testlist[@]}"
```

> We can introduce a new syntax for such stuff, e.g. `+test_name1` is equivalent to `-e test_name1`, and `-test_name2` (already conflicting with option parsing) is equivalent to `--disable test_name`, but not just take any stuff as a test name.

We can also use `!` as a prefix to disable a test if using `-` is difficult. Both characters have other implied meanings (`!` is shell history control).

In fact, what I'd like to do is allow the command-line parameters to work like the test list file, with timing controls too. And likewise, allow the test list file to disable tests (do test lists allow wildcards?). One other thing I'd like is to enable the fatal-skip mode for a single test or plain fatal.

PR #1033, opened 2026-03-17
Allow oversubscription/undersubscription to CPUs

AthenasJimenez:
These are experimental changes to play around with reschedule.

### DIFF
--- a/framework/device/cpu/topology.cpp
+++ b/framework/device/cpu/topology.cpp
@@ -170,8 +170,8 @@ void BarrierDeviceScheduler::reschedule_to_next_device()
     std::unique_lock lock(groups_mutex);
     // Initialize groups on first run
     if (groups.empty()) {
-        int full_groups = num_cpus() / members_per_group;
-        int partial_group_members = num_cpus() % members_per_group;
+        int full_groups = device_count() / members_per_group;
+        int partial_group_members = device_count() % members_per_group;
 
         groups.reserve(full_groups + (partial_group_members > 0));
         for (int i=0; i<full_groups; i++) {
@@ -243,7 +243,7 @@ void QueueDeviceScheduler::shuffle_queue()
     // Must be called with mutex locked
     if (queue.size() == 0) {
         // First use: populate queue with the indexes available
-        for (int i=0; i<num_cpus(); i++)
+        for (int i=0; i<device_count(); i++)
             queue.push_back(i);
     }
 
@@ -254,7 +254,7 @@ void QueueDeviceScheduler::shuffle_queue()
 void RandomDeviceScheduler::reschedule_to_next_device()
 {
     // Select a random cpu index among the ones available
-    int next_idx = unsigned(random()) % num_cpus();
+    int next_idx = unsigned(random()) % device_count();
     pin_to_next_cpu(device_info[next_idx].cpu_number);
 
     return;
@@ -378,7 +378,7 @@ static bool cpu_compare(const cpu_info_t &cpu1, const cpu_info_t &cpu2)
 __attribute__((noinline))
 void TopologyDetector::sort()
 {
-    std::sort(device_info, device_info + num_cpus(), cpu_compare);
+    std::sort(device_info, device_info + device_count(), cpu_compare);
 }
 
 bool TopologyDetector::old_create_mock_topology(const char *topo)
@@ -398,7 +398,7 @@ bool TopologyDetector::old_create_mock_topology(const char *topo)
 
     int cpu_count = 0;
     while (topo && *topo) {
-        if (cpu_count == sApp->thread_count)
+        if (cpu_count == sApp->device_count)
             break;      // can't add more
 
         cpu_info_t *info = &device_info[cpu_count];
@@ -428,7 +428,7 @@ bool TopologyDetector::old_create_mock_topology(const char *topo)
             continue;
     }
 
-    sApp->thread_count = cpu_count;
+    sApp->device_count = cpu_count;
     return true;
 }
 
@@ -455,7 +455,7 @@ bool TopologyDetector::create_mock_topology(const char *topo)
 
     int cpu_count = 0;
     while (topo && *topo) {
-        if (cpu_count == sApp->thread_count)
+        if (cpu_count == sApp->device_count)
             break;      // can't add more
 
         cpu_info_t *info = &device_info[cpu_count];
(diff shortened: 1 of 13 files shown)


framework/sandstone_opts.cpp
```diff
@@ -286,6 +288,11 @@ Common command-line options are:
  -s <STATE>, --rng-state=<STATE>
      Specify the random generator state to reload. The seed is in the form:
      Engine:engine-specific-data
+ --subscription=<PERCENT>
```
mcisowsk, line 291: would anything other than 100%, 200%, 300% etc. be usefull for us? Maybe sth like `oversubscription_multiplier` as integer would be better? With default 1. I also see that right now 200% is max. Why is that?

mcisowsk: what's the benefit in undersubscribing?

AthenasJimenez: > what's the benefit in undersubscribing?

The benefit comes from using `undersubscription` combined with `reschedule`. In this scenario, threads may jump across all available devices. As a result, some devices may be woken up from idle to run a heavy‑weight task, only to become idle again shortly after. For the rest of the tests, without reschedule, the behavior is comparable to -n/--threads.

> would anything other than 100%, 200%, 300% etc. be usefull for us? 

Yes, values like 20%, 30% and 50% may be interesting for us as well.

> Maybe sth like oversubscription_multiplier as integer would be better? With default 1.

When I discussed with @thiagomacieira  and @busykai , the idea was to use percentages. Personally, I think it makes more sense to use `--subscription=50` rather than `--subscription=0.5` or `--subscription=.5`.

> I also see that right now 200% is max. Why is that?

I still don't know how much threads per logical core can be supported, so I used the ratio 2:1. It may change in the near future.

thiagomacieira: We can always support both forms: if the '%' isn't there, it's a fraction. We can also accept a negative number as in core or thread count to leave unused.

framework/device/cpu/topology.cpp
```diff
@@ -1815,24 +1815,26 @@ void update_topology(std::span<const cpu_info_t> new_cpu_info,
...
 
-    int new_thread_count = end - device_info;
-    if (int excess = sApp->thread_count - new_thread_count; excess > 0)
+    int new_device_count = end - device_info;
+    if (int excess = sApp->device_count - new_device_count; excess > 0)
         std::fill_n(end, excess, (cpu_info_t){});
 
-    sApp->thread_count = new_thread_count;
+    sApp->device_count = new_device_count;
+    sApp->thread_count = sApp->device_count;
```
thiagomacieira, line 1823: Wondering if we can leave `thread_count = -1` here, and only fill it in later, so we don't have to doubly-track it. This function is only called from `apply_deviceset_param()` and that only from `main()`.

framework/device/cpu/topology.cpp
```diff
@@ -1815,24 +1815,26 @@ void update_topology(std::span<const cpu_info_t> new_cpu_info,
...
     LogicalProcessorSet result = ambient_logical_processor_set();
-    sApp->thread_count = result.count();
-    if (sApp->thread_count == 0) [[unlikely]] {
+    sApp->device_count = result.count();
+    if (sApp->device_count == 0) [[unlikely]] {
         fprintf(stderr, "%s: internal error: ambient logical processor set appears to be empty!\n",
                 program_invocation_name);
         return result;
     }
+    sApp->thread_count = sApp->device_count;
```
thiagomacieira, line 1837: Ditto as above.

framework/device/cpu/topology.cpp
```diff
@@ -1852,12 +1854,12 @@ void setup_devices<LogicalProcessorSet>(const LogicalProcessorSet &enabled_devic
...
     auto old_cpu_info = std::exchange(device_info, sApp->shmem->device_info + range.starting_device);
-    int old_thread_count = std::exchange(sApp->thread_count, range.device_count);
+    int old_device_count = std::exchange(sApp->device_count, range.device_count);
 
     Topology &topo = cached_topology();
-    if (old_cpu_info != device_info || old_thread_count != sApp->thread_count ||
+    if (old_cpu_info != device_info || old_device_count != sApp->device_count ||
             topo.packages.size() == 0)
         topo = build_topology();
 }
```
thiagomacieira, line 1865: Here we've reduced `device_count` to `range.device_count`. But we haven't updated `thread_count`; you did it in the caller.

framework/sandstone.cpp
```diff
@@ -803,6 +805,11 @@ int device_count()
     return sApp->device_count;
 }
 
+void update_thread_count()
+{
+    sApp->thread_count = std::max(1, sApp->device_count * sApp->subscription_ratio / 100);
```
thiagomacieira, line 810: Overflow check: we're safe. We're pretty far from 21 million devices (`INT_MAX / 100`). Even at a 1000% subscription ratio, this would require 2.1 million devices before we overflow.

framework/sandstone.cpp
```diff
@@ -995,8 +1002,21 @@ int main(int argc, char **argv)
     if (sApp->total_retest_count < -1 || sApp->retest_count == 0)
         sApp->total_retest_count = 10 * sApp->retest_count; // by default, 100
 
+
+    if (opts.subscription_ratio != -1 ) {
+        sApp->subscription_ratio = opts.subscription_ratio;
+
+        // -n/--threads and --subscription cannot be used together. Print warning and ignore -n/--threads
```
thiagomacieira, line 1009: Why not? Why can't I oversubscribe 2 threads? Allowing this would make debugging far simpler, when we can oversubscribe 1 and 2 threads and see what's going on.

If you can `taskset` or `--cpuset` the limit, you should be able to `-n` too.

thiagomacieira (changes requested):

framework/device/cpu/topology.cpp
```diff
@@ -1815,24 +1815,26 @@ void update_topology(std::span<const cpu_info_t> new_cpu_info,
...
 
-    int new_thread_count = end - device_info;
-    if (int excess = sApp->thread_count - new_thread_count; excess > 0)
+    int new_device_count = end - device_info;
+    if (int excess = sApp->device_count - new_device_count; excess > 0)
         std::fill_n(end, excess, (cpu_info_t){});
 
-    sApp->thread_count = new_thread_count;
+    sApp->device_count = new_device_count;
+    update_thread_count();
```
mcisowsk, line 1823: how about we make it an atomic operation?
```c++
void update_counts(int count)
{
   sApp->device_count = count;
   sApp->thread_count = std::max(1, sApp->device_count * sApp->subscription_ratio / 100);
}
```
so that we won't forget to set any of them counts

AthenasJimenez:
Closing this PR in favor of #1082

PR #1042, opened 2026-04-06
FW/Reschedule: Omit warning message when running on Windows

AthenasJimenez:


framework/sandstone.cpp
```diff
@@ -1010,7 +1010,9 @@ int main(int argc, char **argv)
         if (device_scheduler) {
             delete device_scheduler;
             device_scheduler = nullptr;
+#ifndef _WIN32
             logging_printf(LOG_LEVEL_VERBOSE(1), "# WARNING: --reschedule is not supported in this configuration\n");
+#endif
```
thiagomacieira, line 1015: Ideally we'd detect when the fork mode was not specified on the command-line, so we're still in the default one.

But I don't think we should make this more complex. It's already an `#if` elsewhere.
```c++
    ForkMode fork_mode =
#ifdef _WIN32
        ForkMode::exec_each_test;
#else
        ForkMode::fork_each_test;
#endif
```

thiagomacieira (approved):

PR #1056, opened 2026-04-16
FW/Selftests: use thread-safe `random32()` in run functions

mcisowsk:


framework/selftest.cpp
```diff
@@ -440,7 +440,7 @@ static int selftest_check_sequential_init(struct test *test)
 
 static int selftest_check_sequential_run(struct test *test, int thread)
 {
-    usleep(1'000 * (random() % 16u));   // sleep up to 16 ms
+    usleep(1'000 * (random32() % 16u));   // sleep up to 16 ms
```
mcisowsk, line 443: I know this test runs threads sequentially, but let's avoid calling `random()` in `test_run` by principle.

thiagomacieira: `random()` and `random32()` are two front-ends to the same backend:
```c++
[[gnu::noinline]] uint32_t random32()
{
    return sApp->random_engine->generate32(thread_local_rng());
}
...
// POSIX-defined to return in the interval [0, 2^31).
long int random()
{
    return sApp->random_engine->generateInt(thread_local_rng());
}
```

thiagomacieira (commented):
This is unnecessary.

mcisowsk:
unnecessary change

PR #1064, opened 2026-04-21
FW/Win32: fix mmap() with non-zero offset

thiagomacieira:
CreateFileMappingW's dwMaximumSizeHigh/Low parameters describe the total size of the file mapping object.  Passing only `length` means that any view mapped at a non-zero offset that reaches beyond `length` bytes from the start of the file returns ERROR_ACCESS_DENIED from MapViewOfFileEx.

Fix by computing the required mapping object size as off + length. Amends commit ea0080dd71a40d91c59ca8d043639fee4a2a0ae0 in the old repository.


fenrus75 (approved):

busykai (approved):
using `size_t` for map size is actually good since it solves the problem of what's being supported.

PR #1066, opened 2026-04-22
FW/child_debug: print the 32 bytes around $pc when showing a crash

thiagomacieira:
Because this isn't useful:
```yaml
    messages:
    - { level: error, text: 'E> Received signal 4 (Illegal instruction) code=2 (ILL_ILLOPN), RIP = 0x7ffff656f080, trap=6 (UD), error_code = 0x0' }
    - level: warning
      text: |1
       W> #3  0x00007ffff656f080 in ?? ()
       => 0x7ffff656f080:       (bad)
```

That doesn't tell me if it was a garbage/bad instruction, or if it was a good instruction not supported in the current machine that the user's gdb was unable to disassemble.

We use the Linux system call process_vm_readv(2) to read up to 32 bytes centered on $pc (16 bytes before, 16 bytes after) and print them to the log below the register state.

When the 32-byte window crosses a page boundary, each page is read independently so that an error reading page on one side does not suppress bytes from the other side. Bytes that could not be read are shown as "??". For example:
```yaml
    - level: info
      text: |1
       Code around RIP (0x7f0aea6a3ffc):
         0x7f0aea6a3fec: ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ?? ??
         0x7f0aea6a3ffc: ?? ?? ?? ?? 00 00 00 00 00 00 00 00 00 00 00 00
```

This commit adds three new selftests to check those boundary conditions:
- `selftest_sigill_partial`: crash on valid RIP, but near unreadable bytes
- `selftest_sigsegv_readable_unreadable`: crash on invalid RIP with the unreadable page coming after the readable one.
- `selftest_sigsegv_unreadable_readable`: crash on invalid RIP with the unreadable page coming before the readable one.

Plus a modification to `selftest_sigsegv_instruction` so the crash is always near the page boundary. Assuming 0x1000 is also unreadable, this should cause both page reads to fail and no bytes to get logged.

[ChangeLog][Framework] Added a feature to print the bytes of the instructions whenever a crash is detected.

Exercise left for the reader: for architectures where instructions are fixed-size and aligned, like AArch64, consider aligning the value from $pc and printing as words instead of bytes.

### DIFF
--- a/bats/sanity-check/20-selftests.bats
+++ b/bats/sanity-check/20-selftests.bats
@@ -1892,6 +1892,53 @@ selftest_crash_common() {
     fi
 }
 
+crash_code_dump_common() {
+    local testname=$1 signame=$2
+    if $is_windows; then
+        skip "Unix-only test"
+    fi
+    check_gdb_usable
+    sandstone_selftest -n1 -e $testname --on-crash=context
+    [[ "$status" -eq 1 ]]
+    test_yaml_regexp "/exit" fail
+    test_yaml_regexp "/tests/0/result" crash
+    local signum=`kill -l $signame`
+    test_yaml_regexp "/tests/0/threads/1/messages/0/text" \
+        ".*Received signal $signum \(.*\) code=[0-9]+.* RIP = 0x.*"
+    test_yaml_regexp "/tests/0/threads/1/messages/2/text" \
+        "Code around RIP \(0x[0-9a-f]+\):"
+}
+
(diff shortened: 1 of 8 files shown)


thiagomacieira:
Needs #1061

busykai (dismissed):

busykai (approved):
This looks good to me!

PR #1075, opened 2026-04-30
cpu_device: guard has_smt() against invalid topology

### LINKED ISSUE #1074: has_smt() returns true when CPU topology is invalid
## Bug

The `has_smt()` lambda inside `prepare_test_for_device()` (`framework/device/cpu/cpu_device.cpp`) determines SMT availability by comparing adjacent `core_id` values:

```cpp
auto has_smt = []() -> bool {
    for(int idx = 0; idx < thread_count() - 1; idx++) {
        if (device_info[idx].core_id == device_info[idx + 1].core_id)
            return true;
    }
    return false;
};
```

When topology detection fails, all `core_id` fields are initialised to `-1` (unknown). The comparison `-1 == -1` evaluates to `true`, so `has_smt()` incorrectly reports SMT as available. Tests gated on `test_requires_smt` will proceed to run with invalid topology data.

## Expected behaviour

`has_smt()` should return `false` (or skip) when topology is invalid. The fix could use `Topology::topology().isValid()` (already available in the same translation unit) as an early-out:

```cpp
auto has_smt = []() -> bool {
        return false;
    for (int idx = 0; idx < thread_count() - 1; idx++) {
        if (device_info[idx].core_id == device_info[idx + 1].core_id)
            return true;
    }
    return false;
};
```

busykai:
When topology detection fails all `core_id` and `package_id` fields are initialised to `-1`. The adjacent-entry comparison in `has_smt()` would evaluate `-1 == -1` as true and incorrectly report SMT as available, causing tests gated on `test_requires_smt` to run with invalid topology data.

Add an early-out using `Topology::topology().isValid()` so that `has_smt()` returns `false` whenever topology data is not available.

Fixes #1074

### DIFF
--- a/framework/device/cpu/cpu_device.cpp
+++ b/framework/device/cpu/cpu_device.cpp
@@ -57,14 +57,6 @@ void dump_device_info()
 
 TestResult prepare_test_for_device(struct test *test)
 {
-    auto has_core_id_defined = []() -> bool {
-        for(int idx = 0; idx < thread_count(); idx++) {
-            if (device_info[idx].core_id == -1)
-                return false;
-        }
-        return true;
-    };
-
     auto has_smt = []() -> bool {
         for(int idx = 0; idx < thread_count() - 1; idx++) {
             if (device_info[idx].package_id == device_info[idx + 1].package_id &&
@@ -75,7 +67,7 @@ TestResult prepare_test_for_device(struct test *test)
     };
 
     if (test->flags & test_requires_smt) {
-        if (!has_core_id_defined()) {
+        if (!Topology::topology().isValid()) {
             log_skip(CpuTopologyIssueSkipCategory, "Test requires topology information");
             return TestResult::Skipped;
         }


framework/device/cpu/cpu_device.cpp
```diff
@@ -66,6 +66,8 @@ TestResult prepare_test_for_device(struct test *test)
     };
 
     auto has_smt = []() -> bool {
+        if (!Topology::topology().isValid())
+            return false;
```
thiagomacieira: I suggest moving this insertion to the previous lambda, `has_core_id_defined` and maybe rename it to `has_topology_information`. That causes the skip message to be "Test requires topology information" instead of "Test requires SMT".

framework/device/cpu/cpu_device.cpp
thiagomacieira: Or maybe replace that lambda with `Topology::topology().isValid()`, because the conditions under which the vector will be empty are:
1. https://github.com/opendcdiag/opendcdiag/blob/aecffd8dd6d2f9492e8c7ecc6a4b576ca6957503/framework/device/cpu/topology.cpp#L1608-L1611
2. https://github.com/opendcdiag/opendcdiag/blob/aecffd8dd6d2f9492e8c7ecc6a4b576ca6957503/framework/device/cpu/topology.cpp#L1614-L1615
3. https://github.com/opendcdiag/opendcdiag/blob/aecffd8dd6d2f9492e8c7ecc6a4b576ca6957503/framework/device/cpu/topology.cpp#L1623-L1625

In Copilot's explanation:
>  1. `device_info[last].package_id < 0` — the last CPU entry has a negative package_id (checked before the main loop, line 1608–1611).
> 2. Any CPU at the start of a package scan has package_id < 0, core_id < 0, or thread_id < 0 (line 1614–1615) — checked at the top of the while loop for the first thread of each new package.
> 3. Any CPU within a package scan has core_id < 0 or thread_id < 0 (line 1624–1625) — checked in the inner for loop iterating over threads within a package.
>
> All three cases return `Topology({})` immediately — an empty packages vector.
> 
> In short: `isValid()` returns false whenever topology detection finds any CPU entry with an uninitialized/invalid ID field (package_id, core_id, or thread_id set to -1/negative).

That should be a superset of what the lambda is checking.

thiagomacieira (changes requested):
And you need to sign-off your commit

busykai:
Done. I like returning `false` on invalid topology more.

thiagomacieira (approved):

PR #1077, opened 2026-05-05
FW: improve timing by not measuring test prep overhead

mcisowsk:
This commit introduces a new `time_at_test_start` variable, set right before calling `test_init`, and after `prepare_test_for_device`. We'd need it for 2 cases:
1. More true timing of the test in general.
2. For GPUS, `prepare_test_for_device` may additionally implement checking the device state and issuing warm restarts if necessary. This would take time and corrupt the timing results.


thiagomacieira (approved):
Looks reasonable to me.

PR #1078, opened 2026-05-05
FW: deobfuscate variable name in `calculate_wallclock_deadline`

mcisowsk:
This was confusing to name `MonotonicTimePoint::clock::now()` as `later` and then return `later + duration`.

### DIFF
--- a/framework/sandstone_run.cpp
+++ b/framework/sandstone_run.cpp
@@ -538,11 +538,11 @@ inline void test_the_test_data<true>::test_tests_finish(const struct test *the_t
 
 static MonotonicTimePoint calculate_wallclock_deadline(Duration duration, MonotonicTimePoint *pnow = nullptr)
 {
-    MonotonicTimePoint later = MonotonicTimePoint::clock::now();
+    MonotonicTimePoint now = MonotonicTimePoint::clock::now();
     if (pnow)
-        *pnow = later;
+        *pnow = now;
 
-    return later + duration;
+    return now + duration;
 }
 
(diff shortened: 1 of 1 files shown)


thiagomacieira (approved):
FYI, this is a historical legacy of how the code was originally. It was changed to look like this in the old repository's commit 
```
commit 933de75c4b96b709bceb8160b309e241369ba6a4
Author: Thiago Macieira <thiago.macieira@intel.com>
Date:   Wed Aug 19 14:20:57 2020 -0700

    FW: convert the internal time storage to std::chrono
    
    That means we're actually using nanoseconds instead of milliseconds.
```
The old one was added to the original C file in
```
commit b6647019e7af852c375a9bf0f9f47a740e6ef14d
Author: Thiago Macieira <thiago.macieira@intel.com>
Date:   Thu Jul 25 15:25:53 2019 -0700

    FW: Add a feature to run all tests for a specified time or forever
```
and it looked like:
```c
static struct timespec calculate_deadline(int duration, struct timespec *pnow)
{
    struct timespec now, later;
    if (!pnow)
        pnow = &now;

    clock_gettime(CLOCK_MONOTONIC, pnow);
    later = *pnow;

    if (duration > 0) {
        uint64_t nsecs = duration * 1000 * 1000;
        later.tv_nsec += nsecs % (1000 * 1000 * 1000);
        later.tv_sec += nsecs / (1000 * 1000 * 1000);
    }

    return later;
}
```

That is, it calculated *both* "now" and "later" (= "now + duration"). Because premature optimisation and I wanted to avoid calling `clock_gettime()` more than once.

PR #1082, opened 2026-05-06
Introduce device_count to decouple devices and threads tracking

AthenasJimenez:
Historically, the number of threads and the number of devices have been strongly tied — reducing thread_count also restricted the number of devices/CPUs to match and viceversa. This PR is the first step toward changing that behavior.

These commits introduce `device_count` and separate the device logic from the thread logic. In this step, the historic behavior is preserved:`thread_count` is still always equal to `device_count`. In coming PRs, this will no longer be the case.

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -835,6 +835,11 @@ int thread_count()
     return sApp->thread_count;
 }
 
+int device_count()
+{
+    return sApp->device_count;
+}
+
 int8_t sandstone_verbosity_level()
 {
     return std::to_underlying(sApp->shmem->cfg.verbosity);
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -318,15 +318,15 @@ static void init_shmem()
             "PerThreadData::Test size grew, please check if it was intended");
     assert(sApp->current_fork_mode() != SandstoneApplication::ForkMode::child_exec_each_test);
     assert(sApp->shmem == nullptr);
-    assert(thread_count());
+    assert(device_count());
 
     unsigned per_thread_size = sizeof(PerThreadData::Main);
     per_thread_size = ROUND_UP_TO(per_thread_size, alignof(PerThreadData::Test));
-    per_thread_size += sizeof(PerThreadData::Test) * thread_count();
+    per_thread_size += sizeof(PerThreadData::Test) * device_count();
     per_thread_size = ROUND_UP_TO_PAGE(per_thread_size);
 
     unsigned thread_data_offset = sizeof(SandstoneApplication::SharedMemory) +
-            sizeof(Topology::Thread) * thread_count();
+            sizeof(Topology::Thread) * device_count();
     thread_data_offset = ROUND_UP_TO_PAGE(thread_data_offset);
 
     size_t size = thread_data_offset;
@@ -357,6 +357,7 @@ static void commit_shmem()
     size_t main_thread_count = plan.size();
     sApp->shmem->main_thread_count = main_thread_count;
     sApp->shmem->total_thread_count = thread_count();
+    sApp->shmem->total_device_count = device_count();
 
     // unmap the current area, because Windows doesn't allow us to have two
     // blocks for this file
@@ -381,7 +382,7 @@ static void commit_shmem()
     }
 
     // sApp->shmem has probably moved
-    restrict_topology({ 0, thread_count() });
+    restrict_topology({ 0, device_count() });
 }
 
 static void attach_shmem(int fd)
@@ -586,6 +587,7 @@ static int exec_mode_run(int argc, char **argv)
     attach_shmem(parse_int(argv[1]));
     device_info = sApp->shmem->device_info;
     sApp->thread_count = sApp->shmem->total_thread_count;
+    sApp->device_count = sApp->shmem->total_device_count;
     rebuild_topology();
     sApp->user_thread_data.resize(sApp->thread_count);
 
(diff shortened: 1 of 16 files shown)


framework/sandstone_opts.cpp
mcisowsk, line 893: shouldn't we now set `opts.device_count` from `-n`, instead of `opts.thread_count`? From what I understand, we plan to set `sApp->thread_count` based on subscription ratio in the follow up PRs. This would mean that we should remove all the manual setting of `thread_count` in this PR, and temporarily set it to be equal to `device_count`.

framework/sandstone.cpp
```diff
@@ -1027,12 +1034,14 @@ int main(int argc, char **argv)
     if (sApp->total_retest_count < -1 || sApp->retest_count == 0)
         sApp->total_retest_count = 10 * sApp->retest_count; // by default, 100
 
-    if (unsigned(opts.thread_count) < unsigned(sApp->thread_count))
-        restrict_topology({ 0, opts.thread_count });
+    if (unsigned(opts.thread_count) < unsigned(sApp->device_count)) {
```
mcisowsk: then here I'd expect to have
```c++
if (unsigned(opts.device_count) < unsigned(sApp->device_count))
        restrict_topology({ 0, opts.device_count});
sApp->thread_count = sApp->device_count; // temporarily it's 1:1 ratio
```

framework/sandstone_opts.hpp
```diff
@@ -23,6 +23,7 @@ struct ProgramOptions {
     const char* seed = nullptr;
     int max_cores_per_slice = 0;
     int thread_count = -1;
+    int device_count = -1;
```
mcisowsk, line 26: Given my previous comments, I'd rename `thread_count` -> `device_count`, instead of having two of them

framework/sandstone_opts.cpp
AthenasJimenez: `--threads/-n` should keep its current behavior. We still want to be able to set the number of threads, whether we use subscription or not.

For this case, yes, I could set `opts.device_count`, but I’d rather keep it as‑is since both counters will remain separated.

framework/sandstone.cpp
AthenasJimenez: Same as above.

framework/sandstone_opts.hpp
AthenasJimenez: I considered that as well, renaming everything and bringing back `thread_count` where is needed. But I decided to separate the logic/arrays/structs of `device_count` and `thread_count` in this PR. So in coming PRs I can focus on implement the subscription logic.

thiagomacieira (dismissed):

framework/sandstone_opts.hpp
thiagomacieira: I agree with Athenas. What we want is to have two variables: one for the number of threads we'll start and one for the number of devices being tested. The two do not have to be the same.

thiagomacieira (approved):

framework/sandstone_opts.hpp
mcisowsk: we change the behaviour when --cpuset and -n are passed together (now they both tune thread_count, can even be thought of as excluding args (one "overwrites" the other); afterwards they'd tune different variables and the user might get confused with the outcome)

mcisowsk, line 26: where do we set `opts.device_count`?

AthenasJimenez: The behavior when both --cpuset and -n are passed is still the same as before: --cpuset narrows the device set, -n only kicks in if it's smaller. So no user-visible change here.

```
# New behavior
./builddir-release-device_count/opendcdiag -vv -Y2 --disable=* --cpuset=0,1 -n4
  command-line: 'opendcdiag -vv -Y2 --disable=* --cpuset=0,1 -n4'
  version: opendcdiag-<unknown version>
  os: Linux 6.19.11-100.fc42.x86_64
  runtime: glibc 2.41
  openssl: { version: OpenSSL 3.2.6 30 Sep 2025 }
  cpu-info:
  - { logical: 0, package: 0, numa_node: 0, module: 0, core: 0, thread: 0, core_type: p, family: 6, model: 0xc6, stepping: 2, microcode: 0x11b, ppin: null }   # 0
  - { logical: 1, package: 0, numa_node: 0, module: 1, core: 4, thread: 0, core_type: p, family: 6, model: 0xc6, stepping: 2, microcode: 0x11b, ppin: null }   # 1
  test-plans:
    fullsocket: [ { starting_cpu: 0, count: 2 } ]
    isolate_numa: [ { starting_cpu: 0, count: 2 } ]
    heuristic: [ { starting_cpu: 0, count: 2 } ]
```

```
# Previous behavior
./builddir-release/opendcdiag -vv -Y2 --disable=* --cpuset=0,1 -n4
  command-line: 'opendcdiag -vv -Y2 --disable=* --cpuset=0,1 -n4'
  version: opendcdiag-<unknown version>
  os: Linux 6.19.11-100.fc42.x86_64
  runtime: glibc 2.41
  openssl: { version: OpenSSL 3.2.6 30 Sep 2025 }
  cpu-info:
  - { logical: 0, package: 0, numa_node: 0, module: 0, core: 0, thread: 0, core_type: p, family: 6, model: 0xc6, stepping: 2, microcode: 0x11b, ppin: null }   # 0
  - { logical: 1, package: 0, numa_node: 0, module: 1, core: 4, thread: 0, core_type: p, family: 6, model: 0xc6, stepping: 2, microcode: 0x11b, ppin: null }   # 1
  test-plans:
    fullsocket: [ { starting_cpu: 0, count: 2 } ]
    isolate_numa: [ { starting_cpu: 0, count: 2 } ]
    heuristic: [ { starting_cpu: 0, count: 2 } ]
```

As for what happens once we fully decouple them, we have a pending discussion to figure that out.

AthenasJimenez: My bad, I'll remove in a coming PR.

PR #1084, opened 2026-05-07
FW: extend `YamlFormatter::SimpleValue` with 32-bit integer variants

mcisowsk:


thiagomacieira (commented):
Syntactically correct, but why? Why can't you just extend to 64-bit?

mcisowsk:
> Syntactically correct, but why? Why can't you just extend to 64-bit?

You mean use `(u)int64_t`? I could, but let me reverse the question -- why would I cast to such unnecessarily big type? Ideally we'd want to log stuff of their natural type, I guess. 
But I won't insist on merging it.

thiagomacieira:
> > Syntactically correct, but why? Why can't you just extend to 64-bit?
> 
> You mean use `(u)int64_t`? I could, but let me reverse the question -- why would I cast to such unnecessarily big type? Ideally we'd want to log stuff of their natural type, I guess. But I won't insist on merging it.

The size of the `std::variant` won't change, it's fixed at the size of the largest type, plus the alignment (to store the control index). Adding 32-bit integers will simply add to the complexity and dispatching. On little-endian, `uint32_t` will likely be exactly the same payload. So the passing is completely unnecessary.

Is there an advantage in the test code itself? For example, are you having to explicitly cast your integers to 64-bit to avoid ambiguity errors? In that case, there's value in adding those. Or in removing the `uint64_t` because we have a limited range of what we can print anyway.

mcisowsk:
this change emerged when I had to manually cast my values to `uint64_t`, which I took as a bug

thiagomacieira:
> this change emerged when I had to manually cast my values to `uint64_t`, which I took as a bug

Not a bug, but an API inconvenience. Normally, I'd *remove* `uint64_t` because we can't print values over 2^63, but given the nature of the content we write, we need those high-bit values.

mcisowsk:
> > this change emerged when I had to manually cast my values to `uint64_t`, which I took as a bug
> 
> Not a bug, but an API inconvenience. Normally, I'd _remove_ `uint64_t` because we can't print values over 2^63, but given the nature of the content we write, we need those high-bit values.

I can live with this inconvenience, it's up to you what we do with this PR.

thiagomacieira (approved):

thiagomacieira:
> I can live with this inconvenience, it's up to you what we do with this PR.

Approved, it's worth the convenience.

PR #1085, opened 2026-05-08
FW/memfd: start using `MFD_NOEXEC_SEAL`

thiagomacieira:
The Linux kernel really wants us to explicitly declare whether we are creating a memfd that would be used for `execveat()` using `MFD_EXEC` or seal it against ever becoming executable. Depending on the `vm.memfd_noexec` sysctl setting, it would have automatically denied our call, printing

 memfd_create() requires MFD_NOEXEC_SEAL with vm.memfd_noexec=%d

Since we really want to use memfds and not regular files, let's use the new flag, but guard against using it on older kernels.

With Linux 6.6, it wouldn't deny, but would print:

 memfd_create() called without MFD_EXEC or MFD_NOEXEC_SEAL set

### DIFF
--- a/framework/sysdeps/unix/tmpfile.c
+++ b/framework/sysdeps/unix/tmpfile.c
@@ -24,6 +24,10 @@
 #    define memfd_create(name, flags)   -1
 #  endif
 #  define gettid()                      getpid()
+#  ifndef MFD_NOEXEC_SEAL
+#    define MFD_NOEXEC_SEAL 0
+#    define MFD_EXEC 0
+#  endif
 #else
 #  define gettid()                      syscall(SYS_gettid)
 
@@ -34,6 +38,10 @@
 #    endif
 #    define MFD_CLOEXEC 1U
 #    define memfd_create(name, flags)     syscall(319, name, flags)
+# endif
+# ifndef MFD_NOEXEC_SEAL
+#   define MFD_NOEXEC_SEAL 8U
+#   define MFD_EXEC 0x10U
 # endif
 #endif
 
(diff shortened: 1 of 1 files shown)


fenrus75 (approved):

framework/sysdeps/unix/tmpfile.c
```diff
@@ -35,6 +39,10 @@
 #    define MFD_CLOEXEC 1U
 #    define memfd_create(name, flags)     syscall(319, name, flags)
 # endif
+# ifndef MFD_NOEXEC_SEAL
+#   define MFD_NOEXEC_SEAL 8U
+#   define MFD_EXEC 0x10U
+# endif
```
fenrus75, line 45: **Severity: nit** -- Inconsistent preprocessor indentation in the Linux `#else` block.

The new `MFD_NOEXEC_SEAL` block uses 1 space before `ifndef`/`endif` and 3 spaces before `define`, while the existing `MFD_CLOEXEC` block directly above it uses 2 and 4 spaces respectively. Should match:

```c
#  ifndef MFD_NOEXEC_SEAL
#    define MFD_NOEXEC_SEAL 8U
#    define MFD_EXEC 0x10U
#  endif
```

(The non-Linux block at lines 27–30 already uses the correct 2/4-space style.)

fenrus75 (commented):
**Severity: low** -- Missing `[ChangeLog]` entry in the commit message.

Per `commit-message.md`, commits that fix bugs or change visible behaviour must include a `[ChangeLog][Framework]` paragraph. This PR prevents `memfd_create()` from being denied (visible: the tool would abort on systems with `vm.memfd_noexec >= 2`) and suppresses kernel log warnings on Linux 6.6+. A ChangeLog entry is expected.

Suggested addition:
```
[ChangeLog][Framework] The tool now passes MFD_NOEXEC_SEAL to
memfd_create() to avoid kernel log warnings and prevent failures on
systems configured with vm.memfd_noexec >= 2.
```

---

**Note:** The Copilot bot comment claiming that `memfd_noexec_seal` is "cached as unsupported on *any* failure" is a **false positive**. The code correctly only clears the cache on `EINVAL` -- transient errors (EMFILE, ENOMEM, EINTR) do not disable the flag for future calls.

PR #1086, opened 2026-05-11
FW/GPU: implement `IsolateNuma` slicing

mcisowsk:


framework/device/gpu/build_topology.cpp
```diff
@@ -6,6 +6,21 @@
 #include <sandstone.h>
 #include <topology_gpu.h>
 
+#include <algorithm>
+
+namespace {
+void push_to_numa_domains(std::vector<Topology::NumaNode>& numa_domains, Topology::Thread* info)
+{
+    auto domain = std::find_if(numa_domains.begin(), numa_domains.end(), [&](const auto& node) {
+        return node.id() == info->numa_id; // devices with numa_id=-1 (unknown) will be grouped together
```
mcisowsk, line 15: ... which I don't know if is correct ... `Topology::NumaNode::id()` would return `-1` for both empty and unknown domain

framework/device/gpu/build_topology.cpp
```diff
@@ -6,6 +6,21 @@
 #include <sandstone.h>
 #include <topology_gpu.h>
 
+#include <algorithm>
+
+namespace {
+void push_to_numa_domains(std::vector<Topology::NumaNode>& numa_domains, Topology::Thread* info)
```
fenrus75: The `info` parameter is declared `Topology::Thread*` (non-const) but the function only reads from the pointed-to object (`info->numa_id`) and stores it as `EndDevice = const Thread*`. Per the project coding style, parameters should be `const` when possible (low severity).

Suggested fix:
```cpp
void push_to_numa_domains(std::vector<Topology::NumaNode>& numa_domains, const Topology::Thread* info)
```

framework/device/gpu/build_topology.cpp
```diff
@@ -6,6 +6,21 @@
...
+{
+    auto domain = std::find_if(numa_domains.begin(), numa_domains.end(), [&](const auto& node) {
+        return node.id() == info->numa_id; // devices with numa_id=-1 (unknown) will be grouped together
+    });
+    if (domain == numa_domains.end()) {
+        domain = numa_domains.emplace(numa_domains.end());
+    }
+    domain->devices.push_back(info);
+};
```
fenrus75: Nit: trailing `;` after the closing `}` of the function body. A stray semicolon is legal C++ but unconventional. Please remove it.

framework/device/gpu/topology_gpu.h
```diff
@@ -45,6 +45,16 @@ class Topology
 
     std::vector<std::variant<RootDevice, EndDevice>> devices;
 
+    struct NumaNode
+    {
+        std::vector<EndDevice> devices; // here FLAT/COMPOSITE hierarchy does not matter
+        int id() const
+        {
+            return devices.size() ? devices.front()->numa_id : -1;
```
fenrus75: Nit: `devices.size() ?` should be `!devices.empty()` -- the idiomatic C++ form for a non-empty check.

fenrus75 (commented):
Reviewed the NUMA slicing implementation. Found a few minor issues; see inline comments. The overall approach is sound -- detection, grouping, and slicing logic are correct.

**Design note**: The `NumaDomainsGrouping` unit test verifies NUMA domain grouping well, but the contiguous-range detection logic in `slice_plan_init()` (splitting non-contiguous GPU index runs into separate `DeviceRange` entries) has no test coverage. A test that exercises non-contiguous GPU distributions across a NUMA domain would be a useful addition.

*Note: the four existing Copilot-generated inline comments on this PR appear to describe issues that are already addressed by the PR (mock topology initialisation, unit-test value-init, NUMA grouping test), and the FullSystem-index claim is not present in the new code. They can be dismissed.*

framework/device/gpu/build_topology.cpp
thiagomacieira: I don't see an alternative.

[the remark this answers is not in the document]
thiagomacieira: I don't know if this review is stale or not, but it's incorrect. All the unit tests use `make_gpu_info_entry`, which initialises all fields. Even if not, the suggestion is nonsensical...

thiagomacieira (approved):

PR #1089, opened 2026-05-11
Create a basic set of copilot code review rules

fenrus75:


review/framework.md
```diff
@@ -0,0 +1,13 @@
...
+
+
+Mandatory step: Load `framework/sandstone.h` immediately into the review context
+
+
+# Coding style and rules to verify as part of the review
+
+
+- [ ] The framework is written in C++ but all functions that are to be used
+    by tests must be available from the C language
```
thiagomacieira, line 11: We *have* begun adding more goodies that are C++-only...

review/general-c++.md
```diff
@@ -0,0 +1,85 @@
...
+rules override the rules below.
+
+
+# General correctness rules to verify as part of the review
+
+- [ ] Check for, and report Undefined Behavior (UB) as per the relevant C++ standard
+      (assume C++23 if no standard version is specified)
+
+
+## STL items
```
thiagomacieira: "Standard Library". Not everything is a template.

review/general-c++.md
```diff
@@ -0,0 +1,85 @@
...
+
+- [ ] Ensure that when std::vector.reserve() is used, no accesses beyond the
+  end of the actual size happen. reserve(), unlike resize(), only changes
+  the backing behind the vector, but not the size() of the vector.
```
thiagomacieira, line 23: Why do we need to emphasise this?

BTW, it also applies to `std::string`.

review/general-c++.md
```diff
@@ -0,0 +1,85 @@
...
+
+- [ ] Except for boundaries where code interfaces from or to C code, std::string usage
+    is strongly recommended for string handling.
+
+- [ ] STL containers (std::vector etc) and iterators are preferred over custom coded solution
```
thiagomacieira: `std::vector` *is* part of what used to be called "STL".

review/general-c++.md
```diff
@@ -0,0 +1,85 @@
...
+    "using namespace std".
+
+
+## Language standard
+
+- [ ] Use only constructs in the C++ 23 standard that are non-deprecated or
+    obsoleted.
+
+- [ ] Undefined behavior (UB) language constructs are a violation of coding
+    style
```
thiagomacieira, line 43: Again, by definition. So why do we need to emphasise this?

review/general-c++.md
```diff
@@ -0,0 +1,85 @@
...
+
+- [ ] Any use of fixed sized buffers (example: char foo[128]) must be
+    **provably** correct to not overflow the buffer in any condition
+
+- [ ] "using namespace std;" is a deprecated construct
+
+
+## General C++ coding rules
+
+- [ ] Use "const" when possible for method and function parameters
```
thiagomacieira, line 53: I *personally* don't like const parameters, since they aren't really const. That is:
```c++
void f(const int i);
void f(int i);
```
are the same function.

In the body of the function, it does allow the parameter to be declared const so it doesn't get modified:
```c++
// header.h
void f(int i);

// implementation.cpp:
void f(const int i)
{
    int j = i + 1;
}
```

That's different from const-qualified objects passed by reference or pointer, such as:
```c++
void f(const S &);
void f(const char *);

// implementation:
void f(const char *const ptr)
{
    // cannot change ptr
```

review/review-core.md
```diff
@@ -0,0 +1,330 @@
+# OpenDCDiag Patch Review Protocol
+
+
+## OpenDCDiag overview
+OpenDCDiag is a tool to find defects in CPU cores by running tests which
+compare the output of a computation to known good values. OpenDCDiag
+contains a combination of Apache licensed and proprietary source code.
```
thiagomacieira: No proprietary code is allowed in OpenDCDiag.

review/review-core.md
```diff
@@ -0,0 +1,330 @@
...
+
+When reviewing a pull request, get the whole pull request from GitHub, and
+note any prior review findings from this pull request for detailed analysis
+in the Phase 3 tasks.
+
+The review output file for a pull request is named `review-<PR_NUMBER>.txt`
+(e.g. `review-3397.txt`) in the current working directory. If that file exists,
+read it and take note of any prior review findings for use in the Phase 3 tasks.
+If the file does not exist locally, call `get_review(pr_number)` on the
+ci-helper MCP server — it is the canonical store and may contain prior findings
```
thiagomacieira: We won't have a ci-helper MCP server for the public.

I think this line until the end of the section are unnecessary for OpenDCDiag.

review/review-core.md
```diff
@@ -0,0 +1,330 @@
...
+Label every issue with a severity level of "none", "nit", "low", "medium", "high" or "critical"
+
+Severity "nit" means a trivial cosmetic issue that does not affect correctness
+or readability.
+
+Prepare all needed information for reporting later, by putting all findings into your
+context. In addition, write out a preliminary report file using the format
+defined in Phase 4, which can then be updated in Phase 3.
+
+# Phase 3: Deep review of context of the code
```
thiagomacieira, line 129: I'm missing a phase that reviews the commit message. The fact that GitHub does not allow reviewing commit messages in spite of this being asked for 10 years notwithstanding.


Each commit should be atomic in what it changes and should be testable (they should never break the build). Updates to the unit-tests or bats integration tests should be done in the commit that made the change.

Commit messages should reflect that and describe the reason why a change was made and the outcome/effect of that change. They should not describe the code change itself.

Commits that changed visible behaviour, fixed bugs, or added command-line options should have a `[ChangeLog][Framework]` paragraph in the message, written either in the past tense ("Fixed a bug that...", "Added command-line option that...", etc.) or in the present or future tense describing the new behaviour ("The tool can now ...", "The tool will now log...", etc.). The language in ChangeLog messages should be concise and describe the behaviour that a user will see. It should never describe internal behaviour. ChangeLog messages can be accumulated over a series of commits so only the final outcome is described.

Ideally, the PR message itself should include a copy of the ChangeLog message and a summary of the changes in all of the commits.

review/review-core.md
```diff
@@ -0,0 +1,330 @@
...
+
+Before reporting any issue:
+1. Verify the issue can actually occur in practice
+2. Trace the execution path to confirm
+
+
+## Creating pull requests
+
+Never create pull requests automatically. After completing the review, ask
+the user whether pull requests should be created for the issues found.
```
thiagomacieira, line 235: Why are we talking about creating pull requests when reviewing pull requests?

review/review-core.md
```diff
@@ -0,0 +1,330 @@
...
+
+### Step 1: Deduplicate against the MCP store
+Use the ci-helper MCP tool `get_review(pr_number)` to retrieve the
+previously stored review for this PR (if any). Parse its findings
+(file, line, severity, description) and compare against your current
+findings. Drop any finding that is substantively the same as one already
+present in the stored review, accounting for minor line number shifts due
+to subsequent commits. Where findings are inter-related or overlapping,
+consolidate them into a single issue rather than listing separately --
+keep the total number of issues to a minimum.
```
thiagomacieira: No public MCP server

review/style.md
```diff
@@ -0,0 +1,35 @@
+Count how many rules this file contains, put this number in the context,
+and at the end, use this count to make sure no checks were missed. List this
+number when reporting out.
+
+### General Coding style rules
+
+- [ ]  **C++23** (`-std=gnu++23`) and **C17** (`-std=gnu17`); GCC is required (clang may not work)
```
thiagomacieira: For OpenDCDiag: code must compile with both Clang and GCC.

review/style.md
```diff
@@ -0,0 +1,35 @@
...
+- [ ]  **4 spaces** indentation, **no tabs** anywhere in C/C++ (enforced by `.editorconfig`)
+- [ ] **`CamelCase`** for classes and structs; **`snake_case`** for functions and variables; **`ALL_CAPS`** for constants/macros
+- [ ] C++ headers: `*.hpp`; C headers: `*.h`
+- [ ]  **Anonymous namespaces** required around all test-internal structs/classes in C++. See `tests/guided_arithm_apx/guided_arithm_apx.cpp` for a good example
```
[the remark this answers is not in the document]
thiagomacieira: Indeed, update to point to another test.

review/test.md
```diff
@@ -0,0 +1,367 @@
...
+
+- [ ] Files that contain a DECLARE_TEST must be listed in
+    `tests/meson.build` so that the test is included in the application.
+
+
+- [ ] Undefined Behavior (UB): In general, UB is not permitted in OpenDCDiag
+    tests. A key exception is UB of the math/computation types (such as
+    overflow/underflow in addition) since OpenDCDiag tests regularly do this on
+    purpose. If you encounter math type UB in a PR review (but not whole file
+    review), make a note at "none" severity.
```
thiagomacieira: No, not allowed either, because the *compiler* is allowed to remove the effect in the first place.

You may only cause signed integer over/underflows in assembly code or using intrinsics.

review/test.md
```diff
@@ -0,0 +1,367 @@
...
+
+- [ ] memcmp_or_fail() has a 3rd argument that represents the count of
+    elements to compare. This is units of the size of the base argument,
+    and not bytes, unless the base type is a byte. It is a bug to pass
+    in a value that does not match the number of elements in the arrays
+    that are to be compared.
+
+- [ ] For the `memcmp_or_fail()` call that takes a string to be
+    formatted, it is a bug to have a newline ('\n') in the fmt string,
+    as the framework asserts against this.
```
thiagomacieira: Not any more. I thought I'd fixed this in the internal repository, but maybe it's not merged yet.

I need to first find out if any test still has that the newline.

review/test.md
```diff
@@ -0,0 +1,367 @@
...
+    - log_yaml
+    - report_fail_msg
+
+    `log_yaml` logs a message and extra details formatted in YAML; the
+    level argument must be one of the SANDSTONE_LOG_* string
+    constants. Code in C++ should call the `std::map` overload.
+
+    Code should construct the strings and the `std::map` in place, not
+    ahead of time, so the messages and calls are eliminated when
+    `OpenDCDiag_NO_LOGGING` is defined. Example:
```
thiagomacieira: This one is really `SANDSTONE_NO_LOGGING`.

thiagomacieira (changes requested):

review/general-c++.md
fenrus75: known actual bug somewhere -- I forgot where.

review/general-c++.md
fenrus75: by definition, but by listing it here the agent will actually check for this -- without it -- not so much

mrkz:
I have a question on this PR,

Is it intended to provide these review capabilities/rules to the automatic PR Copilot Review or only for agentic (run by users) mode?

While clearly the automatic review mode on the Web is more limited, it may be an early entry point for checks while the agentic review is done by a person.

Should the reply be positive, then the layout for the review files should be modified in 2 key items:

1) The copilot web-based review is [**NOT** able to follow instructions files](https://docs.github.com/en/copilot/how-tos/copilot-on-github/customize-copilot/add-custom-instructions/add-repository-instructions#creating-path-specific-custom-instructions) beyond what's defined in the `.github/{instructions/,copilot-instructions.md}` space. So the review files could be relocated as:
```txt
.                                                                  
├── .github
│   ├── instructions
│   │   ├── framework-instructions.md
│   │   ├── general-c++-instructions.md
│   │   ├── modernize-instructions.md
│   │   ├── review-core-instructions.md
│   │   ├── style-instructions.md
│   │   └── test-instructions.md
│   ├── copilot-instructions.md
```

2) The instructions files should be capped at `4,000` characters as that's the [max limit the copilot web-based review is able to read (per file)](https://docs.github.com/en/copilot/how-tos/use-copilot-agents/request-a-code-review/use-code-review#customizing-copilots-reviews-with-custom-instructions)

also for item 2), the review instructions can be tailored to match certain folders or file patters as part of the instruction definition, one example would be using the following in the `./github/instructions/framework-instructions.md` file:
```yaml
---
applyTo: "framework/*"
---

[...]
```

fenrus75:
I would like the full review rules for local copilot/vscode/claude/etc use

if we need to do a mini version for github UI afterwards I'm not opposed to that but should not come at the expense of having a full comprehensive rule set.

review/general-c++.md
fenrus75: fixed

review/general-c++.md
fenrus75: fixed

review/style.md
fenrus75: fixed

review/test.md
fenrus75: fixed

review/test.md
fenrus75: fixed

PR #1090, opened 2026-05-12
unittests: test `slice_plan_init` for CPUs

mcisowsk:


framework/device/cpu/slicing.cpp
```diff
@@ -0,0 +1,119 @@
+/*
```
thiagomacieira, line 1: The only thing is I'd name this file something different, not after the function in it. Like "slicing.cpp".

framework/sandstone.cpp
```diff
@@ -935,6 +935,22 @@ static bool background_scan_wait()
...
+    for (std::vector<DeviceRange> &plan : sApp->slice_plans.plans) {
+        plan.clear();
+    }
+
+    if (sApp->current_fork_mode() == SandstoneApplication::ForkMode::no_fork || max_cores_per_slice < 0) {
+        // set to full system
+        std::vector plan = { DeviceRange{ 0, thread_count() } };
+        sApp->slice_plans.plans.fill(plan);
+        return;
+    }
```
thiagomacieira, line 949: This probably won't apply to other devices. I guess  you'll refactor when the time comes.

thiagomacieira (dismissed):

framework/device/cpu/slicing.cpp
mcisowsk: there you go

dawid-sabat (dismissed):

dawid-sabat (approved):

PR #1091, opened 2026-05-12
TestRunner class facade to allow custom init/run/cleanup patterns

jposwiata:
Functions with "default" arguments (or shorted lists) might be provided. The runner is able to execute TEST_LOOP itself (if N is provided), per-cpu context is handled. TestRunnerN alias added.

### DIFF
--- a/framework/sandstone.h
+++ b/framework/sandstone.h
@@ -787,84 +787,187 @@ memcmp_or_fail(const T *actual, const T *expected, size_t count, const char *fmt
     _memcmp_or_fail_cb((actual), (expected), (size), NULL, NULL)
 #endif
 
-// Static C++ test runner to instantiate appropriate test class or always skipping one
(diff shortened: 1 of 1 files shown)


mcisowsk:
I wonder how different is this from #691

jposwiata:
> I wonder how different is this from #691

Both are very similar. The same purpose, both use C++ concepts, and both are equally dead.

The interface is far less complex ("just a basic facade", it even doesn't handle "extra" functions, just basic test interface).
This code follows "old" approach (ret codes instead of "knowledge leaking" exceptions, common TEST_LOOP instead of "own"), No "DECLARE_TEST_CLASS" too (my idea of "self containing" declaration was rejected in a former project).
This code introduces "per thread context" and a way to not infect test class with N.
Non-success result from `init()` is handled properly.

PR #1099, opened 2026-05-21
FW/logging: Fix main thread entries missing from stdout

AthenasJimenez:
Main thread entries with no messages were not printed to stdout. The doprint lambda gated both header and message output behind lowest_level <= verbosity, but main threads with no messages have lowest_level = Max (127), so the condition always failed and the stdout output was silently skipped.

Decouple header printing from message printing for the stdout path: print the thread header when want_print or has_messages is true, and print messages only when has_messages is true. This ensures main thread entries (runtime, resource-usage) appear on stdout at -vvv verbosity even when the thread has no messages.

Also add a selftest "selftest_pass has main thread at -vvv" that verifies at least one main thread entry exists with runtime and resource-usage when running at -vvv verbosity.

### DIFF
--- a/bats/sanity-check/20-selftests.bats
+++ b/bats/sanity-check/20-selftests.bats
@@ -342,6 +342,19 @@ selftest_pass() {
     test_yaml_regexp "/tests/1/result" skip
 }
 
+@test "selftest_pass has main thread at -vvv" {
+    # Verify that at least 1 main thread exists with runtime
+    # and resource-usage when running at -vvv verbosity.
+    declare -A yamldump
+    sandstone_selftest -vvv -e selftest_pass
+    [[ "$status" -eq 0 ]]
+    test_yaml_regexp "/tests/0/threads/0/thread" 'main'
+    test_yaml_numeric "/tests/0/threads/0/runtime" 'value >= 0'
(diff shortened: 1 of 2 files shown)


bats/sanity-check/20-selftests.bats
```diff
@@ -2110,3 +2110,16 @@ selftest_interrupt_common() {
     test_yaml_regexp "/tests/2/test" selftest_test_optional_include_optional
     test_yaml_regexp "/tests/3/test" selftest_test_optional_beta_include_optional
 }
+
+@test "selftest_pass has main thread at -vvv" {
```
thiagomacieira: Nitpick: move this closer to `@test "selftest_pass"`, around line 290. Do call `selftest_pass` from here too.

thiagomacieira (commented):
Just the nitpick.

bats/sanity-check/20-selftests.bats
AthenasJimenez: I tried with `selftest_pass` but checking the output was more complex, so I just stay with `sandstone_selftest`.

thiagomacieira (approved):

PR #1100, opened 2026-05-21
FW/logging: Remove redundant inner loop count log_message

AthenasJimenez:
The inner loop count is already reported in the YAML output via the per-thread loop-count field. The log_message call duplicated this information as a verbose info message, adding noise without value.

[ChangeLog][Logging] Remove redundant inner loop count log_message.

### DIFF
--- a/bats/sanity-check/20-selftests.bats
+++ b/bats/sanity-check/20-selftests.bats
@@ -342,9 +342,10 @@ selftest_pass() {
     test_yaml_regexp "/tests/1/result" skip
 }
 
-@test "selftest_pass has main thread at -vvv" {
+@test "selftest_pass has main thread and loop-count at -vvv" {
     # Verify that at least 1 main thread exists with runtime
     # and resource-usage when running at -vvv verbosity.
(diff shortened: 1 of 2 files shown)


thiagomacieira (commented):
Please extend the test you added in #1099 to confirm the entry is printed under `-vvv` even for silent threads.

The "selftest_timedpass" tests already do that, but I'd prefer it further up.

thiagomacieira (changes requested):

mcisowsk (dismissed):
An idea: add a copilot rule to require changelog section in the commit message for "critical" changes. I always forget about it myself.

AthenasJimenez:
> Please extend the test you added in #1099 to confirm the entry is printed under `-vvv` even for silent threads.
> 
> The "selftest_timedpass" tests already do that, but I'd prefer it further up.

Done

AthenasJimenez:
> An idea: add a copilot rule to require changelog section in the commit message for "critical" changes. I always forget about it myself.

I think its a good idea, please refer to #1105

thiagomacieira (approved):

PR #1101, opened 2026-05-22
FW: Decouple thread_count from device_count in slice planning and logging

AthenasJimenez:
Introduce per-slice ThreadRange alongside the existing DeviceRange so that each slice independently tracks how many threads it runs, separate from how many devices (CPUs) it covers. Currently the ratio is 1:1, but this decoupling prepares for future changes where threads != devices.

### DIFF
--- a/framework/device/cpu/slicing.cpp
+++ b/framework/device/cpu/slicing.cpp
@@ -51,14 +51,14 @@ void slice_plan_init_for_device(SlicePlans::SlicesArray& plans, int max_cores_pe
         }
 
         // set up proper plans
-        std::vector<DeviceRange> &isolate_socket = plans[SlicePlans::IsolateSockets];
-        std::vector<DeviceRange> &isolate_numa = plans[SlicePlans::IsolateNuma];
-        std::vector<DeviceRange> &split = plans[SlicePlans::Heuristic];
-        auto push_to = [](std::vector<DeviceRange> &to, auto start, auto end) {
+        auto &isolate_socket = plans[SlicePlans::IsolateSockets];
+        auto &isolate_numa = plans[SlicePlans::IsolateNuma];
+        auto &split = plans[SlicePlans::Heuristic];
+        auto push_to = [](SlicePlans::Slices &to, auto start, auto end) {
             int start_cpu = start[0].threads.front().cpu();
             int end_cpu = end[-1].threads.back().cpu();
             assert(end_cpu >= start_cpu);
-            to.push_back(DeviceRange{ start_cpu, end_cpu + 1 - start_cpu });
+            to.push_back(SlicePlans::Slice{ DeviceRange{ start_cpu, end_cpu + 1 - start_cpu }, {} });
         };
 
         for (const Topology::Package &p : topology.packages) {
@@ -99,12 +99,12 @@ void slice_plan_init_for_device(SlicePlans::SlicesArray& plans, int max_cores_pe
 
     if (max_cores_per_slice == 0) {
         // set to full system
-        std::vector plan = { DeviceRange{ 0, device_count() } };
+        SlicePlans::Slices plan = { SlicePlans::Slice{ DeviceRange{ 0, device_count() }, {} } };
         plans.fill(plan);
     } else {
         // dumb plan, not *cores*
         int slice_count = (max_cpu - 1) / max_cores_per_slice + 1;
-        std::vector<DeviceRange> plan;
+        SlicePlans::Slices plan;
         plan.reserve(slice_count);
 
         int slice_size = max_cpu / slice_count;
@@ -112,8 +112,8 @@ void slice_plan_init_for_device(SlicePlans::SlicesArray& plans, int max_cores_pe
             ++slice_size;       // round up the slice size
         int cpu = 0;
         for ( ; cpu < max_cpu - slice_size; cpu += slice_size)
-            plan.push_back(DeviceRange{ cpu, slice_size });
-        plan.push_back(DeviceRange{ cpu, max_cpu - cpu });
+            plan.push_back(SlicePlans::Slice{ DeviceRange{ cpu, slice_size }, {} });
+        plan.push_back(SlicePlans::Slice{ DeviceRange{ cpu, max_cpu - cpu }, {} });
         plans.fill(plan);
     }
 }
--- a/framework/device/cpu/slicing.cpp
+++ b/framework/device/cpu/slicing.cpp
@@ -117,3 +117,11 @@ void slice_plan_init_for_device(SlicePlans::SlicesArray& plans, int max_cores_pe
         plans.fill(plan);
     }
 }
+
+void slice_plan_init_for_threads(SlicePlans::SlicesArray& plans)
+{
+    for (auto &plan : plans) {
+        for (auto &slice : plan)
+            slice.thread_range = { slice.device_range.starting_device, slice.device_range.device_count }; // 1:1 for now...
+    }
+}
--- a/framework/device/gpu/build_topology.cpp
+++ b/framework/device/gpu/build_topology.cpp
@@ -55,7 +55,7 @@ Topology build_topology()
 
 void slice_plan_init_for_device(SlicePlans::SlicesArray& plans, int max_cores_per_slice)
 {
-    std::vector plan = { DeviceRange{ 0, thread_count() } };
+    SlicePlans::Slices plan = { SlicePlans::Slice{ DeviceRange{ 0, thread_count() }, {} } };
     plans[SlicePlans::IsolateSockets] = plan;
 
     auto& isolate_numa = plans[SlicePlans::IsolateNuma];
@@ -77,12 +77,12 @@ void slice_plan_init_for_device(SlicePlans::SlicesArray& plans, int max_cores_pe
             }
 
             // end of contiguous range
-            isolate_numa.push_back(DeviceRange{ range_start, prev_gpu - range_start + 1 });
+            isolate_numa.push_back(SlicePlans::Slice{ DeviceRange{ range_start, prev_gpu - range_start + 1 }, {} });
             range_start = prev_gpu = gpu;
         }
 
         // include open tailing range
-        isolate_numa.push_back(DeviceRange{ range_start, prev_gpu - range_start + 1 });
+        isolate_numa.push_back(SlicePlans::Slice{ DeviceRange{ range_start, prev_gpu - range_start + 1 }, {} });
     }
 
     if (isolate_numa.empty()) [[unlikely]] {
@@ -147,7 +147,7 @@ void slice_plan_init_for_device(SlicePlans::SlicesArray& plans, int max_cores_pe
 
             if (!contiguous || range_count + g.count > max_devices_per_slice) {
                 // commit the slice
-                heuristic.push_back({ range_start, range_count });
+                heuristic.push_back(SlicePlans::Slice{ DeviceRange{ range_start, range_count }, {} });
                 range_start = g.start;
                 range_count = g.count;
             } else {
@@ -156,7 +156,7 @@ void slice_plan_init_for_device(SlicePlans::SlicesArray& plans, int max_cores_pe
             }
         }
         // end of devices for this NUMA node, close the slice
-        heuristic.push_back({ range_start, range_count });
+        heuristic.push_back(SlicePlans::Slice{ DeviceRange{ range_start, range_count }, {} });
     }
 
     if (heuristic.empty()) [[unlikely]] {
--- a/framework/device/gpu/build_topology.cpp
+++ b/framework/device/gpu/build_topology.cpp
@@ -163,3 +163,11 @@ void slice_plan_init_for_device(SlicePlans::SlicesArray& plans, int max_cores_pe
         heuristic = plan;
     }
 }
+
+void slice_plan_init_for_threads(SlicePlans::SlicesArray& plans)
+{
+    for (auto &plan : plans) {
+        for (auto &slice : plan)
+            slice.thread_range = { slice.device_range.starting_device, slice.device_range.device_count }; // 1:1 for now...
+    }
+}
--- a/framework/device/gpu/unit-tests/topology_tests.cpp
+++ b/framework/device/gpu/unit-tests/topology_tests.cpp
@@ -214,12 +214,11 @@ TEST(Topology, SlicePlansCreation)
     device_info = gpu_info.data();
     topo_global = build_topology();
 
-    auto expect_plan = [](const auto& actual, const std::vector<std::pair<int, int>>& expected) {
+    auto expect_plan = [](const auto& actual, const std::vector<DeviceRange>& expected) {
         ASSERT_EQ(actual.size(), expected.size());
-        auto expected_it = expected.begin();
-        for (size_t i = 0; i < actual.size(); ++i, ++expected_it) {
-            EXPECT_EQ(actual[i].starting_device, expected_it->first);
-            EXPECT_EQ(actual[i].device_count, expected_it->second);
+        for (size_t i = 0; i < actual.size(); ++i) {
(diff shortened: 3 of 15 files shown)


framework/logging.cpp
```diff
@@ -2280,29 +2282,34 @@ void YamlLogger::print_header(std::string_view cmdline, Duration test_duration,
...
+            result += ", threads: ";
+            result += std::to_string(tplan[i].thread_count);
             result += " }";
         }
         return result;
     };
     const std::vector<DeviceRange> &fullsocket = sApp->slice_plans.plans[SlicePlans::IsolateSockets];
     const std::vector<DeviceRange> &isolatenuma = sApp->slice_plans.plans[SlicePlans::IsolateNuma];
     const std::vector<DeviceRange> &heuristic = sApp->slice_plans.plans[SlicePlans::Heuristic];
+    const std::vector<ThreadRange> &fullsocket_t = sApp->slice_plans.thrd_plans[SlicePlans::IsolateSockets];
```
mcisowsk: `something_t` can be mistaken with a type, maybe `fullsocket_trange` and `fullsocket_drange`, or something along these lines?

framework/sandstone_p.h
```diff
@@ -475,6 +475,19 @@ template <typename Lambda> static void for_each_test_thread(Lambda &&l)
         l(sApp->test_thread_data(i), i);
 }
 
+template <typename Lambda> static void for_each_test_thread_in_slice(Lambda &&l, int slice)
+{
+    auto &range = sApp->main_thread_data(slice)->device_range;
+    auto &trange = sApp->main_thread_data(slice)->thread_range;
+
+    for (int i = 0; i < trange.thread_count ; i++) {
```
mcisowsk: loose suggestion
```c++
    for (int t = trange.starting_thread, d = range.starting_device; t < trange.thread_count; t++, d++) {
        l(sApp->test_thread_data(t), t, d);
    }
```

framework/sandstone.cpp
```diff
@@ -950,14 +950,20 @@ static void slice_plan_init(int max_cores_per_slice)
...
+        std::vector plan = { DeviceRange{ 0, device_count() } };
         sApp->slice_plans.plans.fill(plan);
         return;
     }
 
     slice_plan_init_for_device(sApp->slice_plans.plans, max_cores_per_slice);
 }
 
+static void thread_plan_init()
+{
```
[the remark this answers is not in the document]
AthenasJimenez: Fix it in `slice_plan_init_for_thread`. Check below comment.

framework/device/cpu/slicing.cpp
```diff
@@ -117,3 +117,15 @@ void slice_plan_init_for_device(SlicePlans::SlicesArray& plans, int max_cores_pe
...
+void slice_plan_init_for_threads(SlicePlans::SlicesArray& plans, SlicePlans::ThrdSlicesArray& thrd_plans)
+{
+    for (size_t p = 0; p < plans.size(); ++p) {
+        const auto &plan = plans[p];
+        auto &tplan = thrd_plans[p];
+
+        for (size_t i = 0; i < plan.size(); ++i) {
+            tplan.push_back(ThreadRange{ plan[i].starting_device, plan[i].device_count }); // 1:1 for now...
+        }
+    }
```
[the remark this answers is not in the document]
AthenasJimenez: I believe this function is only called once during execution. That said, since `DeviceRange` clears the entries, I’ll add the same behavior for `ThreadRange` as well.

framework/device/gpu/build_topology.cpp
```diff
@@ -163,3 +163,15 @@ void slice_plan_init_for_device(SlicePlans::SlicesArray& plans, int max_cores_pe
...
+void slice_plan_init_for_threads(SlicePlans::SlicesArray& plans, SlicePlans::ThrdSlicesArray& thrd_plans)
+{
+    for (size_t p = 0; p < plans.size(); ++p) {
+        const auto &plan = plans[p];
+        auto &tplan = thrd_plans[p];
+
+        for (size_t i = 0; i < plan.size(); ++i) {
+            tplan.push_back(ThreadRange{ plan[i].starting_device, plan[i].device_count }); // 1:1 for now...
+        }
+    }
```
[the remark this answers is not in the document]
AthenasJimenez: Ditto

framework/sandstone_run.cpp
```diff
@@ -1420,12 +1420,16 @@ static int slices_for_test(const struct test *test)
...
         return 1;
     }
 
     const std::vector<DeviceRange> &plan = sApp->slice_plans.plans[type];
-    for (size_t i = 0; i < plan.size(); ++i)
+    const std::vector<ThreadRange> &tplan = sApp->slice_plans.thrd_plans[type];
+    for (size_t i = 0; i < plan.size(); ++i) {
         sApp->main_thread_data(i)->device_range = plan[i];
+        sApp->main_thread_data(i)->thread_range = tplan[i];
+    }
```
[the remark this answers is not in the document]
AthenasJimenez: Fixed with an assertion.

framework/logging.cpp
```diff
@@ -2208,23 +2208,25 @@ void YamlLogger::print_thread_messages()
...
                 print_one_thread_messages(real_stdout_fd, r, sApp->shmem->cfg.verbosity);
         }
 
         munmap_and_truncate_log(data, r);
     };
+
     for_each_main_thread(doprint, slices.size());
-    for_each_test_thread(doprint);
+    for (size_t s = 0; s < slices.size(); ++s)
+        for_each_test_thread_in_slice(doprint, s);
```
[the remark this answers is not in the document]
AthenasJimenez: Fixed in `run_one_test_init_in_parent` to set proper ranges.

framework/sandstone_p.h
AthenasJimenez: I think I’ll keep it as is for now.

framework/logging.cpp
AthenasJimenez: Fixed

framework/topology.h
```diff
@@ -210,6 +210,9 @@ struct SlicePlans
     using Slices = std::vector<DeviceRange>;
     using SlicesArray = std::array<Slices, 3>;
     SlicesArray plans;
+    using ThrdSlices = std::vector<ThreadRange>;
```
thiagomacieira: Please don't abbreviate.

framework/device/gpu/build_topology.cpp
```diff
@@ -163,3 +163,17 @@ void slice_plan_init_for_device(SlicePlans::SlicesArray& plans, int max_cores_pe
...
+        const auto &plan = plans[p];
+        auto &tplan = thrd_plans[p];
+
+        tplan.clear();
+        tplan.reserve(plan.size());
+        for (size_t i = 0; i < plan.size(); ++i) {
+            tplan.push_back(ThreadRange{ plan[i].starting_device, plan[i].device_count }); // 1:1 for now...
+        }
+    }
+}
```
thiagomacieira, line 173: Duplicate code.

Does it make sense to deduplicate by moving this to `thread_plan_init()` in sandstone.cpp, for when the device did not fill in the plans? See suggestion below.

framework/sandstone.cpp
```diff
@@ -950,14 +950,20 @@ static void slice_plan_init(int max_cores_per_slice)
...
+    sApp->thread_count = device_count();
+    slice_plan_init_for_threads(sApp->slice_plans.plans, sApp->slice_plans.thrd_plans);
```
thiagomacieira:
```suggestion
    slice_plan_init_for_threads(sApp->slice_plans.plans, sApp->slice_plans.thrd_plans);
    if (sApp->slice_plans.thrd_plans[SlicePlans::FullSystem].empty()) {
        // set to 1:1
    }
```

framework/topology.h
```diff
@@ -216,6 +216,7 @@ struct SlicePlans
 };
 
 void slice_plan_init_for_device(SlicePlans::SlicesArray& plans, int max_cores_per_slice);
+void slice_plan_init_for_threads(SlicePlans::SlicesArray& plans, SlicePlans::ThrdSlicesArray& thrd_plans);
```
thiagomacieira: Are the called functions expected/allowed to modify `plans`? If not, make it a `const` ref.

framework/test_data.h
```diff
@@ -19,6 +19,13 @@ struct DeviceRange
     int device_count;
 };
 
+struct ThreadRange
+{
+    // a contiguous range
+    int starting_thread;
+    int thread_count;
+};
```
thiagomacieira, line 27: Do you predict a need for the thread numbers not to be contiguous? Or is having it precalculated helpful in other situations? Otherwise, we could just move the `thread_count` into `DeviceRange` above.

I *think* the thread offset adjustment in the slice start would benefit from having it precalculated.

framework/topology.h
```diff
@@ -210,6 +210,9 @@ struct SlicePlans
     using Slices = std::vector<DeviceRange>;
     using SlicesArray = std::array<Slices, 3>;
     SlicesArray plans;
+    using ThrdSlices = std::vector<ThreadRange>;
+    using ThrdSlicesArray = std::array<ThrdSlices, 3>;
+    ThrdSlicesArray thrd_plans;
```
thiagomacieira: Instead of struct-of-arrays, it might be better to have an array-of-structs:
```suggestion
    struct Slice {
        DeviceRange device_range;
        ThreadRange thread_range;
    };
    using Slices = std::vector<Slice>;
    using SlicesArray = std::array<Slices, 3>;
```

framework/logging.cpp
thiagomacieira: I'm not convinced about this. You didn't modify or remove `for_each_test_thread()`, which indicates it's still in use. But won't every use of it also need the device and thread ID separation?

thiagomacieira (changes requested):

framework/test_data.h
mcisowsk: if we decide to use `DeviceRange` in both places maybe it's time to rename it to `ContiguousRange`? Or some other less finger breaking name

thiagomacieira: Choice of the implementer here. There's also value in having two different structs even if they are otherwise identical: it makes assigning one to the other impossible, removing a class of errors.

AthenasJimenez: In the end, I decided to keep two separate structs. The duplication is minimal (just a few lines), and I think it helps avoid confusion between the two.

framework/topology.h
AthenasJimenez: Agree, implemented.

framework/logging.cpp
AthenasJimenez: I took a deeper look at the logging part, and I think it makes sense to keep both. There are cases like `logging_init_global()` and `logging_close_global()` where slices and devices don’t really matter.

Also, both depend on the total thread_count, not on any specific slice plan. This becomes important with thread ratios, since different slice plans can end up with different total thread counts. In those cases, the framework will stick with the highest one to make sure all threads always get their own file descriptor.

framework/topology.h
AthenasJimenez: You are right, but no longer applicable in the current version.

framework/logging.cpp
thiagomacieira: Ok, I see it's iterating differently: it will skip over thread entries that were not run. So it is different.

framework/sandstone_p.h
```diff
@@ -475,6 +475,21 @@ template <typename Lambda> static void for_each_test_thread(Lambda &&l)
         l(sApp->test_thread_data(i), i);
 }
 
+template <typename Lambda> static void for_each_test_thread_in_slices(Lambda &&l, int max_slices = INT_MAX)
```
thiagomacieira: I might give this a bit of thinking after you merge and the thread-ratio functionality is in. It seems to me the major difference is only *which* slice plan we're using. For the one above, it's a plan that includes all devices and threads, on 1:1 mapping.

We can also use C++20 concepts to detect what the lambda callback supports:
```c++
        Main *data = main_thread_data(i);
        if constexpr (requires { l(data, 0, 0); })
            l(data, -1 - i, deviceid);
        else
            l(data, -1 - i);
```

framework/logging.cpp
```diff
@@ -1861,7 +1861,7 @@ void YamlLogger::print_thread_header(int fd, PerThreadData::Common *data, int de
 #  endif
     };
 #endif
-    dprintf(fd, "%s  - thread: %d\n", indent_spaces().data(), device);
+    dprintf(fd, "%s  - thread: %d\n", indent_spaces().data(), thread);
```
thiagomacieira, line 1865: Does it make sense to insert here:
```c++
    assert(device >= 0);
```
?

thiagomacieira (changes requested):
Changes are ok, but incomplete. Please do a `ninja unittests` in your build and fix the issues there.

framework/device/gpu/unit-tests/topology_tests.cpp
thiagomacieira: Maybe:
```suggestion
    auto expect_plan = [](const auto& actual, const std::vector<DeviceRange>& expected) {
```

You probably won't need to have the type name in the rest of the changes below.

framework/logging.cpp
AthenasJimenez: I think is worth it, added.

framework/sandstone_p.h
AthenasJimenez: I agree there’s probably a better solution, but for now this is the best one I’ve found. Happy to revisit if we come up with something better.

framework/device/gpu/unit-tests/topology_tests.cpp
AthenasJimenez: Yes, fixed.

framework/sandstone_p.h
AthenasJimenez: With latest changes, we will no longer need 2 indexes.

framework/sandstone_p.h
AthenasJimenez: I added an extra commit with a proposed solution — if it looks good, I can squash it into the appropriate place.

I had to keep the loop over `thread_count()` mostly because of `logging_init_global()` in logging.cpp. This function is called during `main()`, and at that point the framework has already set `sApp->main_thread_data_ptr` using the most detailed slice plan (usually heuristic), but the structs are still empty, so `ThreadRange` isn’t populated yet.

I considered populating `ThreadRange` there and removing the loop over `thread_count()`, but that will run into issues once `thread_ratio` is introduced. The most detailed slice plan may not reflect the maximum `thread_count` that will be used.

For example, with a thread_ratio of 100:1 on a 24 logical core system, you might get something like:

```yaml
test-plans:
  fullsocket: [ { starting_cpu: 0, count: 24, threads: 23 } ]
  isolate_numa: [ { starting_cpu: 0, count: 8, threads: 7 }, { starting_cpu: 8, count: 16, threads: 15 } ]
  heuristic: [ { starting_cpu: 0, count: 8, threads: 7 }, { starting_cpu: 8, count: 16, threads: 15 } ]
```
fullsocket max thread_count: 23
heuristic max thread count: 15 + 7 = 22

In those cases, relying on the slice plan wouldn’t guarantee we capture the highest thread_count.

AthenasJimenez: Also, there is also the option of populate the first element in `sApp->main_thread_data_ptr` with the `thread_count()` and leave the rest empty to trick `logging_init_global()`. What do you think?

thiagomacieira: I'm not sure this is the correct change yet, but the change is smaller, so let's go with it. There are a number of solutions for `logging_init_global()`, so whichever you think is most appropriate.

AthenasJimenez: I’ve been digging into this for a while and couldn’t come up with a better option so far. Happy to revisit in the future, but at the moment keeping both seems like the most straightforward solution.

thiagomacieira (approved):

PR #1104, opened 2026-05-26
FW: change alignment of `tests` section

mcisowsk:
`DECLARE_TEST_INNER2` places each `struct test` in section `tests` with alignment `alignof(void*)` (8 bytes on x86_64).
    `struct test` contains `device_features_t` fields (`compiler_minimum_device` and `minimum_cpu`), which is 128-bit in CPU/GPU builds.
    In GPU debug builds builds with clang segfaults were observed, when accessing `test->compiler_minimum_device/minimum_cpu`
    due to misalignment of entries inside `tests` section (8-byte aligned instead of 16-byte aligned). Most probably GCC
    does not emmit aligned SIMD load instructions, since we did not observe crashes before.
    This commit changes the alignment to be `alignof(struct test)`. We do not remove it, as we do not want compiler to overalign.

### DIFF
--- a/framework/sandstone.h
+++ b/framework/sandstone.h
@@ -150,8 +150,9 @@ typedef enum TestQuality {
 /// used in a test's test_init function to indicate that a test should be skipped.
 #define EXIT_SKIP               -255
 
+/// force the alignment so the compiler won't try to (un)helpfully overalign it.
 #define DECLARE_TEST_INNER2(test_id, test_description) \
-    __attribute__((aligned(alignof(void*)), used, section(SANDSTONE_SECTION_PREFIX "tests"))) \
+    __attribute__((aligned(alignof(struct test)), used, section(SANDSTONE_SECTION_PREFIX "tests"))) \
     struct test _test_ ## test_id = {                   \
         .compiler_minimum_device = device_compiler_features,   \
         .id = SANDSTONE_STRINGIFY(test_id),             \


thiagomacieira (commented):
Interesting. See https://gcc.godbolt.org/z/Tref7o5YG:
Clang dislikes lowering the alignment requirements with `alignas(void *)`, but GCC allows it. Both allow it using `__attribute__` because it *does* lower the alignment.

thiagomacieira:
> Interesting. See https://gcc.godbolt.org/z/Tref7o5YG: Clang dislikes lowering the alignment requirements with `alignas(void *)`, but GCC allows it. Both allow it using `__attribute__` because it _does_ lower the alignment.

Actually, it doesn't look like it does:
https://gcc.godbolt.org/z/fd35c6rM9

Both GCC and Clang make it 16. But there's a reason why that `__attribute__` is there....

framework/sandstone.h
```diff
@@ -151,7 +151,7 @@ typedef enum TestQuality {
 #define EXIT_SKIP               -255
 
 #define DECLARE_TEST_INNER2(test_id, test_description) \
-    __attribute__((aligned(alignof(void*)), used, section(SANDSTONE_SECTION_PREFIX "tests"))) \
+    __attribute__((aligned(alignof(device_features_t)), used, section(SANDSTONE_SECTION_PREFIX "tests"))) \
```
thiagomacieira: `device_features_t` may be less than `uint64_t` if simd.conf is configured for 32 or less features, or whatever equivalent may be for the device in question. So this isn't a good check.

Let's make it simply:
```suggestion
    __attribute__((aligned(alignof(struct test)), used, section(SANDSTONE_SECTION_PREFIX "tests"))) \
```
Please add a comment above the macro to indicate we must force the alignment so the compiler won't try to (un)helpfully overalign it.

thiagomacieira:
> > Interesting. See https://gcc.godbolt.org/z/Tref7o5YG: Clang dislikes lowering the alignment requirements with `alignas(void *)`, but GCC allows it. Both allow it using `__attribute__` because it _does_ lower the alignment.
> 
> Actually, it doesn't look like it does: https://gcc.godbolt.org/z/fd35c6rM9
> 
> Both GCC and Clang make it 16.

Ok, bad test. When the `__attribute__` is on the variable declaration, it can lower the alignment.

> But there's a reason why that `__attribute__` is there....

But the question is why I added that in the first place. It was added in the old repository in:
```
commit 37af20c8a5f48c09b909734ff68cb3bc2429efec
Author: Thiago Macieira <thiago.macieira@intel.com>
Date:   Fri Feb 21 22:21:23 2020 -0800

    Move the test.h content into each of the test sources
```

Prior to that, each test directory had a test.h file, which a perl script collected and wrote an array for:
```perl
...
# The array
print("\nstatic __attribute__((used, section(\"tests\"))) struct test regular_tests_array[] = {\n");
map { printf("#include \"../%s\"\n", $_) } @tests;
print <<EOF
};
EOF
```

That got replaced with our familiar:
```diff
diff --git a/framework/sandstone.h b/framework/sandstone.h
index ef024d74a..fcd421f13 100644
--- a/framework/sandstone.h
+++ b/framework/sandstone.h
@@ -78,6 +78,11 @@ extern "C" {
 
 #define EXIT_SKIP               -255
 
+#define DECLARE_TESTS           DECLARE_TESTS2(__COUNTER__)
+#define DECLARE_TESTS2(id)      DECLARE_TEST(id [])
+#define DECLARE_TEST(id)        \
+    static __attribute__((aligned(alignof(void*)), used, section("tests"))) struct test _test_ ## id =
+
 #define RCLOBBEREDLIST "r8",\
                        "r9",\
                        "r10",\
```

thiagomacieira:
> > But there's a reason why that `__attribute__` is there....
> 
> But the question is why I added that in the first place. It was added in the old repository in:

I think the issue is I had to lower the alignment *back* to the struct's alignment. Without the attribute, GCC would sometimes emit a larger alignment.

Yup, here's an example: https://gcc.godbolt.org/z/35TbG67MY. Note how GCC emits:
```asm
        .section        test,"aw"
        .align 32
        .type   s, @object
        .size   s, 112
s:
        .zero   112
```
The structure has a suitable alignment, but the *section* is overaligned.

thiagomacieira (changes requested):

thiagomacieira (approved):

PR #1105, opened 2026-05-26
Review rules: Add ChangeLog tag review rules

AthenasJimenez:
Add a new review instruction file that checks whether commits with user-visible impact include a [ChangeLog] tag in their commit message.

Thank you @mcisowsk for suggesting this.

### DIFF
--- a/.github/instructions/changelog.instructions.md
+++ b/.github/instructions/changelog.instructions.md
@@ -1,0 +1,62 @@
+# ChangeLog tag rules
+
+When reviewing commits or pull requests, check whether the change is
+"critical" — i.e. it has an impact on user-visible behavior — and
+suggest that the commit message includes a `[ChangeLog]` tag if one is
+missing.
+
+## What counts as a critical change
+
+A change is critical and **must** carry a `[ChangeLog]` tag when it does
+any of the following:
+
+- [ ] Introduces a new command-line option or changes/removes an existing one
+- [ ] Changes the default value of an existing option or behavior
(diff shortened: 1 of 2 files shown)


.github/instructions/changelog.instructions.md
```diff
@@ -0,0 +1,62 @@
...
+## Tag format
+
+The tag is placed in the commit message body (not the subject line) on
+its own line, wrapped in square brackets:
+
+```
+[ChangeLog] Brief description of the user-visible change.
+```
+
+An optional scope can be added to categorize the change:
```
thiagomacieira, line 43: The scope is not optional.

thiagomacieira (changes requested):

.github/instructions/changelog.instructions.md
```diff
@@ -0,0 +1,62 @@
...
+"critical" — i.e. it has an impact on user-visible behavior — and
+suggest that the commit message includes a `[ChangeLog]` tag if one is
+missing.
+
+## What counts as a critical change
+
+A change is critical and **must** carry a `[ChangeLog]` tag when it does
+any of the following:
+
+- [ ] Introduces a new command-line option or changes/removes an existing one
```
thiagomacieira, line 13: Only when the option is meant to be used by the user. When it's something for debugging or experimental, not so much. Since that's subjective, I don't think we can have this.

thiagomacieira (changes requested):

PR #1106, opened 2026-05-27
FW: Report the tile ID too

thiagomacieira:


mcisowsk (approved):

framework/device/cpu/cpu_device.cpp
```diff
@@ -42,11 +42,11 @@ void dump_device_info()
     printf("Detected CPU: %s; family-model-stepping (hex): %02x-%02x-%02x; CPU features: %s\n",
            detected, sApp->hwinfo.family, sApp->hwinfo.model, sApp->hwinfo.stepping,
            device_features_to_string(device_features).c_str());
-    printf("# CPU\tPkgID\tCoreID\tThrdID\tModId\tNUMAId\tApicId\tMicrocode\tPPIN\n");
+    printf("# CPU\tPkgID\tCoreID\tThrdID\tModId\tTileId\tNUMAId\tApicId\tMicrocode\tPPIN\n");
```
mcisowsk, line 45: don't we care if it's `<0` in `--dump-device-info`?

thiagomacieira: Since it's a column-driven format, we'd have to scan all entries to see if any of them is -1 and would thus be skipped. It's also hard to write parsers with `awk` and such if the number of columns change depending on the system.

PR #1107, opened 2026-05-27
FW/Opts: replace three booleans for `-l` and `--list-*` with an enum

thiagomacieira:
I'm going to add another entry.

### DIFF
--- a/framework/sandstone_opts.hpp
+++ b/framework/sandstone_opts.hpp
@@ -24,14 +24,17 @@ struct ProgramOptions {
     int max_cores_per_slice = 0;
     int device_count = -1;
     bool fatal_errors = false;
+    enum class ListMode : uint8_t {
+        // action must be list_tests for these
+        WithDescription,
+        RawTests,
+        RawGroups,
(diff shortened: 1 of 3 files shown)


framework/sandstone_opts.hpp
```diff
@@ -29,9 +29,12 @@ struct ProgramOptions {
     std::vector<const char*> deviceset;
 
     std::string list_group_name; // for list_group
-    bool list_tests_include_descriptions = false; // for list_tests
-    bool list_tests_include_tests = false; // for list_tests
-    bool list_tests_include_groups = false; // for list_tests
+    enum {
```
mcisowsk: nit: `enum class : uint8_t` maybe?

mcisowsk (dismissed):

framework/sandstone_opts.hpp
thiagomacieira: Sure.

thiagomacieira: Done.

mcisowsk (approved):

PR #1113, opened 2026-05-29
FW: Add short test IDs

thiagomacieira:
This PR adds a short ID to each test, a 24-bit code. This is added as a pre-calculated value (with a compile-time script) so we can declare a variable with the short ID in the name, aliased to the same test variable. That will cause the linker to fail in case there is a collision. For the tests not declared using `DECLARE_TEST`, we have to do this manually: a) the name `mce_check` is injected by the script so we can use the ID it calculates, and b) for the selftests we calculate on start.

This also adds a framework feature "no-test-names", which will be used by the Intel DCDiag tool, and replaces the hack currently in use in that tool's build to achieve the same.

The `--list-test-ids` option can be used to get the list of IDs from the built tool:
```
$ ./opendcdiag --list-test-ids
smi_count                                e4f3fe
ifs                                      45c8be
ifs_array_bist                           055af9
openssl_sha                              90484c
eigen_gemm_double14                      8b6603
eigen_gemm_cdouble_dynamic_square        9dc26a
eigen_gemm_double_dynamic_square         11a32d
eigen_gemm_float_dynamic_square          4e9b47
eigen_sparse                             f994e1
eigen_svd                                628745
eigen_svd_cdouble_noavx512               064d90
eigen_svd_double                         105e35
eigen_svd_double2                        739cfc
eigen_svd_fvectors                       294f2e
zstd_aaa                                 03fbfa
zstd1                                    0aa453
zstd                                     242dad
zstd19                                   bce17b
zfuzz                                    da94e3
zlib_aaa                                 4784b9
zlib1                                    99e044
zlib                                     0ea55c
zlib9                                    0b6fbc
eigen_svd_cdouble                        f9d743
```


framework/sandstone.h
```diff
@@ -331,6 +340,11 @@ struct test {
     /// filled in by the DECLARE_TEST macro
     device_features_t compiler_minimum_device;
 
+    /// Short identifier of the test. Unique, generated.
+    uint32_t shortid;
```
mcisowsk, line 344: why not `short_id`?

framework/sandstone.h
```diff
@@ -331,6 +340,11 @@ struct test {
...
+    /// Flags stored by the DECLARE_TEST macro. Currently unused.
+    uint32_t macroflags;
```
mcisowsk: is introducing `macroflags` needed in this PR?

framework/sandstone.h
thiagomacieira: No reason, that's the first one that came to mind.

framework/sandstone.h
thiagomacieira: The padding gap will be there anyway (for CPUs at least, with `device_features_t` being 128 bits), so I added it to avoid leaving unpadded. One future use could be the `TestRun` class like #1091, where we may have a completely different layout.

I wonder if we can force the alignment down. Then we won't need it, at least not right now.

mcisowsk: I'd much rather prefer padding than some random member that we're not even sure we need. If we want to have perfectly aligned struct, let's just rearrange the members order, for example move `device_features_t minimum_cpu;` higher, etc.

thiagomacieira: We can't rearrange everything: the ones set by the `DECLARE_TEST` macro must come at the top, so I can only rearrange `compiler_minimum_device`, `shortid`, `id` and `description`. That's  128+32+64+64 bits, so there will always be 32 bits unused somewhere.

After removing the unused macroflags, here's pahole's report for the CPU build:
```c
struct test {
        device_features_t          compiler_minimum_device; /*     0    16 */
        uint32_t                   shortid;              /*    16     4 */

        /* XXX 4 bytes hole, try to pack */

        const char  *              id;                   /*    24     8 */
        const char  *              description;          /*    32     8 */
        const struct test_group  * const * groups;       /*    40     8 */
        initfunc                   test_preinit;         /*    48     8 */
        initfunc                   test_init;            /*    56     8 */
        /* --- cacheline 1 boundary (64 bytes) --- */
        runfunc                    test_run;             /*    64     8 */
        cleanupfunc                test_cleanup;         /*    72     8 */
        cleanupfunc                test_postcleanup;     /*    80     8 */
        kvmconfigfunc              test_kvm_config;      /*    88     8 */
        device_features_t          minimum_cpu;          /*    96    16 */
        int                        desired_duration;     /*   112     4 */
        int                        minimum_duration;     /*   116     4 */
        int                        maximum_duration;     /*   120     4 */
        int                        fracture_loop_count;  /*   124     4 */
        /* --- cacheline 2 boundary (128 bytes) --- */
        test_quality               quality_level;        /*   128     4 */
        test_flags                 flags;                /*   132     4 */
        void *                     data;                 /*   136     8 */
        struct test_data_per_thread * per_thread;        /*   144     8 */

        /* size: 160, cachelines: 3, members: 20 */
        /* sum members: 148, holes: 1, sum holes: 4 */
        /* padding: 8 */
        /* last cacheline: 32 bytes */
};
```

mcisowsk (approved):

PR #1116, opened 2026-06-02
FW: add missing `break` in `list_tests`

mcisowsk:


framework/sandstone.cpp
```diff
@@ -479,6 +479,7 @@ static void list_tests(const ProgramOptions& opts)
             break;
         case ProgramOptions::ListMode::RawGroups:
             printf("@%s\n", g.definition->id);
+            [[fallthrough]];
```
thiagomacieira: This should be `break`. My mistake, sorry.

mcisowsk: then I'll add a `break;` instead

lukasz-tuz (approved):

PR #1117, opened 2026-06-02
FW: do not print anything in `test_formatter`

mcisowsk:
From what I understand, its main purpose is to see if the callback behaves correctly for corner case indices (possible OOB faults). For that we do not need to print anything. It was confusing to see quite a lot of messages polluting the log even though there was no error, during tests development.

_I might be missing the point, so this PR is rather a proposal, than a strong statement._

### DIFF
--- a/framework/sandstone_run.cpp
+++ b/framework/sandstone_run.cpp
@@ -236,18 +236,14 @@ bool _memcmp_or_fail_check_fmt(const char *fmt, ...)
 
 bool SandstoneMemcmpOrFail::test_formatter(std::function<std::string ()> cb, size_t)
 {
-    log_yaml(SANDSTONE_LOG_DEBUG, ("memcmp_or_fail: " + cb()).c_str());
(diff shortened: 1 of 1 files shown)


thiagomacieira (commented):
No, this is indeed intended to print. They are called from an `assert()`:
https://github.com/opendcdiag/opendcdiag/blob/5f1c06ffe8395c36e45e83feb099f4ca2b1c28d1/framework/sandstone.h#L677-L679

So they will be removed in release builds. The idea is that we print what `memcmp_or_fail` would have printed in case of failure, so the CI can check that it's valid YAML and actually works. Ideally, we'd print only once per callback, so we don't a) flood the output; b) do print per different callbacks.

mcisowsk:
closing as this seems to be too bothering to 'fix' - a differentiation between CI and non-CI runs? If someone's bothered about it, just use release builds to run your tests. May revisit in future.

PR #1118, opened 2026-06-02
Add --thread-ratio flag to control thread-to-device ratio per slice

AthenasJimenez:
This PR introduces the `--thread-ratio` command-line option, which allows users to control how many threads are spawned per slice relative to the number of devices in that slice.

This becomes particularly useful when combined with reschedule, as it enables scenarios like running a subset of threads (e.g., 50%) across all available cores. This can leave part of the system temporarily idle during execution, allowing for more flexible scheduling and utilization patterns during tests.

### DIFF
--- a/bats/sanity-check/20-selftests.bats
+++ b/bats/sanity-check/20-selftests.bats
@@ -2132,3 +2132,69 @@ selftest_interrupt_common() {
     test_yaml_regexp "/tests/2/test" selftest_test_optional_include_optional
     test_yaml_regexp "/tests/3/test" selftest_test_optional_beta_include_optional
 }
+
+thread_ratio_expected_threads() {
+    local count=$1
+    local ratio=$2
+    local threads
+    if [[ "$ratio" == -* ]]; then
+        threads=$(( count + ratio ))
+    else
+        threads=$(awk "BEGIN { printf \"%d\", int($count * $ratio) }")
+    fi
+    (( threads = threads > 0 ? threads : 1 ))
+    echo "$threads"
+}
+
+check_thread_ratio_plans() {
+    local ratio=$1
+
+    # fullsocket
+    local fs_count=${yamldump[/test-plans/fullsocket/0/count]}
+    local expected=$(thread_ratio_expected_threads "$fs_count" "$ratio")
+    test_yaml_numeric "/test-plans/fullsocket/0/threads" "value == $expected"
+
+    # isolate_numa slices
+    local slice_count=${yamldump[/test-plans/isolate_numa@len]}
+    for ((i = 0; i < slice_count; i++)); do
+        local count=${yamldump[/test-plans/isolate_numa/$i/count]}
+        expected=$(thread_ratio_expected_threads "$count" "$ratio")
+        test_yaml_numeric "/test-plans/isolate_numa/$i/threads" "value == $expected"
+    done
+
+    # heuristic slices
+    slice_count=${yamldump[/test-plans/heuristic@len]}
+    for ((i = 0; i < slice_count; i++)); do
+        local count=${yamldump[/test-plans/heuristic/$i/count]}
+        expected=$(thread_ratio_expected_threads "$count" "$ratio")
+        test_yaml_numeric "/test-plans/heuristic/$i/threads" "value == $expected"
+    done
+
+    # Verify each test thread is pinned to a logical CPU within its slice's range
+    local thread_idx=$slice_count
+    for ((i = 0; i < slice_count; i++)); do
+        local start=${yamldump[/test-plans/heuristic/$i/starting_cpu]}
+        local count=${yamldump[/test-plans/heuristic/$i/count]}
+        local threads=${yamldump[/test-plans/heuristic/$i/threads]}
+
+        for ((t = 0; t < threads; t++)); do
+            test_yaml_numeric "/tests/0/threads/$thread_idx/id/logical" \
+                "value >= $start && value < $(( start + count ))"
+            thread_idx=$(( thread_idx + 1 ))
+        done
+    done
+}
+
+@test "thread_ratio delta YAML pass output" {
+    declare -A yamldump
+    sandstone_selftest -e selftest_pass -vvv --max-cores-per-slice=2 --thread-ratio=-1
+    test_yaml_regexp "/exit" pass
+    check_thread_ratio_plans -1
+}
+
+@test "thread_ratio 70% YAML pass output" {
+    declare -A yamldump
+    sandstone_selftest -e selftest_pass -vvv --max-cores-per-slice=2 --thread-ratio=70%
+    test_yaml_regexp "/exit" pass
+    check_thread_ratio_plans 0.70
+}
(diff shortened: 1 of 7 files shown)


bats/sanity-check/20-selftests.bats
```diff
@@ -2132,3 +2132,119 @@ selftest_interrupt_common() {
     test_yaml_regexp "/tests/2/test" selftest_test_optional_include_optional
     test_yaml_regexp "/tests/3/test" selftest_test_optional_beta_include_optional
 }
+
+@test "thread_ratio 25% YAML pass output" {
+    declare -A yamldump
+    sandstone_selftest -e selftest_pass -n8 --max-cores-per-slice=4 -vvv --thread-ratio=.25
```
thiagomacieira: GitHub CI only has 4 threads, so `-n8` can't pass. There also used to be some runners with just 2. Please verify the system first.
```
# + builddir/opendcdiag --on-crash=core --on-hang=kill --ignore-mce-errors -Y -o - -n4 --ignore-mce-errors --selftests --timeout=20s --retest-on-failure=0 -Y2 -e selftest_pass -n8 --max-cores-per-slice=4 -vvv --thread-ratio=.25
# builddir/opendcdiag: warning: value out of range for option '-n / --threads': 8 (minimum is 1, maximum 4)
# ---
#   command-line: 'opendcdiag --on-crash=core --on-hang=kill --ignore-mce-errors -Y -o - -n4 --ignore-mce-errors --selftests --timeout=20s --retest-on-failure=0 -Y2 -e selftest_pass -n8 --max-cores-per-slice=4 -vvv --thread-ratio=.25'
#   version: opendcdiag-<unknown version>
#   os: Linux 6.17.0-1015-azure
#   runtime: glibc 2.42
#   openssl: { version: OpenSSL 3.6.2 7 Apr 2026, providers: ['default 3.6.2','sandstone-rand'] }
#   timing: { duration: 1000.000, timeout: 20000.000 }
#   cpu-info:
#   - { logical: 0, package: 36, numa_node: 0, module: 36, core: 1, thread: 0 }   # 0
#   - { logical: 1, package: 36, numa_node: 0, module: 36, core: 2, thread: 0 }   # 1
#   - { logical: 2, package: 36, numa_node: 0, module: 36, core: 3, thread: 0 }   # 2
#   - { logical: 3, package: 36, numa_node: 0, module: 36, core: 4, thread: 0 }   # 3
#   test-plans:
#     fullsocket: [ { starting_cpu: 0, count: 4, threads: 1 } ]
#     isolate_numa: [ { starting_cpu: 0, count: 4, threads: 1 } ]
#     heuristic: [ { starting_cpu: 0, count: 4, threads: 1 } ]
...
# Numeric test failed:
# query:      /test-plans/fullsocket/0/threads
# value =     1
# expression: value == 2
```

You can query `nproc` to see how many threads are available and then adjust the values below. Be careful with rounding.

bats/sanity-check/20-selftests.bats
```diff
@@ -2132,3 +2132,119 @@ selftest_interrupt_common() {
...
+        -e selftest_pass -n8 --max-cores-per-slice=4 -vvv --thread-ratio=-1 --output-format=tap
+
+    # Consider 2 main threads, and 1 per slice
+    [[ "$output" = *"Thread 0 on CPU 0"* ]]
+    [[ "$output" = *"Thread 1 on CPU 1"* ]]
+    [[ "$output" = *"Thread 2 on CPU 2"* ]]
+    [[ "$output" = *"Thread 3 on CPU 4"* ]]
+    [[ "$output" = *"Thread 4 on CPU 5"* ]]
+    [[ "$output" = *"Thread 5 on CPU 6"* ]]
+}
```
thiagomacieira: These aren't necessary.

framework/topology.h
```diff
@@ -33,6 +33,14 @@ struct DeviceScheduler
     virtual void finish_reschedule() = 0;
 };
 
+enum class ThreadRatio {
+    full = 0,       // 100% default behavior
+    quarter,        // 25%
+    half,           // 50%
+    three_quarter,  // 75%
+    full_minus_1,   // 100% - 1
+};
```
thiagomacieira: I suggest two modes only:
```suggestion
enum class ThreadRatio : uint8_t {
    delta,        // default mode, with a delta of 0
    fraction,     // e.g., 75%
};
```

Or maybe even simpler:
```suggestion
using ThreadRatio = std::variant<int, float>;
```

framework/sandstone_opts.cpp
```diff
@@ -902,6 +905,27 @@ struct ProgramOptionsParser {
...
+                opts.thread_ratio = ThreadRatio::quarter;
+            } else if (str_value == "50%" || str_value == ".50" || str_value == "0.50"
+                       || str_value == ".5" || str_value == "0.5") {
+                opts.thread_ratio = ThreadRatio::half;
+            } else if (str_value == "75%" || str_value == ".75" || str_value == "0.75") {
+                opts.thread_ratio = ThreadRatio::three_quarter;
+            } else if (str_value == "100%" || str_value == "1.0" || str_value == "1") {
+                opts.thread_ratio = ThreadRatio::full;
+            } else if (str_value == "-1") {
+                opts.thread_ratio = ThreadRatio::full_minus_1;
```
thiagomacieira:
```suggestion
            bool ok = false;
            char *end = nullptr;
            if (str_value.starts_with('+') || str_value.starts_with('-')) {
                // delta mode
                // not using ParseIntArgument so we have a single error print out
                // explaining how to use the option
                long val = strtol(str_value.c_str(), &end, 10);
                // first approximation; we'll check again in thread_plan_init()
                if (*end != '\0') {
                     ok = false; // did not consume everything
                } else if (val < 0 && val > -app_cfg->device_count + 1) {
                    opts.thread_ratio = int(val);
                    ok = true;
                } else if (val > 0 && val < app_cfg->device_count - 1) {
                    // ratios over 1 not supported (yet)
                }
            } else {
                // parse as fraction
                float val = strtof(str_value.c_str(), &end);
                if (*end == '%') {
                    val /= 100;
                    ++end;
                }
                if (*end == '\0' && val > 0 && val < 1) {
                    opts.thread_ratio  = val;
                    ok = true;
                }
            }
            if (!ok) {
                fprintf(ERR_STREAM, ....);
                return EX_USAGE;
            }
```

maybe move to a separate function for cleanliness

thiagomacieira (changes requested):

framework/device/cpu/slicing.cpp
```diff
@@ -118,10 +118,26 @@ void slice_plan_init_for_device(SlicePlans::SlicesArray& plans, int max_cores_pe
...
-            slice.thread_range = { slice.device_range.starting_device, slice.device_range.device_count }; // 1:1 for now...
+        int thread_offset = 0;
+        for (auto &slice : plan) {
+            int n = slice.device_range.device_count;
+            int thread_count;
+            if (auto *delta = std::get_if<int>(&ratio_type)) {
+                thread_count = n + *delta;
+            } else {
+                thread_count = static_cast<int>(n * std::get<float>(ratio_type));
+            }
```
thiagomacieira, line 134: 👍

thiagomacieira (approved):

PR #1119, opened 2026-06-02
FW/logging: Fix thread header entries missing from stdout for TAP and Key-value

AthenasJimenez:
Thread entries with no messages were not printed to stdout. The `doprint` lambda gated both header and message output behind lowest_level <= verbosity, but threads with no messages have lowest_level = Max (127), so the condition always failed and the stdout output was silently skipped.

Decouple header printing from message printing for the stdout path: print the thread header when want_print or has_messages is true, and print messages only when has_messages is true. This ensures main thread entries (runtime, resource-usage) appear on stdout at -vvv verbosity even when the thread has no messages.

Also add a selftest for each output format that verifies at least one main thread entry exists and that all test threads are present.

A similar issue was previously fixed for the YAML format #1099. In this case, it affected not only the main threads but also test threads, due to changes introduced in another patch  #1100

### DIFF
--- a/bats/sanity-check/20-selftests.bats
+++ b/bats/sanity-check/20-selftests.bats
@@ -342,7 +342,7 @@ selftest_pass() {
     test_yaml_regexp "/tests/1/result" skip
 }
 
-@test "selftest_pass has main thread and loop-count at -vvv" {
+@test "selftest_pass YAML has main thread and loop-count at -vvv" {
     # Verify that at least 1 main thread exists with runtime
     # and resource-usage when running at -vvv verbosity.
     # Also confirm loop-count is printed for silent threads.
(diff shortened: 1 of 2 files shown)


bats/sanity-check/20-selftests.bats
```diff
@@ -359,6 +359,33 @@ selftest_pass() {
     done
 }
 
+@test "selftest_pass TAP has main thread and loop-count at -vvv" {
+    # Verify that at least 1 main thread exists with runtime
+    # and resource-usage when running at -vvv verbosity.
+    # Also confirm loop-count is printed for silent threads.
```
thiagomacieira, line 365: You haven't checked the `loop-count` output.

thiagomacieira (dismissed):

bats/sanity-check/20-selftests.bats
AthenasJimenez: Wops, fixed

thiagomacieira (approved):

PR #1120, opened 2026-06-03
FW/GPU: fix initializing slice plan with `device_count()`, not `thread_count()`

mcisowsk:


thiagomacieira (commented):
Looks ok, but @AthenasJimenez  to approve.

AthenasJimenez:
To be honest, these changes don’t fully make sense to me yet, but I’ll take a closer look tomorrow and follow up with what I find.

Also, just as a general note, it really helps to include reproduction steps when reporting an issue or proposing a fix. I’ve run into this problem before, so I was able to reproduce it on my side, but I’m not completely sure if it’s the exact same case.

AthenasJimenez:
Proposed fix on #1125

lukasz-tuz (approved):

mcisowsk:
> Proposed fix on #1125

not fixing the problem entirely, so this PR persists (changed it though)

PR #1123, opened 2026-06-04
FW: add `--ulog` option to log test state to mapped memory

thiagomacieira:
The option takes three `FILE=OFFSET` arguments, each pointing to a byte offset in a file. The framework mmap()s the page containing the offset and writes to the three resulting pointers at the start of each test:

-  [0] `test->shortid << 8 | retest_count`
-  [1] low 32 bits of the current random seed
-  [2] 0 (reserved for future expansion)

The feature is enabled by default and can be disabled at build time with the `no-ulog` framework option.

[ChangeLog][Framework] Added `--ulog FILE=OFFSET` option (must be specified exactly three times) to write the current test short ID, random seed, and retest count to user-specified locations in memory-mapped files.

This is very much a prototype for now.

### DIFF
--- a/bats/sanity-check/28-ulog.bats
+++ b/bats/sanity-check/28-ulog.bats
@@ -1,0 +1,215 @@
+#!/usr/bin/bats
+# -*- mode: sh -*-
+# Copyright 2026 Intel Corporation.
+# SPDX-License-Identifier: Apache-2.0
+load ../testenv
+load helpers
+
+function setup_file() {
+    setup_sandstone
+
+    # Probe whether --ulog is supported: if recognized, a bad arg produces
+    # "FILE=OFFSET"; if not recognized, it produces an unknown-option error.
+    run $SANDSTONE --ulog noequal --ulog noequal --ulog noequal
+    if [[ "$output" != *"FILE=OFFSET"* ]]; then
+        export ulog_skip_reason="--ulog not supported in this build"
+        return
+    fi
+
+    # MAP_SHARED_VALIDATE|MAP_SYNC requires a DAX-capable filesystem; tmpfs also
+    # works. Check by inspecting the filesystem type of the temp directory.
+    local probe_file fstype
+    probe_file=$(mktempfile ulog-probe-XXXXXX)
+    fstype=$(stat -f -c %T "$probe_file" 2>/dev/null)
+    rm -f "$probe_file"
+    if [[ -n "$fstype" && "$fstype" != "tmpfs" ]]; then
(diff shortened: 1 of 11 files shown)


bats/sanity-check/28-ulog.bats
```diff
@@ -0,0 +1,215 @@
...
+    # Probe whether --ulog is supported: if recognized, a bad arg produces
+    # "FILE=OFFSET"; if not recognized, it produces an unknown-option error.
+    run $SANDSTONE --ulog noequal --ulog noequal --ulog noequal
+    if [[ "$output" != *"FILE=OFFSET"* ]]; then
+        export ulog_skip_reason="--ulog not supported in this build"
+        return
+    fi
+
+    # MAP_SHARED_VALIDATE|MAP_SYNC requires a DAX-capable filesystem; tmpfs also
+    # works. Check by inspecting the filesystem type of the temp directory.
```
mrkz, line 20: shouldn't this be documented somewhere else additionally, instead of only on the sanity tests? 

Given the description here, I'd assume this feature its not as effective otherwise, hence my thought of elevating the documentation of this requirement.

AthenasJimenez (approved):
On selftests I think it only misses to check when no value after `=` . ie. `--ulog $f=`. Otherwise, it looks good to me.

bats/sanity-check/28-ulog.bats
thiagomacieira: Good point, but I haven't documented *anywhere* yet. Once we do, you're right this needs to be included.

framework/random.cpp
```diff
@@ -554,6 +554,11 @@ void random_advance_seed()
     rand();
 }
 
+uint32_t random_seed_low32()
```
mrkz, line 557: I see the commit mentions that some seed types would not provide the required details in this ulog stamp, I only have a couple of questions;

- I would assume that the covered/supported seed is the default used in the framework, is that an accurate assumption?
- What would be the observed behavior in the case of the not-happy-path seed used for this feature? if I'm not mistaken, it would only provide bad data for the seed logged.
- is there any future plan to address that so that we cover for other seeds later?
- In a negative case for the previous question, I think it would be better to either document (log,warning,even on the documentation) or restrict the usage whatsoever for misleading data (that is, assuming that my reasoning is accurate, otherwise I'd be happy to read the corrections).

thiagomacieira: Ideally we'd print a warning that the ulog does not include all the information that would be needed to recreate the execution. I didn't implement that because this is experimental.

PR #1127, opened 2026-06-07
FW: Use relative thread and device indices in child for_each_test_thread

AthenasJimenez:
The previous double-offset fix introduced separate data_idx and device_idx variables for array access, but still passed the absolute thread index to the lambda. This is inconsistent in child processes where device_info has been rebased by restrict_topology() — lambdas that index into device_info need relative indices to match.

Simplify by computing thread and device as relative (i) in child processes and absolute (starting_thread + i, starting_device + i) in the parent, then using these unified values for both data access and lambda arguments.

### DIFF
--- a/framework/sandstone_p.h
+++ b/framework/sandstone_p.h
@@ -471,23 +471,22 @@ template <typename Lambda> static void for_each_main_thread(Lambda &&l, int max_
 
 template <typename Lambda> static void for_each_test_thread(Lambda &&l, int max_slices = INT_MAX)
 {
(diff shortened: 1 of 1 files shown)


framework/sandstone_p.h
thiagomacieira:
```suggestion
    bool is_main_process = sApp->is_main_process();
    int slices = is_main_process ? std::min(sApp->shmem->main_thread_count, max_slices) : 1;
```
then use this bool before.`is_main_process()` calls `getpid()` and you're calling it in a loop.

thiagomacieira (changes requested):

thiagomacieira (approved):

PR #1139, opened 2026-06-23
Tests/GPU: add `event_monitor` special test for tracking HW events

mcisowsk:


scripts/generate-short-ids.pl
```diff
@@ -60,7 +60,7 @@ ($@)
 my @test_subdirs = find_dirs($me, @ARGV);
 
 # Find all possible tests in the source code
-my @alltests = qw(mce_check);
+my @alltests = qw(mce_check event_monitor);
```
mcisowsk, line 63: right now short ids are global and contain all tests for both devices. Would we want to change that in future?

tests/gpu/event_monitor/event_monitor.cpp
```diff
@@ -0,0 +1,169 @@
...
+    static constexpr zes_event_type_flags_t event_types =
+        ZES_EVENT_TYPE_FLAG_DEVICE_DETACH | ZES_EVENT_TYPE_FLAG_FREQ_THROTTLED |
+        ZES_EVENT_TYPE_FLAG_TEMP_CRITICAL | ZES_EVENT_TYPE_FLAG_MEM_HEALTH |
+        ZES_EVENT_TYPE_FLAG_FABRIC_PORT_HEALTH | ZES_EVENT_TYPE_FLAG_PCI_LINK_HEALTH |
+        ZES_EVENT_TYPE_FLAG_RAS_CORRECTABLE_ERRORS | ZES_EVENT_TYPE_FLAG_RAS_UNCORRECTABLE_ERRORS |
+        ZES_EVENT_TYPE_FLAG_DEVICE_RESET_REQUIRED;// | ZES_EVENT_TYPE_FLAG_SURVIVABILITY_MODE_DETECTED; // TODO: ZE_RESULT_ERROR_INVALID_ENUMERATION 0xffff < events
+
+    auto td = new test_data;
+
+    bool skip_logged = false;
```
mcisowsk: I feel like the fact that we have to use such flag is not a good sign about our API

mcisowsk: - seems like done in #1142

tests/gpu/event_monitor/event_monitor.cpp
```diff
@@ -0,0 +1,170 @@
...
+    static constexpr zes_event_type_flags_t event_types =
+        ZES_EVENT_TYPE_FLAG_DEVICE_DETACH | ZES_EVENT_TYPE_FLAG_FREQ_THROTTLED |
+        ZES_EVENT_TYPE_FLAG_TEMP_CRITICAL | ZES_EVENT_TYPE_FLAG_MEM_HEALTH |
+        ZES_EVENT_TYPE_FLAG_FABRIC_PORT_HEALTH | ZES_EVENT_TYPE_FLAG_PCI_LINK_HEALTH |
+        ZES_EVENT_TYPE_FLAG_RAS_CORRECTABLE_ERRORS | ZES_EVENT_TYPE_FLAG_RAS_UNCORRECTABLE_ERRORS |
+        ZES_EVENT_TYPE_FLAG_DEVICE_RESET_REQUIRED;// | ZES_EVENT_TYPE_FLAG_SURVIVABILITY_MODE_DETECTED; // TODO: ZE_RESULT_ERROR_INVALID_ENUMERATION 0xffff < events
+
+    auto td = new test_data;
+
+    // TODO: prepare_test_for_device has not yet been called! We must set it manually before calling for_each*
```
mcisowsk: - see #1144 which solves that

framework/sandstone.h
```diff
@@ -819,6 +819,21 @@ memcmp_or_fail(const T *actual, const T *expected, size_t count, const char *fmt
...
+#define LOG_SKIP_OR_PRINT(cat, fmt, ...) \
+    do { \
+        if (!sApp->shmem) { \
+            fprintf(stderr, fmt "\n", ##__VA_ARGS__); \
+        } else if (logging_in_test) { \
+            log_skip(cat, fmt, ##__VA_ARGS__); \
+        } else { \
+            logging_printf(LOG_LEVEL_QUIET, fmt "\n", ##__VA_ARGS__); \
+        } \
+    } while (0)
```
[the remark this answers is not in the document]
mcisowsk: old code

tests/gpu/event_monitor/event_monitor.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+    int ret = for_each_zes_device_within_topo([&](zes_device_handle_t device_handle, ze_driver_handle_t driver, const MultiSliceGpu&) {
+        if (auto ret = zesDeviceEventRegister(device_handle, event_types); ret != ZE_RESULT_SUCCESS) {
+            log_skip(RuntimeSkipCategory, "Unable to create event monitor: %s", to_string(ret));
+            return EXIT_SKIP;
+        }
+        td->thread_data.devices.emplace_back(device_handle);
+        td->thread_data.driver = driver;
+        return EXIT_SUCCESS;
+    });
```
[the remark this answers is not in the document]
mcisowsk: but that's how it behaves for subdevices

tests/gpu/event_monitor/event_monitor.cpp
```diff
@@ -0,0 +1,167 @@
...
+    if (ret != EXIT_SUCCESS) {
+        return ret;
+    }
+
+    td->thread_data.n_events.resize(td->thread_data.devices.size());
+    td->thread_data.events.resize(td->thread_data.devices.size());
+
```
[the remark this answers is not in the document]
mcisowsk: zero-devices-case should be handled earlier by the framework.

tests/gpu/event_monitor/event_monitor.cpp
```diff
@@ -0,0 +1,172 @@
...
+        std::vector<uint64_t> n_events; // cumulative count of how many times API reported events on device
+        std::vector<zes_event_type_flags_t> events; // bitmask of all events observed (ORed), per device
+        std::atomic<bool> thread_should_run = true;
+    } thread_data; // data used in thread
+    std::thread thread;
+};
+
+int listener_preinit(struct test* test)
+{
+    // events that we want to track
```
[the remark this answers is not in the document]
mcisowsk: well, you're right

scripts/generate-short-ids.pl
thiagomacieira: So long as there's no shortid collision, it won't be a problem. If there is, we'll need to rethink.

Right now, the manual declaration isn't including the `_testid_xxxxxx` variable, so we won't detect the collision. See below.

tests/gpu/event_monitor/event_monitor.cpp
```diff
@@ -0,0 +1,178 @@
...
+// event_monitor is a special test.
+struct test event_monitor_test = {
+    .shortid = CONCAT(0x, TEST_ID_event_monitor),
+#if SANDSTONE_NO_TEST_NAMES
+    .id = SANDSTONE_STRINGIFY(TEST_ID_event_monitor),
+    .description = nullptr,
+#else
+    .id = "event_monitor",
+    .description = "Monitor GPU device events for hardware issues",
+#endif // SANDSTONE_NO_TEST_NAMES
```
thiagomacieira: Not for this commit, but we should add a macro to automate this. I'm thinking `DECLARE_MANUAL_TEST` which is exactly `DECLARE_TEST` minus the section.

To do that, we'd have both macros use a common one that takes the extra attributes or empty.

tests/gpu/event_monitor/event_monitor.cpp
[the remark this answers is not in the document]
thiagomacieira: But it's Linux, where we know it works, in spite of what POSIX says isn't required to be supported. Besides, you've set the `test_in_parent` flag, meaning that we *won't* `fork()` for this test. The `run()` function (which is empty) will be executed in the parent process too.

But if you don't want to support `-fyes` for GPUs, that's fine.

tests/gpu/event_monitor/event_monitor.cpp
```diff
@@ -0,0 +1,178 @@
...
+    .test_preinit = listener_preinit,
+    .test_run = [](struct test*, int) { return EXIT_SUCCESS; },
+    .test_postcleanup = listener_postcleanup,
+    .desired_duration = -1,
+    .fracture_loop_count = -1,
+    .quality_level = TEST_QUALITY_PROD,
+    .flags = test_in_parent,
```
thiagomacieira: Consider `test_schedule_sequential`, so the framework doesn't launch threads for each device.

tests/gpu/event_monitor/event_monitor.cpp
```diff
@@ -0,0 +1,178 @@
...
+            if (cur_n_events) {
+                for (size_t i = 0; i < td->thread_data.devices.size(); i++) {
+                    if (cur_events[i]) {
+                        td->thread_data.events[i] |= cur_events[i];
+                        td->thread_data.n_events[i]++;
+                    }
+                }
+            }
+
+            std::this_thread::sleep_for(150ms);
```
thiagomacieira: Why 150 ms? Would it help to wake this thread up after every test (a heartbeat)? Why do we need a thread at all, instead of getting the details at the end of the execution?

tests/gpu/event_monitor/event_monitor.cpp
```diff
@@ -0,0 +1,178 @@
...
+        return EXIT_SUCCESS;
+    });
+    if (ret != EXIT_SUCCESS) {
+        return ret;
+    }
+
+    td->thread_data.n_events.resize(td->thread_data.devices.size());
+    td->thread_data.events.resize(td->thread_data.devices.size());
+
+    auto listener_runner = [td = td.get()]() {
```
thiagomacieira: Stylistic: this is a pretty long function. I'd prefer a real function, not a lambda. You can pass the `test_data` as a parameter.

tests/gpu/event_monitor/event_monitor.cpp
```diff
@@ -0,0 +1,178 @@
...
+                    }
+                }
+            }
+
+            std::this_thread::sleep_for(150ms);
+        }
+    };
+
+    try {
+        td->thread = std::thread(listener_runner);
```
thiagomacieira: Consider using a `std::jthread`, which has [`request_stop()`](https://en.cppreference.com/cpp/thread/jthread/request_stop).

It's a major overkill to replace an atomic boolean with a `stop_token` though...

tests/gpu/event_monitor/event_monitor.cpp
```diff
@@ -0,0 +1,178 @@
...
+        using namespace std::chrono_literals;
+
+        assert(td != nullptr);
+
+        static constexpr uint32_t timeout = 10; // ms
+
+        uint32_t cur_n_events; // the actual number of devices that generated event(s), per call.
+        std::vector<zes_event_type_flags_t> cur_events(td->thread_data.devices.size());
+
+        while (td->thread_data.thread_should_run.load()) {
```
thiagomacieira:
```suggestion
        while (td->thread_data.thread_should_run.load(std::memory_order::relaxed)) {
```

tests/gpu/event_monitor/event_monitor.cpp
```diff
@@ -0,0 +1,178 @@
...
+
+    return EXIT_SUCCESS;
+}
+
+int listener_postcleanup(struct test* test)
+{
+    // join the thread, log results
+    auto td = static_cast<test_data*>(test->data);
+
+    td->thread_data.thread_should_run.store(false);
```
thiagomacieira:
```suggestion
    td->thread_data.thread_should_run.store(false, std::memory_order::relaxed);
```

tests/gpu/event_monitor/event_monitor.cpp
```diff
@@ -0,0 +1,178 @@
...
+    td->thread.join();
+
+    bool any_event =
+        std::any_of(td->thread_data.n_events.begin(), td->thread_data.n_events.end(), [](auto n) { return n > 0; });
+
+    if (any_event) {
+        for (size_t i = 0; i < td->thread_data.devices.size(); i++) {
+            if (td->thread_data.n_events[i] > 0) {
+                log_info("GPU %zu: %" PRIu64 " event notifications, mask: %#x",
+                        i, td->thread_data.n_events[i], td->thread_data.events[i]);
```
thiagomacieira: Is there a 1:1 matching of entries here with the framework thread_num?

If so
```suggestion
                log_message(int(i), SANDSTONE_LOG_INFO "%" PRIu64 " event notifications, mask: %#x",
                        td->thread_data.n_events[i], td->thread_data.events[i]);
```

thiagomacieira (changes requested):

tests/gpu/event_monitor/event_monitor.cpp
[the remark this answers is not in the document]
mcisowsk: it will fork for other tests, while the listener thread still running. At least that's the idea of this test - that listener thread is alive throughout the entire program run

tests/gpu/event_monitor/event_monitor.cpp
mcisowsk: that was the idea, so that we can potentially detect which test caused a certain event to occur (for that we would need to track the timestamp as well, which we currently do not do, though). 150ms is at the top of my head, arbitrary. Could be also the value of `-t`

tests/gpu/event_monitor/event_monitor.cpp
mcisowsk: it should

tests/gpu/event_monitor/event_monitor.cpp
mcisowsk: good to know, then maybe next time :)

tests/gpu/event_monitor/event_monitor.cpp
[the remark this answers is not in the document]
thiagomacieira: Yes, but in the parent process. That's irrelevant for the tests.

thiagomacieira (approved):

PR #1140, opened 2026-06-23
simd.conf: add support for detecting ACEv1

thiagomacieira:
From https://x86ecosystem.org/wp-content/uploads/2026/06/ACE_v1_Specification_public_1_15.pdf

That document does not describe whether Linux will require a different arch_prl(ARCH_REQ_XCOMP_PERM) to enable the BSR state. This patch assumes the kernel will enable it when we request XTILEDATA and XTILECFG. If that ends up not being the case, we'll need to update this.


busykai (approved):
Looks OK as a stub. There's much more to uncover when it's going to implemented, I guess...

PR #1141, opened 2026-06-23
FW/Random: add --random-control option

thiagomacieira:
Add a new `--random-control=<CONTROLS>` command-line option that modifies RNG behavior. Controls are specified as a comma-separated list, allowing multiple to be combined and more to be added in the future without interface changes.

Two controls are introduced:

**`no-thread-mixin`** — By default, each thread's RNG is seeded by XOR-ing a per-device topology value into the global seed, giving every thread a unique random sequence. With `no-thread-mixin`, this mixin is skipped and all threads start from the same RNG state.

**`fixed-seed`** — By default, the global RNG seed is advanced after each test, so consecutive tests produce different random sequences. With `fixed-seed`, the seed is not advanced, so every test starts from the same seed. Combined with `-s ENGINE:SEED`, this makes runs fully reproducible regardless of how many tests run.

The two controls are independent and can be combined:
```
--random-control=no-thread-mixin,fixed-seed
```

BATS selftests are added for each control individually and for the combined case.

### DIFF
--- a/framework/sandstone_opts.cpp
+++ b/framework/sandstone_opts.cpp
@@ -300,6 +300,8 @@ Common command-line options are:
      Comma-separated list of controls that modify RNG behavior. Available controls:
        uniform-thread-seed  Uniformly initialize the random generator in each thread
                             to the same state.
+       fixed-seed           Do not advance the global seed between tests. Every test
+                            will start from the same seed.
  -v, -q, --verbose, --quiet
      Set logging output verbosity level. Default is quiet.
  --version
@@ -963,6 +965,14 @@ struct ProgramOptionsParser {
                 if (token == "uniform-thread-seed") {
                     opts.shmem_cfg.random_control_flags |=
                         uint32_t(TestConfig::RandomControl::uniform_thread_seed);
+                } else if (token == "fixed-seed") {
+                    opts.shmem_cfg.random_control_flags |=
+                        uint32_t(TestConfig::RandomControl::fixed_seed);
+                } else if (token == "fixed-uniform-seed" || token == "uniform-fixed-seed") {
+                    // shorthand forms
+                    opts.shmem_cfg.random_control_flags |=
+                            uint32_t(TestConfig::RandomControl::fixed_seed) |
+                            uint32_t(TestConfig::RandomControl::uniform_thread_seed);
                 } else {
                     fprintf(ERR_STREAM, "%s: unknown --random-control value: '%.*s'\n",
                             argv[0], int(token.size()), token.data());
--- a/framework/sandstone_opts.cpp
+++ b/framework/sandstone_opts.cpp
@@ -68,6 +68,7 @@ enum {
     output_format_option,
     quality_option,
     quick_run_option,
+    random_control_option,
     raw_list_test_ids,
     raw_list_tests,
     raw_list_group_members,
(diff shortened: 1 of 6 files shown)


mcisowsk (dismissed):
lgtm, if there is a need for such a feature

framework/sandstone_opts.cpp
```diff
@@ -948,6 +957,28 @@ struct ProgramOptionsParser {
             }
         }
 
+        if (auto value = string_opt_for(random_control_option)) {
+            std::string_view sv(value);
+            while (!sv.empty()) {
```
mcisowsk, line 962: you could also use `boost::split` which is already used in this file to split comma separated stuff

brad-kelly:
One more possible control that might be of use is to have it put the seed back to the start with every test, but allow the normal advance that happens on fragments.  this way people can replay a set of seeds that came from a previous run that fails on a later seed.

**--random-control=rewind**

```
Test starts
Fragment 1 : Seed A
Fragment 2 : Seed B
Fragment 3 : Seed C
Test Ends

Test starts again
Fragment 1 : Seed A
Fragment 2 : Seed B
Fragment 3 : Seed C
Test Ends
```

brad-kelly:
I don't think this is working properly:

```bash
% ./builddir/opendcdiag -e eigen_svd -vvv --random-control=fixed-seed -s LCG:839871525  -o log.yaml
& grep seed *.yaml
command-line: 'opendcdiag -e eigen_svd -vvv --random-control=fixed-seed -s LCG:839871525 -o log.yaml'
  state: { seed: 'LCG:839871525', iteration: 0, retry: false }
  state: { seed: 'LCG:1242095209', iteration: 1, retry: false }
  state: { seed: 'LCG:1581893046', iteration: 2, retry: false }
  state: { seed: 'LCG:1483187087', iteration: 3, retry: false }
  state: { seed: 'LCG:2114052891', iteration: 4, retry: false }
  state: { seed: 'LCG:1171679668', iteration: 5, retry: false }
  state: { seed: 'LCG:2019926636', iteration: 6, retry: false }
  state: { seed: 'LCG:1678621615', iteration: 7, retry: false }
  state: { seed: 'LCG:2038492708', iteration: 8, retry: false }
  state: { seed: 'LCG:233318681', iteration: 9, retry: false }
  state: { seed: 'LCG:1121805683', iteration: 10, retry: false }
  state: { seed: 'LCG:1881964988', iteration: 11, retry: false }
  state: { seed: 'LCG:1881964988', iteration: 0, retry: false }


```

brad-kelly (changes requested):
See my previous comment that fixed-seed is not holding the seed fixed

thiagomacieira:
> See my previous comment that fixed-seed is not holding the seed fixed

Fixed.

brad-kelly:
I do not think `no-thread-mixin` is a good name for this option.  `fixed-seed` is great, but can we name this more from the user's perspective?

After talking it over with a copilot, it made a few good suggestions:
* `shared-seed`
* `uniform-seed`
* `synch-seeds`

Personally, I like `uniform-seed`.  It seems to imply that, without this all the threads are different - which is something people often have trouble wrapping their heads around wrt to the randomness.

I also kind of wonder if we need the comma or the repeated "seed" if we do something like:
* `--random-control=fixed-seed`
* `--random-control=uniform-seed`
* `--random-control=fixed-uniform-seed`

You can split on the dash and ignore the "seed" part which means the user can be omit that part if they want.

thiagomacieira:
> I do not think `no-thread-mixin` is a good name for this option. `fixed-seed` is great, but can we name this more from the user's perspective?
> 
> After talking it over with a copilot, it made a few good suggestions:
> 
> * `shared-seed`
> * `uniform-seed`
> * `synch-seeds`
> 
> Personally, I like `uniform-seed`. It seems to imply that, without this all the threads are different - which is something people often have trouble wrapping their heads around wrt to the randomness.

Me too. Or maybe `thread-uniform-seed`, to indicate it's uniform for all threads.

> I also kind of wonder if we need the comma or the repeated "seed" if we do something like:
> 
> * `--random-control=fixed-seed`
> * `--random-control=uniform-seed`
> * `--random-control=fixed-uniform-seed`
> 
> You can split on the dash and ignore the "seed" part which means the user can be omit that part if they want.

I can add a mode that sets both, yes.

PR #1144, opened 2026-06-25
FW: port `logging_in_test` to common part of the framework

mcisowsk:
This commit also fixes `finish_test_for_device` not being called in case of a fail.

### DIFF
--- a/framework/device/gpu/ze_check.h
+++ b/framework/device/gpu/ze_check.h
@@ -11,33 +11,13 @@
 
 #include <level_zero/ze_api.h>
 
-#include <stdio.h>
-
-extern bool logging_in_test;
-
-/// Log a runtime skip reason during test execution; outside tests, print a quiet message instead.
-#define LOG_RUNTIME_SKIP_OR_PRINT(fmt, ...) \
-    do { \
-        if (!sApp->shmem) { \
-            fprintf(stderr, fmt "\n", ##__VA_ARGS__); \
-        } else if (logging_in_test) { \
-            log_skip(RuntimeSkipCategory, fmt, ##__VA_ARGS__); \
-        } else { \
-            logging_printf(LOG_LEVEL_QUIET, fmt "\n", ##__VA_ARGS__); \
-        } \
-    } while (0)
-
 /// Check return value of an L0 API call.
 /// In test context it returns EXIT_SKIP; outside tests it returns EXIT_FAILURE.
 #define ZE_CHECK(...) \
     do { \
         auto result = (__VA_ARGS__); \
         if (result != ZE_RESULT_SUCCESS) { \
-            LOG_RUNTIME_SKIP_OR_PRINT("L0 API call failed with status %s", to_string(result)); \
-            if (logging_in_test) { \
-                return EXIT_SKIP; \
-            } \
-            return EXIT_FAILURE; \
+            return log_skip_or_print(RuntimeSkipCategory, "L0 API call failed with status %s", to_string(result)); \
         } \
     } while (0)
 
--- a/framework/device/gpu/ze_enumeration.cpp
+++ b/framework/device/gpu/ze_enumeration.cpp
@@ -31,8 +31,7 @@ int for_each_ze_device(const std::function<int(ze_device_handle_t, ze_driver_han
     // Always pick the last ("newest") one.
     auto ze_driver = drivers.back();
     if (!ze_driver) {
-        LOG_RUNTIME_SKIP_OR_PRINT("Could not initialize driver handle");
-        return logging_in_test ? EXIT_SKIP : EXIT_FAILURE;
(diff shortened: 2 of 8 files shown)


framework/sandstone_run.cpp
```diff
@@ -1199,7 +1214,6 @@ static TestResult run_one_test_inner(struct test *test, bool init_in_aux_thread
             }
         }
 
-        finish_test_for_device(test);
         sApp->test_tests_finish(test);
```
mcisowsk, line 1217: now `test_tests_finish` goes before `finish_test_for_device` - is this ok? In fact it should be the opposite order from the prepare sequence and now it seems like the right order

framework/device/gpu/ze_check.h
```diff
@@ -7,33 +7,18 @@
 #define INC_ZE_CHECK_H
 
 #include "sandstone_p.h"
+#include "logging.h"
```
thiagomacieira: Please don't include `logging.h` in this header if it is meant to be included by tests.

Instead, declare the function you want in sandstone.h

framework/device/gpu/ze_enumeration.cpp
```diff
@@ -31,7 +31,7 @@ int for_each_ze_device(const std::function<int(ze_device_handle_t, ze_driver_han
     // Always pick the last ("newest") one.
     auto ze_driver = drivers.back();
     if (!ze_driver) {
-        LOG_RUNTIME_SKIP_OR_PRINT("Could not initialize driver handle");
+        LOG_SKIP_OR_PRINT(RuntimeSkipCategory, "Could not initialize driver handle");
         return logging_in_test ? EXIT_SKIP : EXIT_FAILURE;
```
thiagomacieira: And make the function return either of the two values automatically, so you could just write:
```suggestion
        return log_skip_or_print(RuntimeSkipCategory, "Could not initialize driver handle");
```

framework/logging.cpp
```diff
@@ -116,6 +116,8 @@ enum class AbstractLogger::LogTypes : uint8_t
 };
 using LogTypes = AbstractLogger::LogTypes;
 
+bool logging_in_test = false;
```
thiagomacieira: One way of doing this test, at least on Linux, is `gettid() != getpid()`, because the two come from the same namespace. That won't work on Windows, of course, so I don't think it's worth saving a byte.

I'd also rename it to `test_has_started`.

thiagomacieira (changes requested):

framework/logging.cpp
mcisowsk: this variable is set back to false once the test has finished, so I don't think this name is better. But we can rename it so it does not contain the "logging" part

framework/logging.cpp
```diff
@@ -2373,3 +2373,23 @@ TestResult logging_print_results(std::span<const ChildExitStatus> status, const
...
+    std::string msg = create_filtered_message_string(fmt, va);
+    va_end(va);
+
+    if (!sApp->shmem) {
+        fprintf(stderr, "%s\n", msg.c_str());
+    } else if (logging_in_test) {
+        log_skip(cat, "%s", msg.c_str());
+    } else {
+        logging_printf(LOG_LEVEL_QUIET, "%s\n", msg.c_str());
+    }
```
thiagomacieira, line 2392: Need some thinking about no-logging mode. It hasn't been an issue for the GPU yet.

framework/sandstone_p.h
```diff
@@ -644,6 +644,8 @@ void logging_init(const struct test *test);
 void logging_init_child_preexec();
 void logging_finish();
 TestResult logging_print_results(std::span<const ChildExitStatus> status, const struct test *test);
+extern bool logging_in_test;
```
thiagomacieira: Nitpick: move to the top of this block.

framework/sandstone_run.cpp
```diff
@@ -1123,6 +1123,20 @@ static void wait_for_children(ChildrenList &children, const struct test *test)
     }
 }
 
+extern bool logging_in_test;
```
thiagomacieira, line 1126: I think we have a variable for this:
https://github.com/opendcdiag/opendcdiag/blob/b9da2f3153436ed266589afcc8403e10972d65c7/framework/sandstone_run.cpp#L70

It is private to sandstone_run.cpp, so logging.cpp can't see it, but we can fix that. We can look into that later.

We should also fix the preinit to set that variable.

framework/sandstone_run.cpp
thiagomacieira: Looks ok.

thiagomacieira (dismissed):

framework/sandstone_run.cpp
mcisowsk: and in what context would you like to use this variable?

dawid-sabat (approved):

PR #1148, opened 2026-06-29
sandstone_context_dump: skip checking for OSXSAVE bit if APX or AVX

thiagomacieira:
If we have been running AVX or APX instructions, whatever allowed the application to run must have already done this check.

### DIFF
--- a/framework/sandstone_context_dump.cpp
+++ b/framework/sandstone_context_dump.cpp
@@ -521,8 +521,14 @@ void dump_xsave(std::string &f, const void *xsave_area, size_t xsave_size, int x
         // sanity check it
         uint64_t xgetbv0 = XSave::X87 | XSave::SseState;
 
+#if defined(__AVX__) || defined(__APX_F__)
+        // OSXSAVE is definitely active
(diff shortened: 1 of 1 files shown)


mcisowsk (approved):

framework/sandstone_context_dump.cpp
```diff
@@ -521,8 +521,14 @@ void dump_xsave(std::string &f, const void *xsave_area, size_t xsave_size, int x
         // sanity check it
         uint64_t xgetbv0 = XSave::X87 | XSave::SseState;
 
+#if defined(__AVX__) || defined(__APX_F__)
```
mcisowsk, line 524: was thinking about maybe something like:
```c++
#if !defined(__AVX__) && !defined(__APX_F__)
        uint32_t eax, ebx, ecx, edx;
        __cpuid(1, eax, ebx, ecx, edx);
        if (bit_OSXSAVE & ecx)
#endif
               xgetbv0 = do_xgetbv();
```
but it seems to be uglier

PR #1150, opened 2026-06-30
FW: introduce new `idxd` device_type

mcisowsk:
With empty implementations and an only assumption that device in this context is the IDXD work queue.

### DIFF
--- a/framework/device/idxd/idxd_device.h
+++ b/framework/device/idxd/idxd_device.h
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2026 Intel Corporation.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef INC_IDXD_DEVICE_H
+#define INC_IDXD_DEVICE_H
+
+#include <stdint.h>
+
+// Work Queue is an atomic piece of a programmable accelerator HW,
+// therefore device_info should store WQs, not DSA/IAX devices.
+// Inside it we should only store immutable fields.
+struct wq_info_t
+{
+    /// Logical OS processor number.
+    /// On Unix systems, this is a sequential ID; on Windows, it encodes
+    /// 64 * ProcessorGroup + ProcessorNumber
+    int cpu_number;
+
+    /// Package ID in the system.
+    int16_t package_id;
+
+    /// ...
+    /// ...
+    /// ...
+
+#ifdef __cplusplus
+    int wq() const;        ///! Internal WQ number
+#endif
+};
+
+// Alias for use in common framework code
+typedef struct wq_info_t device_info_t;
+
+extern struct wq_info_t *device_info;
+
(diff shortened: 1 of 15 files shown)


framework/device/idxd/idxd_device.h
```diff
@@ -0,0 +1,50 @@
...
+
+// Work Queue is an atomic piece of a programmable accelerator HW,
+// therefore device_info should store WQs, not DSA/IAX devices.
+// Inside it we should only store immutable fields.
+struct wq_info_t
+{
+    /// Logical OS processor number.
+    /// On Unix systems, this is a sequential ID; on Windows, it encodes
+    /// 64 * ProcessorGroup + ProcessorNumber
+    int cpu_number;
```
mcisowsk, line 19: can we assume that `cpu_number` won't change for a particular WQ? If not, can we at least forbid reconfiguration in-between-tests to change that?

thiagomacieira: Strictly speaking, the device doesn't have a cpu number. Any CPU can talk to the device.

But talking to people doing debugging, sometimes they want to force a pairing. So I recommend we just use an OS CPU number here, or -1 indicating "any". How do you do it on GPU when the user doesn't specify a paired CPU?

framework/device/idxd/idxd_device.h
```diff
@@ -0,0 +1,50 @@
...
+
+#ifdef __cplusplus
+inline int wq_info_t::wq() const
+{
+    return this - ::device_info;
+}
+#endif
+
+// Not used at the moment
+typedef unsigned __int128 device_features_t;
```
thiagomacieira, line 46: Why `__int128`? Plain `unsigned` should suffice here, until we need more.

Should we use this to separate DSA from IAX?

framework/device/idxd/logging_idxd.cpp
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+#include "logging.h"
+
+#if !SANDSTONE_NO_LOGGING
+std::string AbstractLogger::thread_id_header_for_device(int thread, LogLevelVerbosity verbosity)
+{
+    std::string line;
+    return line;
+}
```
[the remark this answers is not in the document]
thiagomacieira: AI is right. Is this coming in a later commit?

PS: we *can*  use `std::format()` for the IDXD build.

thiagomacieira (approved):

framework/device/idxd/logging_idxd.cpp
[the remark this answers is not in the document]
mcisowsk: it is

framework/device/idxd/idxd_device.h
mcisowsk: right now it's just copied from other devices, tbh

PR #1151, opened 2026-06-30
FW/IDXD: fill `device_info` and `Topology`

mcisowsk:
The idea is to store immutable features of a work queue in `wq_info_t`. `Topology` is assumed to be changing in-between tests due to reconfiguration of the system. Each test is assumed to be able to adjust the current config of the system (and therefore the `Topology`) to its needs.

### DIFF
--- a/framework/device/idxd/idxd_device.h
+++ b/framework/device/idxd/idxd_device.h
@@ -6,7 +6,18 @@
 #ifndef INC_IDXD_DEVICE_H
 #define INC_IDXD_DEVICE_H
 
+#include <accel-config/libaccel_config.h>
+
 #include <stdint.h>
+
+
+struct bdf_t
+{
+    uint16_t domain;
+    uint8_t bus;
+    uint8_t device : 5;
+    uint8_t function : 3;
+};
 
 // Work Queue is an atomic piece of a programmable accelerator HW,
 // therefore device_info should store WQs, not DSA/IAX devices.
@@ -21,9 +32,20 @@ struct wq_info_t
     /// Package ID in the system.
     int16_t package_id;
 
-    /// ...
-    /// ...
-    /// ...
+    /// PCI location.
+    struct bdf_t bdf;
+
+    /// WQ unique index within a device.
+    int wq_id;
+
+    /// Device that this WQ belongs to.
+    int device_id;
+
+    /// Device type that this WQ belongs to (DSA/IAX).
+    enum accfg_device_type dev_type;
+
+    /// Device version (V1/V2/V3).
+    enum accfg_device_version dev_version;
 
 #ifdef __cplusplus
     int wq() const;        ///! Internal WQ number
--- a/framework/device/idxd/topology_idxd.hpp
+++ b/framework/device/idxd/topology_idxd.hpp
@@ -6,29 +6,138 @@
 #ifndef INC_TOPOLOGY_IDXD_HPP
 #define INC_TOPOLOGY_IDXD_HPP
 
+#include "idxd_device.h"
+
 #include <accel-config/libaccel_config.h>
 
+#include <cstdint>
+#include <optional>
+#include <utility>
 #include <vector>
 
-/// Immutable and unique id of a queue: device_id and wq_id, as in qw<device_id>.<wq_id>.
-struct WorkQueueId
+/// RAII wrapper around accfg_ctx. The lifetime of the underlying context is the same as this wrapper's.
+/// accfg_ref is not used at this point.
+struct AccfgCtx
 {
-    accfg_device_type device_type;
-    int device_id;
-    int wq_id;
+    accfg_ctx* ctx = nullptr;
+
+    AccfgCtx() = default;
+    AccfgCtx(AccfgCtx&& other) noexcept
+        : ctx(std::exchange(other.ctx, nullptr))
+    {}
+    ~AccfgCtx() {
+        if (ctx) accfg_unref(ctx);
+    }
+
+    // non-copyable
+    AccfgCtx(const AccfgCtx&) = delete;
+    AccfgCtx& operator=(const AccfgCtx&) = delete;
+
+    int init();
+    accfg_ctx* get() const { return ctx; }
 };
 
-using WorkQueueSet = std::vector<WorkQueueId>;
+/// Immutable and unique id of a queue: device_id and wq_id, as in qw<device_id>.<wq_id>.
+/// Since detect/setup_devices() is guaranteed to happen in the same process, we can store
+/// accfg handles as they won't become invalid in the meantime (require accfg_ctx* though).
+struct WorkQueueId
+{
+    accfg_device* device_handle = nullptr;
+    accfg_device_type device_type = accfg_device_type::ACCFG_DEVICE_TYPE_UNKNOWN;
+    int device_id = -1;
+    int wq_id = -1;
+};
+
+/// To be able to store handles, we must carry around the ctx as well.
+struct WorkQueueSet
+{
+    AccfgCtx ctx;
+    std::vector<WorkQueueId> visible_wqs;
+    bool empty() const noexcept { return visible_wqs.empty(); }
+};
+
 using EnabledDevices = WorkQueueSet;
 
+// TODO: Apart from obvious Enabled and Disabled states, there are also Quiescing
+// and Locked transitional/protective states. If those states are spotted during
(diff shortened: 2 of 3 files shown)


framework/device/idxd/idxd_device.h
```diff
@@ -8,6 +8,20 @@
...
+    DEV_TYPE_DSA,
+    DEV_TYPE_IAX,
+} dev_type_t;
+
+struct bdf_t
+{
+    uint16_t domain;
+    uint8_t bus;
+    uint8_t device;
+    uint8_t function;
```
thiagomacieira: Could be:
```suggestion
    uint8_t device : 5;
    uint8_t function : 3;
```
(or the other way around for endianness)

which would make `bdf_t` be exactly 32 bits (right now, it's 48 bits). I don't know if it is worth the effort... probably not.

framework/device/idxd/idxd_device.h
```diff
@@ -21,6 +35,17 @@ struct wq_info_t
...
+    /// WQ unique index within a device.
+    int wq_id;
+    /// Device that this WQ belongs to.
+    int device_id;
+
+    /// Device type that this WQ belongs to (DSA/IAX).
+    dev_type_t dev_type;
+
+    /// PCI location.
+    struct bdf_t bdf;
```
thiagomacieira: Nitpick for packing: move the `package_id` next to (above or below) `bdf_t` as both require 16-bit alignment but the rest of the types require 32 bits.

framework/device/idxd/topology_idxd.hpp
```diff
@@ -6,26 +6,96 @@
...
 
 using WorkQueueSet = std::vector<WorkQueueId>;
 using EnabledDevices = WorkQueueSet;
 
+enum class WorkQueueMode
+{
+    Dedicated,
+    Shared,
+    Unknown,
+};
```
thiagomacieira:
```suggestion
enum class WorkQueueMode : int8_t
{
    Dedicated,
    Shared,
    Unknown = -1,
};
```

framework/device/idxd/topology_idxd.hpp
```diff
@@ -6,26 +6,96 @@
...
+
+enum class WorkQueueState
+{
+    Enabled,
+    Disabled,
+    Unknown,
+};
```
thiagomacieira:
```suggestion
enum class WorkQueueState : int8_t
{
    Enabled,
    Disabled,
    Unknown = -1,
};
```

framework/device/idxd/topology_idxd.hpp
```diff
@@ -6,26 +6,96 @@
...
+
+/// Unlike for other devices, IDXD Topology is subject to change in-between tests.
```
thiagomacieira, line 67: So should we add a field to `struct test` so each one can declare the topology they want?

framework/device/idxd/topology_idxd.hpp
```diff
@@ -6,26 +6,96 @@
...
+    };
+
+    struct Node
+    {
+        int id;
+        // TODO:
+        // Decide:
+        //   - std::vector<std::variant<DsaDevice, IaxDevice>>
+        //   - std::vector<DsaDevice> and std::vector<IaxDevice>
+        //   - one std::vector<Device> and a type member to Device?
```
thiagomacieira: I think it will depend on how we write the tests.

One idea that occurred to me is to use the type as the `device_feature`, e.g.
```c++
    .minimum_features = device_feature_dsa_v1,
```
In that case, it might be easier for us to the third: `std::vector<Device>` with each device having different features.

But I am not sure if this helps. I think we'll need to explore the `sandstone_run.cpp` code to see what fits best. Note the test that uses both DSA and IAX: that has to be supported too.

framework/device/idxd/idxd_device.h
mcisowsk: yes, I think it is

framework/device/idxd/topology_idxd.hpp
mcisowsk: either this or some lightweight `struct IdxdTest` wrapper, containing the config

framework/device/idxd/topology_idxd.hpp
mcisowsk: nice idea with the features

framework/device/idxd/idxd_device.h
```diff
@@ -8,6 +8,28 @@
 
 #include <stdint.h>
 
+typedef enum
+{
+    DEV_TYPE_DSA,
+    DEV_TYPE_IAX,
+} dev_type_t;
+
+typedef enum
```
mcisowsk: note that in following up PRs this would probably change to some libaccel-config structs. So those are mostly interim. We can also first merge #1153 and then use those structs in this PR rigth away.

framework/device/idxd/topology_idxd.hpp
```diff
@@ -6,26 +6,134 @@
...
+        bool supports_block_on_fault = false;
+        bool supports_ats_disable = false;
+        bool supports_prs_disable = false;
+
+        uint32_t op_cap[8] = {};
+
+        std::vector<Group> groups;
+    };
+
+    struct Node
```
mcisowsk: this layer is probably redundant, as we'd not divide devices further into numa nodes (no point in that)

framework/device/idxd/idxd_device.h
```diff
@@ -21,6 +43,21 @@ struct wq_info_t
...
+    int wq_id;
+
+    /// Device that this WQ belongs to.
+    int device_id;
+
+    /// Device type that this WQ belongs to (DSA/IAX).
+    dev_type_t dev_type;
+
+    /// Device version (V1/V2/V3).
+    dev_version_t dev_version;
```
mcisowsk: could we store a pointer to topology for easier access to mutable config parameters in tests?
```c++
/// Pointer to Topology::WorkQueue which contains mutable config parameters of the wq.
Topology::WorkQueue* wq;
```
or
```c++
/// Pointer to Topology::WorkQueue which contains mutable config parameters of the wq.
void* wq_topo;
```

framework/device/idxd/idxd_device.h
mcisowsk: done

framework/device/idxd/idxd_device.h
```diff
@@ -8,6 +8,28 @@
...
+    DEV_VERSION_V3 = 0x300,
+} dev_version_t;
+
+struct bdf_t
+{
+    uint16_t domain;
+    uint8_t bus;
+    uint8_t device : 5;
+    uint8_t function : 3;
+};
```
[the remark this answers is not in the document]
thiagomacieira: It's not cross-platform: this is expected only on x86 since it's Intel-specific accelerators.

framework/device/idxd/idxd_device.h
thiagomacieira: I don't understand the use-case. What are you trying to do? If it is about the test describing what it wants, I'd expect it to be in `struct test`. If it is the test getting information about the configured state, then either this struct suffices or it can simply get `Topology::topology()` and query from there.

framework/device/idxd/topology_idxd.hpp
```diff
@@ -6,29 +6,138 @@
...
+        accfg_wq_mode mode = accfg_wq_mode::ACCFG_WQ_MODE_UNKNOWN;
+        accfg_wq_state state = accfg_wq_state::ACCFG_WQ_UNKNOWN;
+        accfg_wq_type type = accfg_wq_type::ACCFG_WQT_NONE;
+
+        int group_id = -1; // same as Group::id
+
+        // present on all queues
+        uint64_t size = 0;
+        uint32_t max_batch_size = 0;
+        uint64_t max_transfer_size = 0;
```
thiagomacieira: Nitpick: sort to avoid padding: `id` after `wq`, `max_batch_size` before `size`. That places 6 `int`-sized quantities together, requiring no padding to the `uint64_t` that follows.

thiagomacieira (dismissed):

framework/device/idxd/idxd_device.h
mcisowsk: idea is that we may need to query for the mutable config parameters from the `info` level, for example when printing system state. It's faster than searching through the topology.

dawid-sabat (approved):

PR #1153, opened 2026-07-01
FW/IDXD: add `libaccel-config` dependency

mcisowsk:
For now it's dynamic.

### DIFF
--- a/.github/actions/build-idxd/action.yaml
+++ b/.github/actions/build-idxd/action.yaml
@@ -49,7 +49,9 @@ runs:
             ninja-build \
             python3-yaml \
             yq \
-            zlib1g-dev
+            zlib1g-dev \
+            uuid-dev \
+            libaccel-config-dev
         echo "::endgroup::"
     - uses: actions/checkout@v6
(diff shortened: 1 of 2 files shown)


.github/actions/build-idxd/action.yaml
```diff
@@ -49,7 +49,9 @@ runs:
             ninja-build \
             python3-yaml \
             yq \
-            zlib1g-dev
+            zlib1g-dev \
+            uuid-dev \
```
thiagomacieira, line 53: Shouldn't need to add this. it's not used in the headers and shouldn't be needed for dynamic linking.

thiagomacieira (approved):

.github/actions/build-idxd/action.yaml
mcisowsk: a test usage of `#include <accel-config/libaccel_config.h>` proved otherwise

thiagomacieira: Ack.

PR #1154, opened 2026-07-01
FW/IDXD: implement `detect_devices`

mcisowsk:


framework/device/idxd/topology.cpp
```diff
@@ -37,14 +39,58 @@ void print_temperature_of_device()
 {
 }
 
+namespace {
+struct AccfgCtx
```
mcisowsk, line 43: think about moving to topology_idxd.hpp or a separate file

thiagomacieira: Do you predict needing the Ctx anywhere else? Is it useful in tests?

framework/device/idxd/topology.cpp
```diff
@@ -37,16 +39,60 @@ void print_temperature_of_device()
...
+} // end anonymous namespace
+
+/// Collect all WQs visible in the system. Do not create any hierarchy of them at this point.
 template <>
 WorkQueueSet detect_devices<WorkQueueSet>()
 {
-    WorkQueueSet enabled_devices;
+    WorkQueueSet visible_wqs;
+
+    AccfgCtx ctx;
```
thiagomacieira, line 73: We're destroying it at the end of this function. Is that legal? I'd expect with `unref` that we'd need to keep it around.

thiagomacieira (approved):

framework/device/idxd/topology.cpp
mcisowsk: we can assume that it won't survive forking so I think it's better to treat it as a function-local short-lived ctx for now. Similar as it is for GPUs

framework/device/idxd/topology.cpp
mcisowsk: I predict it being initialized by the framework internally (by some "pre-run" function), rather than used directly in tests.

PR #1157, opened 2026-07-07
FW/IDXD: implement `setup_devices` & `build_topology`

mcisowsk:


framework/device/idxd/topology.cpp
```diff
@@ -39,69 +62,163 @@ void print_temperature_of_device()
...
+        auto it = bdf_cache.find(enabled.device_id);
+        if (it == bdf_cache.end()) {
+            it = bdf_cache.emplace(enabled.device_id, detect_bdf_via_os(enabled.device_handle)).first;
+        }
+        info->bdf = it->second;
+
+        info->device_id = enabled.device_id;
+        info->wq_id = enabled.wq_id;
+        info->dev_type = enabled.device_type;
+        info->dev_version = (accfg_device_version)accfg_device_get_version(enabled.device_handle);
```
mcisowsk: probably `device_handle` is enough, no need to carry around `wq_handle`

framework/device/idxd/idxd_device.h
```diff
@@ -54,8 +54,11 @@ struct wq_info_t
     enum accfg_device_type dev_type;
 
     /// Device version (V1/V2/V3).
-    dev_version_t dev_version;
-    // enum accfg_device_version dev_version;
+    // dev_version_t dev_version;
+    enum accfg_device_version dev_version;
+
+    /// Pointer to Topology::WorkQueue which contains mutable config parameters of the wq.
+    void* wq_topo;
```
thiagomacieira: See comment on the previous PR: I don't understand the use-case for this (yet).

framework/device/idxd/build_topology.cpp
```diff
@@ -0,0 +1,108 @@
...
+    if (int v = accfg_wq_get_prs_disable(wq_handle); v >= 0) {
+        wq.prs_disable = v;
+    }
+}
+
+// search for existing group or create new
+void append_topo_device(Topology::Device& device, accfg_device* device_handle, wq_info_t* info)
+{
+    auto wq_handle = accfg_device_wq_get_by_id(device_handle, info->wq_id);
+    auto group_id = accfg_wq_get_group_id(wq_handle);
```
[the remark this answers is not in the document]
thiagomacieira: Can it return null under the conditions that this function will be called in?

framework/device/idxd/build_topology.cpp
```diff
@@ -0,0 +1,110 @@
...
+        device.groups.begin(), device.groups.end(), [&](const auto& g) { return g.id == group_id; });
+        it != device.groups.end())
+    {
+        append_topo_group(*it, device, wq_handle, info);
+    } else {
+        auto& group = device.groups.emplace_back();
+        group.id = group_id;
+        // group.engines = // TODO
+        append_topo_group(group, device, wq_handle, info);
+    }
```
thiagomacieira:
```suggestion
    auto it = std::ranges::find_if(device.groups, [&](const auto& g) { return g.id == group_id; });
    if (it == device.groups.end()) {
        // new group
        it = groups.emplace(it);
        it->group_id = group_id;
    }
    append_topo_group(*it, device, wq_handle, info);
```

framework/device/idxd/build_topology.cpp
```diff
@@ -0,0 +1,110 @@
...
+    topo.devices.reserve(device_count()); // reserve to max possible size - we need addresses stable for WorkQueue::this_device
+
+    wq_info_t* info = device_info;
+    const wq_info_t* cend = device_info + device_count();
+
+    while (info != cend) {
+        if (auto it = std::find_if(
+            topo.devices.begin(), topo.devices.end(), [&](const auto& d) { return d.id == info->device_id; });
+            it != topo.devices.end())
+        {
```
thiagomacieira: See above. Multi-line ifs are ugly; refactor if you can.

framework/device/idxd/topology.cpp
```diff
@@ -82,9 +105,120 @@ void create_mock_topology(const char *topo)
...
+
+    std::ifstream infile;
+    infile.open(file.data());
+    if (!infile.is_open()) { [[unlikely]]
+        fprintf(stderr, "%s: internal error: unable to find physical_package_id file: %m\n",
+                program_invocation_name);
+        return res;
+    }
+    infile >> res;
+    infile.close();
```
thiagomacieira: Please don't copy `ifstream`

framework/device/idxd/topology.cpp
```diff
@@ -82,9 +105,120 @@ void create_mock_topology(const char *topo)
...
+    std::error_code ec;
+    const auto target_path = std::filesystem::read_symlink(link_path, ec);
+    if (ec) [[unlikely]] {
+        return bdf;
+    }
+
+    std::stringstream bdf_ss{target_path.filename().c_str()};
+    char clobber;
+    uint16_t b{}, d{}, f{}; // can't be character types as is in bdf_t.
+    bdf_ss >> std::hex >> bdf.domain >> clobber >> b >> clobber >> d >> clobber >> f;
```
thiagomacieira: `sscanf` would be more than enough, or just `strtoul`

framework/device/idxd/topology.cpp
```diff
@@ -82,9 +105,120 @@ void create_mock_topology(const char *topo)
...
+
+    [[maybe_unused]] int i = 0;
+    for (LogicalProcessor lp = set.next(); lp != LogicalProcessor::None; ++i) {
+        auto& next_cpu = res.emplace_back(std::to_underlying(lp));
+        lp = set.next(LogicalProcessor(next_cpu + 1));
+    }
+    assert(i == set.count());
+
+    return res;
+}
```
thiagomacieira: Move it to `LogicalProcessorSet` itself.

framework/device/idxd/build_topology.cpp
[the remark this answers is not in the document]
mcisowsk: I don't think so but we should probably always check the returned pointer, no matter what.

framework/device/idxd/build_topology.cpp
mcisowsk: nice idiom

framework/device/idxd/topology.cpp
mcisowsk: where ?

framework/device/idxd/topology.cpp
mcisowsk: as you wish

framework/device/idxd/idxd_device.h
mcisowsk: helper pointer so that we are able to print the entire current system config snapshot (with both mutable and immutable parameters of the WQs) without additional traversal of the `Topology`. Moreover, tests may also want to query for mutable parameters, and this way they could to it straight from `wq_info_t*`, as `static_cast<const Topology::WorkQueue*>(info->wq_topo)-><whatever mutable config parameter>`

mcisowsk:
I missed the fact that we cannot store raw pointers inside `device_info` which is located in the shared memory. After fork the pointed address is useless. Let's store the topology path coordinates instead.

framework/device/idxd/idxd_device.h
```diff
@@ -47,6 +47,16 @@ struct wq_info_t
     /// Device version (V1/V2/V3).
     enum accfg_device_version dev_version;
 
+    /// WQ name, as seen in /dev/<dev_type>/.
+    const char* name;
+
+    /// Path within Topology.
```
mcisowsk: this should be updated after topo rebuild

framework/device/idxd/idxd_device.h
mcisowsk: actually I changed my mind

framework/device/idxd/idxd_device.h
```diff
@@ -47,6 +47,16 @@ struct wq_info_t
     /// Device version (V1/V2/V3).
     enum accfg_device_version dev_version;
 
+    /// WQ name, as seen in /dev/<dev_type>/.
+    const char* name;
```
mcisowsk: I think we can't store raw pointer if we want it to work in exec-mode, let's resign from it. Unless we set it with pointer to static storage, which I doubt

framework/device/idxd/idxd_device.h
```diff
@@ -47,6 +47,16 @@ struct wq_info_t
...
+
+    /// Path within Topology.
+    struct
+    {
+        int dev;
+        int grp;
```
thiagomacieira: No need to abbreviate here.

framework/device/idxd/build_topology.cpp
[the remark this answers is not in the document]
thiagomacieira: `get_by_id` can't deal with null pointers:
https://github.com/intel/idxd-config/blob/f8dcd7e8688e22aa25389322f6da950e4f9600e9/accfg/lib/libaccfg.c#L1832-L1841
the `foreach` will crash with a null pointer.

`accfg_wq_get_group_id` crashes with a null pointer too:
https://github.com/intel/idxd-config/blob/f8dcd7e8688e22aa25389322f6da950e4f9600e9/accfg/lib/libaccfg.c#L2449-L2461

The question for us is whether this should be a return (indicating it's a reasonable condition) or an assertion (indicating we have a bug in our own code, because this shouldn't have happened). I don't think dealing with the config changing behind our backs is an issue.

framework/device/idxd/topology.cpp
```diff
@@ -82,9 +105,107 @@ void create_mock_topology(const char *topo)
...
 template <>
-void setup_devices<WorkQueueSet>(const WorkQueueSet &enabled_devices)
+void setup_devices<WorkQueueSet>(const WorkQueueSet& enabled_devices)
 {
+    device_info = sApp->shmem->device_info;
+
+    if (const char* mock_topo = getenv("SANDSTONE_MOCK_TOPOLOGY"); SandstoneConfig::Debug && mock_topo && *mock_topo) {
+        // create_mock_topology(mock_topo); // TODO: implement me
+        return;
+    }
```
thiagomacieira: Do it like the CPU code: first the `Debug` then `getenv()`, otherwise the `getenv()` remains in release builds.

```suggestion
    if (SandstoneConfig::Debug) {
        if (const char* mock_topo = getenv("SANDSTONE_MOCK_TOPOLOGY"); mock_topo && *mock_topo) {
            create_mock_topology(mock_topo);
            return;
        }
    }
```

thiagomacieira (changes requested):
We need to decide what to do with those null-pointer warnings from the AI. First: are they legitimate issues, or could they only happen if we messed up this code, because we are scanning for workqueues and engines and groups that we've enumerated? That will tell us if we should use `assert` or not.

There are two nitpicks. I'll fix ifstream afterwards - it's overkill.

mcisowsk:
I went with `assert`s all over the place, because `detect_devices` already detected it once. We may change that in future

framework/device/idxd/topology.cpp
```diff
@@ -216,9 +239,113 @@ void create_mock_topology(const char *topo)
...
+
+    wq_info_t* info = device_info;
+    [[maybe_unused]] const wq_info_t* cend = device_info + device_count();
+
+    std::map<int, bdf_t> bdf_cache; // bdfs are unique per device
+
+    int cpu_ind = 0;
+    for (const auto &enabled : enabled_devices.visible_wqs) {
+        info->cpu_number = enabled_cpus[cpu_ind++];
+        info->package_id = detect_package_id_via_os(info->cpu_number);
```
thiagomacieira, line 330: Somewhere here we need to match device package to the CPU package. It'll probably be the domain part of the BDF, but I don't think it's required.

framework/device/idxd/topology.cpp
```diff
@@ -233,6 +360,28 @@ void analyze_test_failures_for_topology(const struct test *test, const PerThread
...
 }
 
+std::vector<const Topology::WorkQueue*> Topology::targetable_wqs(
+        std::optional<accfg_device_type> device_type,
+        std::optional<accfg_wq_mode> mode,
+        std::optional<unsigned int> op) const
+{
+    std::vector<const WorkQueue*> result;
+    for (const Device &device : devices) {
+        if (device_type && device.dev_type != *device_type)
```
thiagomacieira, line 370: Hint:
```suggestion
        if (device_type == device.dev_type)
```

`std::optional`'s associated `operator==` returns true only if the optional has a value and it matches. If it's disengaged, it returns false immediately.

thiagomacieira (approved):

PR #1158, opened 2026-07-07
FW: add `DECLARE_MANUAL_TEST` macro for declaring special tests

mcisowsk:
### **To discuss**
The variable name changed, but I don't think it's worth altering the macro to create a nicer names for manual declarations. It is what it is. 
Also, note that we now set
```
.compiler_minimum_device = device_compiler_features
```
for the `mce_check` test. Is this ok?

### DIFF
--- a/framework/sandstone.cpp
+++ b/framework/sandstone.cpp
@@ -1141,7 +1141,7 @@ int main(int argc, char **argv)
     /* Add device-specific monitoring tests to the set. They will be kept last by
      * SandstoneTestSet in case randomization is requested. */
     for (auto special_test : test_set->get_special_tests()) {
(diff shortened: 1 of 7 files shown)


thiagomacieira:
> ### **To discuss**
> The variable name changed, but I don't think it's worth altering the macro to create a nicer names for manual declarations. It is what it is.

Agreed.

> Also, note that we now set
> 
> ```
> .compiler_minimum_device = device_compiler_features
> ```
> 
> for the `mce_check` test. Is this ok?

Yes. That is a protection against the .cpp file being compiled with a higher `-march=` setting than the framework. It shouldn't happen for the MCE test, but if it did, we had better not run it, because it could otherwise crash.

thiagomacieira (approved):

PR #1159, opened 2026-07-09
FW/IDXD: implement base class for tests

mcisowsk:
The class provides implementation of the `run()` only. Opens WQ, then closes it afterwards. There are also a bunch of helpers for test development.

- EDIT: substituted with #1214

### DIFF
--- a/framework/device/idxd/base.hpp
+++ b/framework/device/idxd/base.hpp
@@ -1,0 +1,155 @@
+/*
+ * Copyright 2026 Intel Corporation.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef IDXD_BASE_H
+#define IDXD_BASE_H
+
+#include "sandstone.h"
+#include "idxd_device.h"
+#include "topology_idxd.hpp"
+
+#include <concepts>
+#include <cstdlib>
+#include <format>
+#include <string>
+
+#include <fcntl.h>
+#include <sys/mman.h>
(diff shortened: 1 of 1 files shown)


framework/device/idxd/common.hpp
```diff
@@ -0,0 +1,33 @@
...
+     * quote character.
+     */
+    std::string engine_str;
+    if (wq_topo->this_group->engine_count == 1) {
+        engine_str = std::format("engine: '{}', ", wq_topo->this_group->engines[0].name);
+    }
+
+    log_thread_context("type: %s, bdf: '%s', numa_node: %d, device: '%s', group: '%s', %swq: %s",
+                       to_string(info->dev_type), info->bdf, wq_topo->numa_node, wq_topo->this_device->name,
+                       wq_topo->this_group->name, engine_str.c_str(), info->name);
```
mcisowsk: this function shows what's missing in `wq_info_t` and `Topology`

framework/device/idxd/base.hpp
```diff
@@ -0,0 +1,121 @@
...
+#include <cstdlib>
+#include <format>
+#include <string>
+
+#include <fcntl.h>
+#include <sys/mman.h>
+#include <unistd.h>
+
+// taken from idxd_utils.h
+struct idxd_wq_handle_t
```
mcisowsk, line 22: think about which file would it fit best

PR #1160, opened 2026-07-10
FW/CPU Topology: add support for splitting groups on CPU dies

thiagomacieira:
Because on some Intel systems, we get no NUMA splitting inside of a package in spite of different memory channels. That may be a BIOS configuration. Without this change, we end up slicing the socket at boundaries that cause anywhere from "sub-optimal" to "incredibly bad" performance with some memory-intensive tests.

[ChangeLog][Topology] The framework will now use the CPU die information from CPUID to create a slicing plan.

[ChangeLog][YAML Schema Change] The `cpu-info` array in the log header will contain a `die` field for CPU cores, if the system provides that information. The field is absent otherwise. That replaces the `tile` field.

### DIFF
--- a/bats/sanity-check/20-selftests.bats
+++ b/bats/sanity-check/20-selftests.bats
@@ -2229,13 +2229,20 @@ check_thread_ratio_plans() {
     local expected=$(thread_ratio_expected_threads "$fs_count" "$ratio")
     test_yaml_numeric "/test-plans/fullsocket/0/threads" "value == $expected"
 
-    # isolate_numa slices
-    local slice_count=${yamldump[/test-plans/isolate_numa@len]}
-    for ((i = 0; i < slice_count; i++)); do
-        local count=${yamldump[/test-plans/isolate_numa/$i/count]}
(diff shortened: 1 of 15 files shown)


bats/sanity-check/20-selftests.bats
```diff
@@ -2218,7 +2218,7 @@ check_thread_ratio_plans() {
     local logical_id_key=logical
     local start_id_key=starting_cpu
     local device_info_key=cpu-info
-    if [[ "$SANDSTONE_DEVICE_TYPE" = "GPU" ]]; then
+    if [[ "$SANDSTONE_DEVICE_TYPE" != "CPU" ]]; then
```
busykai: Wouldn't this become an issue if another non-cpu device is added later?

busykai (dismissed):

bats/sanity-check/20-selftests.bats
thiagomacieira: We have already added a non-CPU device (IDXD device). I guess I shouldn't have made this change here, though. It belongs in the commit where we update the IDXD logging.

busykai (approved):

PR #1163, opened 2026-07-14
FW/IDXD: implement idxd config & topology rebuild mechanism

mcisowsk:


tests/idxd/idxd_dummy_config.cpp
```diff
@@ -0,0 +1,50 @@
...
+ * Copyright 2026 Intel Corporation.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include "sandstone.h"
+#include "idxd_config.hpp"
+
+namespace {
+// This object has static storage duration, so tests can point to it safely.
+idxd_config_t dummy_idxd_config = {
```
mcisowsk: what if a certain test just wants to enable all wqs and target them, no matter the other parts of the config? Should we add support to keywords such as "ALL_WQS", instead of specific config where we define each wq individually?

framework/device/idxd/idxd_config.cpp
```diff
@@ -0,0 +1,433 @@
...
+}
+
+int read_config(idxd_config_t::config_t& into)
+{
+    into.clear();
+
+    accfg_ctx* ctx = nullptr;
+    if (accfg_new(&ctx) < 0 || !ctx) {
+        return log_skip_or_print(RuntimeSkipCategory, "Failed to initialize accfg context");
+    }
```
thiagomacieira: `accfg_new` returns an error condition. And if it succeeds, it it won't leave `ctx` as null, so you're reverifying the internals.

The current only condition it can fail is `ENOMEM`, which can't happen for us. So feel free to turn this to an `assert()`.

framework/device/idxd/idxd_config.cpp
```diff
@@ -0,0 +1,433 @@
...
+
+        accfg_group* group;
+        accfg_group_foreach(device, group) {
+            idxd_config_t::config_t::group_t grp;
+            grp.device_id = dev.device_id;
+            grp.group_id = accfg_group_get_id(group);
+
+            if (int v = accfg_group_get_read_buffers_reserved(group); v >= 0) {
+                grp.params.read_buffers_reserved = static_cast<unsigned int>(v);
+                grp.has_read_buffers_reserved = true;
```
thiagomacieira: Can we just use the same technique of storing a negative number (-1) to indicate the feature setting is not specified/present?

framework/device/idxd/idxd_device.cpp
```diff
@@ -18,11 +19,26 @@ void dump_device_info()
 
 TestResult prepare_test_for_device(struct test *test)
 {
+    if (test->idxd_config) {
+        if (test->idxd_config->apply_desired() != EXIT_SUCCESS) {
+            return TestResult::Skipped; // or Failed?
```
thiagomacieira: Probably depends on why it failed to apply. If it is an unsupported config, skip; if there was an error, plaform-fail and exit.

framework/device/idxd/idxd_device.cpp
```diff
@@ -18,11 +19,26 @@ void dump_device_info()
...
+        rebuild_topology();
+    }
     return TestResult::Passed;
 }
 
 void finish_test_for_device(struct test *test)
 {
+    if (test->idxd_config) {
+        if (test->idxd_config->restore_previous() != EXIT_SUCCESS) {
+            log_warning("Failed to restore previous IDXD configuration"); // TODO: we should terminate the program in that case!
```
thiagomacieira: Yes, we should.

Improvement for the future: delay restoring until the next test starts. If the new config is the same as the *current* config, give it this previous so it will restore at its end (or pass it on). That avoids changing back and forth.

framework/sandstone.h
```diff
@@ -413,6 +417,13 @@ struct test {
     /// For tests that skipped in preinit, this pointer points to a string with skip message.
     void *data;
     struct test_data_per_thread *per_thread;
+
+#if SANDSTONE_DEVICE_IDXD
+    /// IDXD configuration required for the test. Can be nullptr if any configuration suffices.
+    /// It should be defined at the stage of test declaration and is applied in prepare_test_for_device().
+    /// TODO: put at the end to not corrupt struct layout, but it's not really a "private" field.
```
thiagomacieira: Aside from the fields that the `DECLARE_TEST` sets, the order of fields is unimportant, except in the way that existing C++ tests declare the fields using designated initialisers (must be in order). So if you move this to before the private data fields, we'll be fine too.

framework/device/idxd/idxd_config.hpp
```diff
@@ -0,0 +1,94 @@
...
+    int apply_desired();
+    int restore_previous();
+
+    struct config_t
+    {
+        struct device_t
+        {
+            int device_id = -1;
+            bool enabled = false;
+            dev_parameters params = {};
```
thiagomacieira: I'd rather we used our own type for this, so we could avoid `libaccel_config.h` include up top. Those types aren't used in the libaccel-config API, leading us to have to write get/set code like you did in the .cpp, so they're not really worth the trouble.

Using our own would also allow us to use NSDMI to initialise all numeric fields to -1 or `UINT_MAX` instead of zero, allowing zeroes where applicable (assuming C++-only support, which you've implied by having all of this in a header that `#include <vector>` and others).

tests/idxd/idxd_dummy_config.cpp
```diff
@@ -0,0 +1,50 @@
...
+    },
+};
+} // end anonymous namespace
+
+DECLARE_TEST(idxd_dummy_config, "Dummy IDXD test showing static idxd_config_t wiring")
+    .test_run = [](struct test*, int) { return EXIT_SUCCESS; },
+    .desired_duration = -1,
+    .quality_level = TEST_QUALITY_BETA,
+    .flags = test_schedule_sequential | test_in_parent,
+    .idxd_config = &dummy_idxd_config,
```
thiagomacieira, line 58: GCC has an extension that allows us to define the variable here. We use that for the test groups, for example (an array, but the principle stands).

tests/idxd/idxd_dummy_config.cpp
thiagomacieira: Yes, we should have pre-defined, useful configs. Two ways of doing that;
1. define global variables that we fill in during start up with the settings we want (mind `-fexec` runs!)
2. define *pointers* to dummy addresses, such as `#define ALL_WQS ((idxt_config_t*)1)` (ensure that the compilers constant-initialise the test definitions, though)

thiagomacieira (changes requested):

framework/device/idxd/idxd_config.cpp
mcisowsk: it's a hack so that we could use `group_parameters` from libaccel-config - removing it

framework/device/idxd/idxd_config.hpp
mcisowsk: yes, makes sense. Especially that we gained almost nothing and had to store the boolean fields

mcisowsk: I'd keep field `accfg_op_config op_config = {};` though

tests/idxd/idxd_dummy_config.cpp
mcisowsk: might look dirty in this particular case - the config is quite long

PR #1164, opened 2026-07-15
FW: distinguish test-the-test failures from test failures

thiagomacieira:
For the purposes of determining whether to show timing failures. We do want to see the timing information even in case a previous test-the-test failure was recorded. In this case, the failure was a too-high memory use problem.

So just check if the test has failed right at the top of the execution, before we check anything else.


mcisowsk (approved):
lgtm

PR #1166, opened 2026-07-15
FW: add `group_special` for easier disabling of special tests

mcisowsk:
In selftests, using `--ignore-mce-errors` for this purpose is good enough for CPUs, but not for other devices.

### DIFF
--- a/bats/sanity-check/helpers.bash
+++ b/bats/sanity-check/helpers.bash
@@ -4,7 +4,7 @@
 
 sandstone_selftest() {
     VALIDATION=dump
-    run_sandstone_yaml ${MAX_PROC:+-n$MAX_PROC} --ignore-mce-errors --selftests --timeout=20s --retest-on-failure=0 -Y2 "$@"
+    run_sandstone_yaml ${MAX_PROC:+-n$MAX_PROC} --ignore-mce-errors --disable "@special" --selftests --timeout=20s --retest-on-failure=0 -Y2 "$@"
 }
 
 extract_from_yaml() {
--- a/framework/sandstone_tests.cpp
+++ b/framework/sandstone_tests.cpp
@@ -17,23 +17,31 @@
 void SandstoneTestSet::load_all_tests()
 {
     std::span<struct test> known_tests = test_source();
+
+    auto add_to_all_group_map = [&](struct test &t) {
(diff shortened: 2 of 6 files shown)


framework/sandstone_tests.cpp
```diff
@@ -17,23 +17,31 @@
...
-        }
+        add_to_all_group_map(t);
     }
 
     /* add "special" tests */
     special_tests = special_tests_for_device();
     special_tests.push_back(&_test_mce_check); // mce always added last
+    for (struct test *t : special_tests) {
+        add_to_all_group_map(*t);
+    }
```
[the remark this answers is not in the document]
mcisowsk: no need for this. If someone forgot to assign his special test this group (as me in the first revision of this commit), then it's his/her fault

bats/sanity-check/helpers.bash
```diff
@@ -4,7 +4,7 @@
 
 sandstone_selftest() {
     VALIDATION=dump
-    run_sandstone_yaml ${MAX_PROC:+-n$MAX_PROC} --ignore-mce-errors --selftests --timeout=20s --retest-on-failure=0 -Y2 "$@"
+    run_sandstone_yaml ${MAX_PROC:+-n$MAX_PROC} --ignore-mce-errors --disable "@special" --selftests --timeout=20s --retest-on-failure=0 -Y2 "$@"
```
mcisowsk, line 7: one can argue if we should remove `--ignore-mce-errors` from here as its sole purpose was to disable special tests

thiagomacieira: If you meant "remove from this command-line", I agree. If you meant "remove the option from sandstone_opts.cpp", then no, it has one more effect:
https://github.com/opendcdiag/opendcdiag/blob/60449622e7b1a3e93e6344fae68622a629a237fb/framework/sysdeps/unix/child_debug.cpp#L229-L237

This was a feature request by a customer who had parts with known bad RAM and wanted the scanning tool to keep going in spite of those, to see if it there were further problems.

thiagomacieira (approved):

bats/sanity-check/helpers.bash
mcisowsk: yes, from this exact  cmdline

PR #1167, opened 2026-07-15
FW/IDXD: define and detect `device_features`

mcisowsk:


framework/device/idxd/idxd_features.h
```diff
@@ -0,0 +1,54 @@
...
+#ifndef INC_IDXD_FEATURES_H
+#define INC_IDXD_FEATURES_H
+
+#include <stdint.h>
+
+typedef unsigned __int128 device_features_t;
+#define IDXD_FEATURE_CONSTANT(bit) (((device_features_t) 1) << (bit))
+
+// TODO: is this version common for all devices in the system (both dsa & iax)?
+// It there a possibility to dsa_v1 and iax_v2, or (even worse) dsa_v1 and dsa_v2 on one system?
```
mcisowsk: in case that dsa & iax can have different versions, we can divide `device_features` into two halves maybe? 64bits per each should suffice as well. But the latter case would be harder to handle nicely

framework/device/idxd/idxd_features.h
```diff
@@ -0,0 +1,54 @@
...
+#define device_feature_v1       IDXD_FEATURE_CONSTANT(0)
+#define device_feature_v2       IDXD_FEATURE_CONSTANT(1)
+#define device_feature_v3       IDXD_FEATURE_CONSTANT(2)
```
thiagomacieira: So I recommend you remove these and just support like this...

framework/device/idxd/idxd_features.h
thiagomacieira: I don't think two different accelerator of the same accelerator versions is going to happen. We probably won't need to worry about that, like we don't need to worry about heterogeneous CPU ISA.

But different accelerators? It could happen.

framework/device/idxd/premain.cpp
```diff
@@ -0,0 +1,49 @@
...
+        if (ver >= ACCFG_DEVICE_VERSION_2)
+            features |= device_feature_v2;
+        if (ver > ACCFG_DEVICE_VERSION_2)
+            features |= device_feature_v3;
+
+        enum accfg_device_type dev_type = accfg_device_get_type(device);
+        if (dev_type == ACCFG_DEVICE_DSA)
+            features |= device_feature_dsa;
+        else if (dev_type == ACCFG_DEVICE_IAX)
+            features |= device_feature_iax;
```
thiagomacieira:
```suggestion
        device_features base;
        if (dev_type == ACCFG_DEVICE_DSA)
            base = device_feature_dsa_v1;
        else
            base = device_feature_iax_v1;
        features |= base;
        if (ver > ACCFG_DEVICE_VERSION_2)
            features |= (base << 1);
        if (ver > ACCFG_DEVICE_VERSION_2)
            features |= (base << 2);
```

framework/device/idxd/premain.cpp
```diff
@@ -0,0 +1,49 @@
...
+extern "C" {
+__attribute__((constructor(101), used))
+static void premain(int argc, char **argv, char **envp)
+{
+    (void) argc;
+    (void) envp;
+
+    device_features = detect();
+}
+}
```
thiagomacieira: I don't like this in `premain`. We needed it for CPU because of `-march=` reasons, but for any other devices we should detect in `main()`.

framework/device/idxd/idxd_features.h
```diff
@@ -0,0 +1,54 @@
...
+// know if those features can be assumed to be common for the entire system.
+
+// we assume that "compiler" will always have them all
+static const device_features_t device_compiler_features = 0
+    | device_feature_v1
+    | device_feature_v2
+    | device_feature_v3
+    | device_feature_dsa
+    | device_feature_iax
+;
```
thiagomacieira: This is wrong. That would make the framework ignore detected bits and always attempt to run, in spite of what we detected via libaccel-config.

framework/device/idxd/idxd_device.h
```diff
@@ -64,9 +66,7 @@ inline int wq_info_t::wq() const
 }
 #endif
 
-// Not used at the moment
-typedef unsigned __int128 device_features_t;
-static const device_features_t device_compiler_features = 0;
+// Define for legacy reasons?
 #define cpu_has_feature(f)      ((device_compiler_features & (f)) == (f) || (device_features & (f)) == (f))
```
thiagomacieira: Rename to `device_has_feature` ?

thiagomacieira (changes requested):

framework/device/idxd/premain.cpp
mcisowsk: right, I just copied the mechanism from CPU without much thinking

framework/device/idxd/idxd_features.h
mcisowsk: but compiler (`libaccel-config` in our case) would always "support" dsa & iax, same as with versions, since we control the version of the library that we use (and we see it has v1,v2,v3). It's way different from real "compiler" features from CPU. In my thinking all we have to worry about are runtime features

framework/device/idxd/idxd_device.h
mcisowsk: yes, would have to

framework/device/idxd/premain.cpp
mcisowsk: why?

framework/device/idxd/idxd_features.h
mcisowsk: Ok, I changed that to =0 so that we won't return early from `cpu_has_feature`

mcisowsk:
could we also add this here somewhere?
```c++
inline bool has_opcode(const accfg_op_cap& op_cap, unsigned want)
{
    const unsigned word = want / 32;
    const unsigned bit = want % 32;
    if (word >= (sizeof(op_cap.bits) / sizeof(op_cap.bits[0])))
        return false;
    return (op_cap.bits[word] & (1u << bit)) != 0;
}
```

framework/device/idxd/idxd_features.h
```diff
@@ -0,0 +1,69 @@
+/*
+ * Copyright 2026 Intel Corporation.
```
mcisowsk, line 2: tbf I am bit lost as what should be stored here and treated as `device_features`. I guess that a certain bit should be set if at least 1 WQ can satisfy certain condition. but what if dsa has fill_op, and iax not? 🤔 😕

framework/device/idxd/premain.cpp
thiagomacieira: You've implemented what I asked, just differently. The reason for me was that we should have different bits for each device, so a test could declare it needs DSAv1 and IAXv1 (or v3; unlikely to be heterogeneous).

framework/device/idxd/idxd_features.h
thiagomacieira: I suspect most features are accelerator-type-specific: fill-op is going to be DSA, compress is IAX. The tests will know that.

The only issue is if there is a feature that is common to both *and* they are not available at the same time. In that case, please add the accelerator name to the feature, e.g., "iax_op_foobar" and "dsa_op_foobar".

framework/device/idxd/idxd_features.h
```diff
@@ -0,0 +1,69 @@
...
+#define device_feature_iax      IDXD_FEATURE_CONSTANT(4)
+
+// Type-specific version bits avoid false positives on mixed systems such as
+// DSA V1 plus IAX V2.
+#define device_feature_dsa_v1   IDXD_FEATURE_CONSTANT(16)
+#define device_feature_dsa_v2   IDXD_FEATURE_CONSTANT(17)
+#define device_feature_dsa_v3   IDXD_FEATURE_CONSTANT(18)
+#define device_feature_iax_v1   IDXD_FEATURE_CONSTANT(32)
+#define device_feature_iax_v2   IDXD_FEATURE_CONSTANT(33)
+#define device_feature_iax_v3   IDXD_FEATURE_CONSTANT(34)
```
thiagomacieira, line 26: Are the version numbers redundant with the capabilities? Do we need both?

Food for thought.

framework/device/idxd/topology.cpp
```diff
@@ -47,7 +74,109 @@ int AccfgCtx::init()
...
+{
+    device_features_t features = 0;
+
+    unsigned int ver = accfg_device_get_version(device);
+    if (ver >= ACCFG_DEVICE_VERSION_1)
+        features |= device_feature_v1;
+    if (ver >= ACCFG_DEVICE_VERSION_2)
+        features |= device_feature_v2;
+    if (ver > ACCFG_DEVICE_VERSION_2)
+        features |= device_feature_v3;
```
thiagomacieira: Remove these as they are now unnecessary.

framework/sandstone.cpp
```diff
@@ -603,6 +606,9 @@ static int exec_mode_run(int argc, char **argv)
     int child_number = parse_int(argv[0]);
 
     attach_shmem(parse_int(argv[1]));
+#if SANDSTONE_DEVICE_IDXD
```
thiagomacieira, line 609: I'd expect this to be `!CPU`. Doesn't GPU need this too?

thiagomacieira (dismissed):

framework/sandstone.cpp
mcisowsk: no features in GPU builds for now

framework/device/idxd/topology.cpp
mcisowsk: so no need for a "version overall"?

thiagomacieira: I don't think so. It's also pretty easy to just declare the two accelerators at a specific version if that's what you really want.

lukasz-tuz (approved):

dawid-sabat (approved):

PR #1169, opened 2026-07-16
FW/CPU: add support for apportioning the L3 to each core group

thiagomacieira:
Following up on #1160, this PR adds support for `l3_cache_size` in `Topology::CoreGroup`. For C code, there's a new `id` field in `struct cache_info_t` that can be used to match with other CPU's cache IDs and confirm whether they are the same (shared) or not.

The YAML header will also include the detected cache information. For example, on my TGL laptop:
```yaml
cache-info:
  L1:
    size: { instruction: 32768, data: 49152 }
  L2:
    size: { unified: 1310720 }
  L3:
    size: { unified: 12582912 }
    breakdown:
    - { size: 12582912, starting_cpu: 0, count: 8 }
```

The apportionment per core group is only reported in a self-test, because it depends on how the slice planning went (and is thus dependent on the `--max-cores-per-slice=` command-line option).

### DIFF
--- a/bats/sanity-check/25-cpu-topology.bats
+++ b/bats/sanity-check/25-cpu-topology.bats
@@ -129,6 +129,41 @@ test_fail_socket1() {
     ); done
 }
 
+@test "cache-info" {
+    declare -A yamldump
+    run_sandstone_yaml --disable=\*
+
+    # cache-info must be present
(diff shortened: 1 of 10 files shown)


AthenasJimenez (approved):
LGTM

PR #1171, opened 2026-07-17
FW/ulog: mark some current zero bits as the version number

thiagomacieira:
In case we need to change the schema and/or the algorithm for calculating the short IDs. Effectively a no-op, because it's still zero.

Also don't make a function call between writing to the slots.


mcisowsk (approved):
lgtm

PR #1172, opened 2026-07-20
FW/GPU: refactor `ZeHostDataPtr` to support `void*& get()`

mcisowsk:


mcisowsk:
false alarm, may revisit in future though

PR #1175, opened 2026-07-22
FW/IDXD: add some more API for querying for opcodes

mcisowsk:
So that we don't have to query `topology()`, nor manually map features to opcodes in tests.

### DIFF
--- a/framework/device/idxd/idxd_device.h
+++ b/framework/device/idxd/idxd_device.h
@@ -81,6 +81,9 @@ inline bool has_opcode(const accfg_op_cap& op_cap, unsigned opcode)
     }
     return (op_cap.bits[idx] & (1u << bit)) != 0;
 }
+
+bool has_opcode(const wq_info_t& info, unsigned opcode);
+bool has_feature(const wq_info_t& info, device_features_t feature);
 #endif
 
 #define device_has_feature(f)      ((device_compiler_features & (f)) == (f) || (device_features & (f)) == (f))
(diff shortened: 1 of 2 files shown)


framework/device/idxd/topology.cpp
```diff
@@ -47,7 +47,68 @@ constexpr unsigned IDXD_OPCODE_SCAN              = 0x50;
 constexpr unsigned IDXD_OPCODE_EXTRACT           = 0x52;
 constexpr unsigned IDXD_OPCODE_SELECT            = 0x53;
 constexpr unsigned IDXD_OPCODE_EXPAND            = 0x56;
-} // namespace
+
+unsigned feature_to_opcode(device_features_t feature)
+{
+    switch (feature) {
```
mcisowsk, line 53: enforce that we do not call it with `device_feature_dsa` etc? assert?

framework/device/idxd/idxd_device.h
```diff
@@ -81,6 +81,9 @@ inline bool has_opcode(const accfg_op_cap& op_cap, unsigned opcode)
     }
     return (op_cap.bits[idx] & (1u << bit)) != 0;
 }
+
+bool has_opcode(const wq_info_t& info, unsigned opcode);
+bool has_opcode(const wq_info_t& info, device_features_t feature);
```
thiagomacieira: The problem I see is that both `unsigned` and `device_features_t` are plain integers, so it's unfortunately too possible to select the wrong overload. This won't compile (for good reason):
```c++
    has_opcode(wq, 1);
```
because `1` is `int` and it would be an ambiguous overload. What concerns me is when the wrong one got selected.

I'd name them differently so they are not overloads. The second one should be `has_feature`.

thiagomacieira (changes requested):

thiagomacieira (approved):

PR #1176, opened 2026-07-22
FW/IDXD: implement `apply_deviceset_param`

mcisowsk:
Supported modes:
`--deviceset=dsa0,iax2`       -> target all WQ of specified devices
`--deviceset=wq0.1,wq1.2`     -> target specific WQs, cpus are auto attached
`--deviceset=wq0.1c0,wq1.2c5` -> target specific WQs, with manually specified cpus

### DIFF
--- a/framework/device/idxd/topology.cpp
+++ b/framework/device/idxd/topology.cpp
@@ -10,14 +10,23 @@
 
 #include <accel-config/libaccel_config.h>
 
+#include <algorithm>
 #include <cassert>
+#include <cerrno>
+#include <cstring>
+#include <cstdint>
 #include <cstdio>
 #include <cstdlib>
 #include <filesystem>
 #include <format>
 #include <map>
+#include <set>
+#include <span>
+#include <string>
 #include <system_error>
+#include <limits>
 #include <optional>
+#include <utility>
 #include <vector>
 
 struct wq_info_t* device_info = nullptr;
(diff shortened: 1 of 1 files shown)


framework/device/idxd/topology.cpp
```diff
@@ -48,8 +62,196 @@ void make_rescheduler(RescheduleMode mode)
...
+        exit(EX_USAGE);
+    }
+    if (n != int(n)) {
+        fprintf(stderr, "%s: error: Invalid device set parameter: %s (out of range)\n",
+                program_invocation_name, orig_arg);
+        exit(EX_USAGE);
+    }
+    arg = endptr;       // advance
+    return int(n);
+};
```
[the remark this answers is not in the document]
thiagomacieira: Casting `long n` to `int` is not UB. It's not even implementation-defined behaviour any more since C++20 (two's complement arithmetic is enforced).

The other points are correct. But pre-existing; we need to fix all call sites.

framework/device/idxd/topology.cpp
```diff
@@ -71,8 +79,183 @@ const Topology& Topology::topology()
...
+        int wq_id;
+        bool operator()(const wq_info_t &wq)
+        { return wq.device_id == device_id && wq.wq_id == wq_id; }
+    };
+
+    std::span<wq_info_t> old_wq_info(device_info, sApp->device_count);
+    std::vector<wq_info_t> new_wq_info;
+    int total_matches = 0;
+
+    std::set<std::pair<uint32_t, uint32_t>> result; // set of unique wqs to disallow duplicate entries
```
thiagomacieira, line 199: I really dislike `std::pair`. Please use a real type with meaning here.

From the use, it looks like a `wq_info_t`.

thiagomacieira (changes requested):

framework/device/idxd/topology.cpp
mcisowsk: I think that from the later constructions we can easily tell the meaning, but let's proceed with an additional struct

mcisowsk: well, actually, `std::set<std::pair>` does the sorting for us. Are you sure you want to introduce all the necessary boilerplate only so that we won't use `std::pair`? I am not convinced

thiagomacieira: It's as easy as adding 
```c++
friend auto operator<=>(mytype, mytype) const noexcept = default;
```

But I'll accept.

thiagomacieira (approved):

PR #1190, opened 2026-08-19
Bats: Fix ppin handling when it starts with 0x0...

dawid-sabat:


bats/sanity-check/25-cpu-topology.bats
```diff
@@ -116,7 +116,10 @@ test_fail_socket1() {
                 test_yaml_numeric "/cpu-info/$i/microcode" "value == $v"
             fi
             if v=`cat topology/ppin 2>/dev/null`; then
-                test_yaml_expr "/cpu-info/$i/ppin" = "${v#0x}"
+                v=${v#0x}
+                v=${v,,}
+                printf -v v '%016s' "$v"
```
thiagomacieira: I think we can remove the first two lines, no?
```suggestion
                # re-format with exactly 16 hex digits, no prefix
                printf -v v '%016x' "$v"
```
There should be no need to replace spaces with zeroes below.

thiagomacieira (changes requested):

thiagomacieira (approved):

PR #1195, opened 2026-08-21
CI: use older Debian snapshot to fix GPU build

mcisowsk:
So that we can downgrade to working versions of libs. Latest SID uses unstable versions, and we're observing strange behaviours during compilation of OpenCL kernels.

### DIFF
--- a/.github/actions/build-gpu/action.yaml
+++ b/.github/actions/build-gpu/action.yaml
@@ -27,6 +27,9 @@ runs:
     - name: Install distro packages
       shell: bash
       run: |
+        DEBIAN_SNAPSHOT=20260818T000000Z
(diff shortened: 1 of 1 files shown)


.github/actions/build-gpu/action.yaml
```diff
@@ -27,6 +27,9 @@ runs:
     - name: Install distro packages
       shell: bash
       run: |
+        DEBIAN_SNAPSHOT=20260818T000000Z
```
thiagomacieira, line 30: This is probably a temporary solution only: I suspect the snapshots will be deleted soon.

thiagomacieira (commented):
I prefer #1196

thiagomacieira:
Went with the ubuntu fix in #1196.

PR #1197, opened 2026-08-21
FW: add the ability to trace tests with the SDE tool

thiagomacieira:
This adds the `--sde-trace=` option, which can be specified multiple times to provide multiple options to the tool. The tool executable will be found in $PATH (as "sde64") if not provided as the first argument.

This option implies a few extra settings to make running tests more useful:
 `-n1` (can be overridden)
 `--max-test-loop-count=1`

The output file will be named after the current test being run, but without a timestamp. Those files may get overwritten when run again.

Example:
```bash
./opendcdiag -o - --sde-trace=/opt/intel/sde-external-10.8.0-2026-03-15-lin/sde64 --sde-trace=-nvl -e zlib1
```

You *probably* want to run this in an optimised build of the tool.

If you run with more than one thread (`-n2`), you may want to pass either `--sde-trace=-dt_print_tid` or `--sde-trace=-dt_per_thread` so you can tell what belongs to each thread.

### DIFF
--- a/framework/sandstone_opts.cpp
+++ b/framework/sandstone_opts.cpp
@@ -905,6 +905,14 @@ struct ProgramOptionsParser {
                 fprintf(ERR_STREAM, "unknown value to -f\n");
                 return EX_USAGE;
             }
+            if (app_cfg->fork_mode != SandstoneApplication::ForkMode::exec_each_test &&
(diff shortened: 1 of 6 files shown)


mcisowsk (approved):
lgtm

PR #1201, opened 2026-08-25
FW/IDXD: implement `thread_id_header_for_device`

mcisowsk:
Example:
```
  - thread: 0
    id: { device: iax1, wq: 4, group: 123, logical_cpu: 12, pci_address: 0000:6c:01.0 }
```

### DIFF
--- a/framework/device/idxd/logging_idxd.cpp
+++ b/framework/device/idxd/logging_idxd.cpp
@@ -4,11 +4,77 @@
  */
 
 #include "logging.h"
+#include "idxd_device.h"
+#include "topology_idxd.hpp"
+
+#include <algorithm>
+#include <format>
+#include <string>
 
 #if !SANDSTONE_NO_LOGGING
+static std::string to_string(accfg_device_type device_type)
+{
+    switch (device_type) {
+    case ACCFG_DEVICE_DSA:
+        return "dsa";
+    case ACCFG_DEVICE_IAX:
+        return "iax";
+    default:
+        return {};
+    }
+}
+
+static auto calc_spacing()
+{
+    static const auto spacing = []() {
+        struct {
+            int device;
+            int wq;
+            int cpu;
+            uint32_t group;
+        } result = {};
+        const auto digit_count = [](auto value) {
+            int result = 0;
+            do { value /= 10; result++; } while (value != 0);
+            return result;
+        };
+        const auto max_value = [](auto member) {
+            return std::max_element(
+                device_info, device_info + thread_count(),
+                [member](const auto& first, const auto& second) { return first.*member < second.*member; }
+            )->*member;
+        };
+
+        result.device = digit_count(max_value(&wq_info_t::device_id));
+        result.wq = digit_count(max_value(&wq_info_t::wq_id));
+        result.cpu = digit_count(max_value(&wq_info_t::cpu_number));
+        const auto max_groups = std::max_element(
+            Topology::topology().devices.begin(), Topology::topology().devices.end(),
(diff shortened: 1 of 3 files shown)


thiagomacieira (dismissed):

framework/device/idxd/logging_idxd.cpp
```diff
@@ -4,11 +4,54 @@
...
+        return "iax";
+    default:
+        return {};
+    }
+}
+
+static auto calc_spacing()
+{
+    // Note: this assumes the topology won't change after the first time this
+    // function is called. And it shouldn't, in the sense of devices' ids at least.
```
jgasparakis: We need to think this assumption a bit more (see my other comment https://github.com/opendcdiag/opendcdiag/pull/1204#discussion_r3919844353)

framework/device/idxd/logging_idxd.cpp
```diff
@@ -4,11 +4,54 @@
...
 std::string AbstractLogger::thread_id_header_for_device(int thread, LogLevelVerbosity verbosity)
 {
+    const wq_info_t* info = device_info + thread;
     std::string line;
+
+    line = std::format("{{ device: {}{}, wq: {}, ", to_string(info->dev_type), info->device_id, info->wq_id);
+    line += std::format("logical_cpu: {:{}}, ", info->cpu_number, calc_spacing());
+    line += std::format("pci_address: {:04x}:{:02x}:{:02x}.{:01x}",
+        info->bdf.domain, info->bdf.bus, info->bdf.device, info->bdf.function
+    );
```
jgasparakis, line 75: We also need the group id

mcisowsk: At first I wanted to report just "invariants", but you're probably right. It'd be useful

framework/device/idxd/logging_idxd.cpp
mcisowsk: `device_id`, `wq_id`, and `cpu_number` won't change. If we add `group_id`, then only this field is susceptible to topology changes. We need to come up with some universal spacing, maybe 3 digits is enough?
EDIT: we have `accfg_device_get_max_groups()` :)

framework/device/idxd/logging_idxd.cpp
thiagomacieira: Looks like you've added it.

framework/device/idxd/logging_idxd.cpp
```diff
@@ -4,11 +4,77 @@
...
+        };
+
+        result.device = digit_count(max_value(&wq_info_t::device_id));
+        result.wq = digit_count(max_value(&wq_info_t::wq_id));
+        result.cpu = digit_count(max_value(&wq_info_t::cpu_number));
+        const auto max_groups = std::max_element(
+            Topology::topology().devices.begin(), Topology::topology().devices.end(),
+            [](const auto& first, const auto& second) { return first.max_groups < second.max_groups; }
+        )->max_groups;
+        result.group = digit_count(max_groups - 1);
```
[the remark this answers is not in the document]
thiagomacieira: Can `max_groups == 0` in a valid run where we'd reach here?

thiagomacieira (dismissed):

framework/device/idxd/logging_idxd.cpp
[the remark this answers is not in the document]
mcisowsk: yeah, better to handle it

thiagomacieira (approved):

PR #1204, opened 2026-09-02
FW/IDXD: add detection of required CPU features

mcisowsk:
Only `waitpkg` and `enqcmd` are required system-wide. We use `__builtin_cpu_supports` to detect and tests requiring other CPU features should use it as well.

### DIFF
--- a/framework/device/idxd/topology.cpp
+++ b/framework/device/idxd/topology.cpp
@@ -29,6 +29,14 @@
 #include <utility>
 #include <vector>
 
+#include <cpuid.h>
+
+#ifdef __linux__
+#  include <asm/prctl.h>
+#  include <sys/syscall.h>
+#  include <unistd.h>
+#endif
+
 struct wq_info_t* device_info = nullptr;
 
 namespace {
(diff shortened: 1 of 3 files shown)


framework/device/idxd/topology.cpp
```diff
@@ -488,11 +546,12 @@ static device_features_t detect_features(accfg_device* device)
 
 device_features_t detect_features()
 {
+    device_features_t features = detect_cpu_features();
+
     AccfgCtx ctx;
     if (auto ret = ctx.init(); ret)
```
jgasparakis, line 550: What is the assumption of the return code of the library ctx.init() here? And how is it connected to returning the cpu features instead of idxd features? I understand if ctx.init() fails you cannot get the features but why specifically in that case return the cpu ones?

Also, if a sideband configuration is taking place (like the user running accel-config) ctx.init() might not be successful.  Maybe worth retrying 2-3 times with a few 200 msec pauses, or something like that.

Finally. Is this going to be called at sandstone init or every time a test is executed? Each approach has its advantages and disadvantages, it would be nice if we could have a command line option to choose between the two.

mcisowsk: Yeah, we may need to be more verbose about potential error on `ctx.init()`. We can do the retries as you suggest, and maybe exit app in case of no success, because there is no point in running further in such case.

This ought to be called everytime we use libaccel-config to do stuff, the stuff being either device discovery, features discovery, or applying idxd configs (TBD https://github.com/opendcdiag/opendcdiag/pull/1163). The first two happen once, the last would happen most likely at the `prepare_test_for_device`, so before each test. In the test itself, libaccel-config is not useful

thiagomacieira (changes requested):
Absence of enqcmd or waitpkg would make running any IDXD test impossible, no? The `ENQCMD` and `UMWAIT`  instructions are required to run anything, so they shouldn't be applied to tests. If they are applied, it's application-wide and we can use `__builtin_cpu_supports` from GCC.

Likewise for tests that do AMX and other accelerators (based on an old reading that AMX was an accelerator): just do the `__builtin_cpu_supports()` in the test itself.

As for BMI/BMI2, don't detect. I don't know where they are being used in internal tests, but just stop doing that or make the build be a minimum of Haswell.

framework/sandstone.cpp
```diff
@@ -608,6 +608,9 @@ static int exec_mode_run(int argc, char **argv)
     attach_shmem(parse_int(argv[1]));
 #if SANDSTONE_DEVICE_IDXD
     device_features = detect_features();
+    if (!device_features) {
+        return EX_DATAERR;
+    }
```
thiagomacieira, line 613: Where is the main process' detection?

thiagomacieira (approved):

framework/sandstone.cpp
mcisowsk: in stup_devices

PR #1205, opened 2026-09-02
FW/IDXD: include `core_id` in `wq_info_t`

mcisowsk:
For legacy reasons. May be deleted in future.

### DIFF
--- a/framework/device/idxd/topology.cpp
+++ b/framework/device/idxd/topology.cpp
@@ -565,6 +565,29 @@ int16_t detect_package_id_via_os(int cpu)
     return res;
 }
 
+int16_t detect_core_id_via_os(int cpu)
+{
+    int16_t res = -1;
+    if (cpu < 0) { [[unlikely]]
+        return res;
+    }
+    auto file = std::format("/sys/devices/system/cpu/cpu{}/topology/core_id", cpu);
+
+    FILE* fp = fopen(file.c_str(), "r");
+    if (!fp) { [[unlikely]]
+        fprintf(stderr, "%s: internal error: unable to find core_id file: %m\n",
+                program_invocation_name);
+        return res;
+    }
+    int val;
+    if (std::fscanf(fp, "%d", &val) == 1) {
+        res = static_cast<int16_t>(val);
+    }
+    fclose(fp);
+
+    return res;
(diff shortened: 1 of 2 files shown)


framework/device/idxd/topology.cpp
```diff
@@ -565,6 +565,29 @@ int16_t detect_package_id_via_os(int cpu)
     return res;
 }
 
+int16_t detect_core_id_via_os(int cpu)
```
mcisowsk, line 568: make it a template function common for core_id and physical_package_id

framework/device/idxd/topology.cpp
```diff
@@ -565,6 +565,29 @@ int16_t detect_package_id_via_os(int cpu)
...
+{
+    int16_t res = -1;
+    if (cpu < 0) { [[unlikely]]
```
jgasparakis, line 571: Why is this needed? In what case is cpu going to be < 0 ?

mcisowsk: Just to be on the safe side. It's an `int`, so technically we have to check that.

thiagomacieira, line 571: Wrong placement of the `[[unlikely]]`: it should be before the brace open. As it is, this applies to the `return` statement.

You could argue this makes the return unlikely, so the condition should be too. I make the same argument to the AI but it disagrees. Testing with [Godbolt](https://gcc.godbolt.org/z/T583KqvGn) does not show any warnings or changes in output with GCC and Clang (admittedly, micro-benchmark). It does show with MSVC and in fact it produced the same with either placement.

But let's DTRT and place it correctly.

thiagomacieira (changes requested):

framework/device/idxd/topology.cpp
thiagomacieira: Interesting, if I change the test case a little, GCC emits `jmp correct` for the incorrect function: https://gcc.godbolt.org/z/xee7zaqnf. It does that even if the `baseline` function is moved down. It does *not* `jmp baseline` if you remove the `correct` function. That means GCC considers the two token-wise identical.

thiagomacieira (approved):

PR #1208, opened 2026-09-03
FW/IDXD: retry initialization of `libaccel-config` in case of a failure

mcisowsk:
As suggested

### DIFF
--- a/framework/device/idxd/topology.cpp
+++ b/framework/device/idxd/topology.cpp
@@ -382,7 +382,13 @@ void print_temperature_of_device()
 
 int AccfgCtx::init()
 {
-    if (accfg_new(&ctx) < 0) {
+    static constexpr int MAX_RETRIES = 3;
+    int attempt = 0;
+    int ret = 0;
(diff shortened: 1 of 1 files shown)


framework/device/idxd/topology.cpp
```diff
@@ -382,7 +382,13 @@ void print_temperature_of_device()
 
 int AccfgCtx::init()
 {
-    if (accfg_new(&ctx) < 0) {
+    static constexpr int MAX_RETRIES = 3;
```
thiagomacieira, line 385: Add a comment why this is a good idea: what are we expecting to change if we retry?

thiagomacieira (changes requested):

framework/device/idxd/topology.cpp
jgasparakis: This was my suggestion thinking that there might be a configuration being applied externally while this call is made and it might fail. Doing a quick retry could just be a way to wait until this configuration gets applied and then we can proceed. But looking at the code of accfg_new() I see the only way for this to fail is a failure in calloc(). So this retry might need to go somewhere else but in this commit we should drop it, it is unnecessary.

PR #1209, opened 2026-09-03
FW/IDXD-Topology: store vector indices instead of raw IDs

mcisowsk:
IDs are sparse.

### DIFF
--- a/framework/device/idxd/build_topology.cpp
+++ b/framework/device/idxd/build_topology.cpp
@@ -10,6 +10,7 @@
 
 #include <algorithm>
 #include <cassert>
+#include <cstddef>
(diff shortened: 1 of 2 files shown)


thiagomacieira (approved):
Seems fragile to me...

mcisowsk:
> Seems fragile to me...

Would we prefer `std::map devices;` instead in `Topology` ?

thiagomacieira:
> > Seems fragile to me...
> 
> Would we prefer `std::map devices;` instead in `Topology` ?

Not necessarily. We do keep CPU indices into the `cpu_info` array, so this isn't very different. But we have to be careful.

PR #1212, opened 2026-09-08
FW: exit on non-existing test passed to `--disable`

### LINKED ISSUE #1071: No messaging when disabling invalid test.
Let's try to disable a test but make a mistake in spelling:

Example:
```
./opendcdiag -Y -vv --disable=mec_check
```

If I do the same for `--enable` I'm getting the correct messaging;
>Cannot find matching tests for 'mec_check'

Suggestion:
We need at least a warning that the user tries to disable non existing test or we should duplicate behavior from `--enable` and fail unless `--ignore-unknown-tests` is used.

mcisowsk:
[ChangeLog][Framework] Changed `--disable` to report an error when it matches no tests, instead of silently ignoring it.

----

Alternatively think about printing a warning instead.

### DIFF
--- a/bats/sanity-check/20-selftests.bats
+++ b/bats/sanity-check/20-selftests.bats
@@ -168,7 +168,11 @@ tap_negative_check() {
     if [[ "$SANDSTONE_DEVICE_TYPE" = "GPU" ]]; then
         skip "TAP skipped for GPU"
     fi
-    local -a opts=(--output-format=tap --quick --selftests --quiet --disable="*fork" -e @positive)
+    local -a opts=(--output-format=tap --quick --selftests --quiet -e @positive)
+    if ! $is_windows; then
+        # tests not defined for windows
+        opts+=(--disable="*fork")
+    fi
     $SANDSTONE "${opts[@]}" > $BATS_TEST_TMPDIR/output.tap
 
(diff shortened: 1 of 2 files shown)


dawid-sabat (commented):
We should be consistent with `--ignore-unknown-tests` behavior:

```
Ignore unknown tests specified with --enable, --disable, or test list file.
```

I thought exiting with error on unknown test disabled it too excessive but considering how ignoring unknown test should work this change makes sense.

By the way do we cover `--disable`/`--enable` and `--ignore-unknown-tests` interaction in bats?

mcisowsk:
We do, and now it finally makes sense for `--disable` case:
```sh
@test "--ignore-unknown-tests" {
    local invalid_test=this_test_does_not_exist
    declare -A yamldump

    # Confirm it doesn't exist
    set -- -e $invalid_test -e selftest_pass
    run $SANDSTONE --selftest "$@"
    ((status == 64))
    [[ "$output" = *\'$invalid_test\'* ]]

    # Now check we can ignore it
    selftest_pass --ignore-unknown-test "$@"

    # Ditto on ignoring it
    selftest_pass --ignore-unknown-test --disable=$invalid_test "$@" # -> previously it would pass even without --ignore-unknown-test
}
```

thiagomacieira (approved):

PR #1214, opened 2026-09-08
FW/IDXD: add C helpers for opening WQ

mcisowsk:
Helpers for opening and closing IDXD work-queue portals, along with the handle structure used by those helpers. These will be used by IDXD tests. We use C for legacy reasons.

### DIFF
--- a/framework/device/idxd/base.h
+++ b/framework/device/idxd/base.h
@@ -1,0 +1,86 @@
+/*
+ * Copyright 2026 Intel Corporation.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef IDXD_BASE_H
+#define IDXD_BASE_H
+
+#include "sandstone.h"
+#include "idxd_device.h"
+
+#include <fcntl.h>
+#include <limits.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <sys/mman.h>
+#include <unistd.h>
+
+#define WQ_PORTAL_SIZE   4096
+
+struct idxd_wq_handle_t
+{
+    int fd;
+    void* reg;
+
+    void* request;
+    void* completion;
+};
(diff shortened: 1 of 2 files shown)


framework/device/idxd/base.h
```diff
@@ -0,0 +1,86 @@
...
+        break;
+    default:
+        log_skip(RuntimeSkipCategory, "Unknown device type");
+        return EXIT_SKIP;
+    }
+
+    char path[PATH_MAX];
+    snprintf(path, sizeof(path), "/dev/%s/wq%d.%d", device_type, info->device_id, info->wq_id);
+
+    handle->fd = open(path, O_RDWR);
```
thiagomacieira, line 49: `O_CLOEXEC`

framework/device/idxd/base.h
```diff
@@ -0,0 +1,86 @@
...
+    if (handle->fd < 0) {
+        log_skip(RuntimeSkipCategory, "Cannot open portal");
```
thiagomacieira, line 51: `%m` (we're on Linux, so that's an acceptable way to include strerror). Repeats below.

thiagomacieira (changes requested):
Why is this a header? What is going to use this? Need more explanation in the commit message.

thiagomacieira:
> Why is this a header? What is going to use this? Need more explanation in the commit message.

Sorry, need still more info. Do these need to be `inline`? These are `static inline` (because C likes to do things differently), so they will be emitted in each and every one of the tests that end up using them. I'd prefer they be in C++ so we avoid the more dangerous C aspects of these, plus integrate with the framework wherever applicable.

Are they temporary? That is, a porting layer until OpenDCDiag properly supports whatever those tests needed? If so, say so in the commit message.

PR #1215, opened 2026-09-10
FW: dump device state regardless of `CrashContext`

mcisowsk:
It's not using the context anyway. It's useful for GPUs.

### DIFF
--- a/framework/sysdeps/unix/child_debug.cpp
+++ b/framework/sysdeps/unix/child_debug.cpp
@@ -980,21 +980,25 @@ static void print_crash_info(int slice, const char *pidstr, CrashContext &ctx)
         generate_backtrace(pidstr, slice, handle, thread);
     }
 
-    if (!handle)
-        return;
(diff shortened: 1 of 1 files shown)


framework/sysdeps/unix/child_debug.cpp
```diff
@@ -980,21 +980,25 @@ static void print_crash_info(int slice, const char *pidstr, CrashContext &ctx)
         generate_backtrace(pidstr, slice, handle, thread);
     }
 
-    if (!handle)
+    std::string log;
+    dump_device_state(log, thread);
```
thiagomacieira, line 984: CPU one only does anything if `log` is not empty. So this is going to remove the `"Registers"` prefix.

This needs a bit of refactoring.

thiagomacieira (changes requested):

thiagomacieira (approved):
Look ok now. Please squash so I can approve.

PR #1217, opened 2026-09-11
FW/IDXD: do not force `cpu_ind` to be unique

mcisowsk:
In fact, for some systems it won't be even possible, as there can be more WQs than CPUs. This is fine, though.


thiagomacieira (approved):
While this change is correct, I think we need to do things differently. The way this is going to go, we'll overload some CPUs to run the tests, with context-switching. Maybe that's good, maybe not...

mcisowsk:
> While this change is correct, I think we need to do things differently. The way this is going to go, we'll overload some CPUs to run the tests, with context-switching. Maybe that's good, maybe not...

We would, but less cpus than wqs is not a reason to not run at all. We can either print a warning or restrict `num_devices` to number of CPUs in the system.

PR #1220, opened 2026-09-16
FW: Add memcmp_pattern_or_fail helper function

lukasz-tuz:
Commit adds a new helper function which can be used to verify if a buffer is filled with a specific pattern. The most common case is to use memset to fill a buffer with a value and then use the memcmp_byte_or_fail wrapper to verify there are no corruptions in the buffer's contents.


framework/sandstone.h
```diff
@@ -710,18 +727,7 @@ static inline void memcmp_or_fail(const T *actual, const T *expected, size_t cou
...
-        token = &formatter;
-        cb = [](const void *token, void *, ptrdiff_t idx) -> std::string {
-            auto formatter = static_cast<const Fn *>(token);
-            if constexpr (std::is_invocable_v<Fn>)
-                return (*formatter)();
-            else
-                return (*formatter)(idx);
-        };
-    }
+    make_formatter_cb(formatter, cb, token);
```
thiagomacieira, line 730:
```suggestion
    auto [cb, token] = make_formatter_cb(formatter);
```
And see above.

framework/sandstone.h
```diff
@@ -688,6 +688,23 @@ report(const void *, const void *, size_t, DataType, FormatterCallback,
...
+            cb = [](const void *token, void *, ptrdiff_t idx) -> std::string {
+                auto formatter = static_cast<const Fn *>(token);
+                if constexpr (std::is_invocable_v<Fn>)
+                    return (*formatter)();
+                else
+                    return (*formatter)(idx);
+            };
+        }
+    }
+}
```
thiagomacieira, line 706:
```suggestion
template <FormatterFunction Fn> static inline auto make_formatter_cb(const Fn &formatter)
{
    struct R {
        FormatterCallback cb;
        void *token;
    } result = {};
    if constexpr (!std::is_null_pointer_v<Fn> && !SandstoneConfig::NoLogging) {
        token = &formatter;
        cb = [](const void *token, void *, ptrdiff_t idx) -> std::string {
            auto formatter = static_cast<const Fn *>(token);
            if constexpr (std::is_invocable_v<Fn>)
                return (*formatter)();
            else
                return (*formatter)(idx);
        };
    }
    return result;
}
```

thiagomacieira (changes requested):
Looks ok. Please add a selftest that passes (ok to add to one of the existing, passing tests) and a negative one next to the "datacomparefail" existing ones.

thiagomacieira (changes requested):
Looks good. Please squash.

PR #1226, opened 2026-09-21
simd.conf: enable ACE v1 spec features

busykai:
Enable the CPUID feature flags defined by the ACE v1 Specification:

- Fixed the ACE_VSN register: the spec defines it at
  CPUID.(EAX=1Dh,ECX=2):EAX[7:0], but simd.conf and the runtime
  detection code in cpuid_internal.h both read it from ECX instead.
- ace1 (ACE_VSN >= 1), requires "ace"
- avx10_v1_aux (leaf 0x24, sub-leaf 1, ECX[2]), requires "avx10.1"
- avx10_v2_aux (leaf 0x24, sub-leaf 1, ECX[3]), requires "avx10.2"

Registered the new Leaf1D_02EAX and Leaf24_01ECX leaves in
x86simd_generate.pl and added the CPUID query in cpuid_internal.h
for avx10_v1_aux/avx10_v2_aux runtime detection.


framework/device/cpu/cpuid_internal.h
```diff
@@ -206,6 +206,10 @@ static device_features_t detect_cpu()
             features |= cpu_feature_avx10_4;
 #endif
         assert(avx10ver < 5 && "Internal error: update code above!");
+
+        // AVX10_V1_AUX / AVX10_V2_AUX live in leaf 0x24, sub-leaf 1, ECX
```
thiagomacieira: Remove comment. There will like be more features in the future and this would become stale.

framework/device/cpu/cpuid_internal.h
```diff
@@ -206,6 +206,10 @@ static device_features_t detect_cpu()
...
+        __cpuid_count(0x24, 1, eax, ebx, ecx, edx);
+        features |= parse_register(Leaf24_01ECX, ecx);
```
thiagomacieira: Check that the sub-leaf is allowed:
```suggestion
        if (eax >= 1) {
            __cpuid_count(0x24, 1, eax, ebx, ecx, edx);
            features |= parse_register(Leaf24_01ECX, ecx);
        }            
```

thiagomacieira (changes requested):

thiagomacieira (approved):

